PR #10, opened 2016-02-16
Docs

edunham:
The README stuff could be wrong, but it details the steps I've taken so far to get http://54.176.36.178 looking like http://ncameron.org/perf-rustc/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 rustc-perf
 ----------
 
-This is the website for Rust compiler performance monitoring. 
+This is the website for Rust compiler performance monitoring. The website
+contains a backend to process the raw data and expose it to the frontend,
+which displays graphs to the user.
 
-https://github.com/nrc/benchmarks contains the raw data from which rustc-perf
-pulls. 
+https://github.com/nrc/rustc-timing contains the raw data from which rustc-perf
+pulls.
 
 https://github.com/nrc/benchmarks runs elsewhere and performs the tests whose
 results rustc-perf displays, then pushes them to the benchmarks repo.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -0,0 +1,42 @@
+rustc-perf
+----------
+
+This is the website for Rust compiler performance monitoring. 
+
+https://github.com/nrc/benchmarks contains the raw data from which rustc-perf
```
nrc: SHoudl be https://github.com/nrc/rustc-timing

README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```
+sudo apt-get install npm apache2 nodejs
+git clone https://github.com/nrc/rustc-perf.git /var/www/html/
+cd /var/www/html/rustc-perf; ./init.sh
+```
+
+The following lines in `/etc/apache2/sties-enabled/000-default.conf` proxy
+traffic to the app's custom port. This is historical cruft from its previous
+shared setup.
```
nrc: It's not clear if this is historical cruft that is needed anyway or should be deleted by the user

init.sh
```diff
@@ -0,0 +1,2 @@
+sudo npm install nodegit
+git clone https://github.com/nick29581/rustc-timing.git data
```
nrc: s/nick29581/nrc

README.md
edunham: oops, that was a pastefail. Thanks for catching it!

README.md
edunham: I didn't know that either when writing the commit. Pretty sure it's actually essential for the "both halves of frontend share one host" setup. Rewritten.

PR #19, opened 2016-07-31
Fixed stats page not showing results

Mark-Simulacrum:
Code checked for the length property on an object, instead of on the object's keys.


nrc:
Thank you!

PR #20, opened 2016-07-31
Update JS backend code to ES6.

Mark-Simulacrum:
Modernizes and makes the code more readable. No intentional changes to
functionality (other than fixing the median functionality as discussed on IRC). Also updated the 60 day comment to 30 days.

### DIFF
--- a/backend/perf.js
+++ b/backend/perf.js
@@ -148,7 +148,7 @@ function read_files(repo_loc, callback) {
 		}
 
 		for (let timing of times) {
-			for (let phase in timing.times)
+			for (let phase in timing.times) {
 				if (!phase_list.includes(phase)) {
 					phase_list.push(phase);
 				}
@@ -212,11 +212,15 @@ function summarise_data(i, date, kind, summary_state) {
 		by_crate[crate_name] = {};
 		let cur_summary = {};
 		for (let phase of Object.keys(weeks[0][crate_name])) {
+            if (!weeks[1][crate_name][phase] || !weeks[2][crate_name][phase]) {
+                continue;
+            }
+
 		    // Find the median value.
 			let values = [
-				week_data[0][crate_name][phase].time,
-				week_data[1][crate_name][phase].time,
-				week_data[2][crate_name][phase].time
+				weeks[0][crate_name][phase].time,
+				weeks[1][crate_name][phase].time,
+				weeks[2][crate_name][phase].time
 			];
 			values.sort((a, b) => a - b);
 			let median = values[1];
--- a/backend/perf.js
+++ b/backend/perf.js
@@ -27,13 +27,15 @@ const WEEKS_IN_SUMMARY = 12;
 
 // Data stored by test name and date, each date maps to a table storing the times by crate and by phase.
 // See make_data() for the object layout.
-var data = {};
 var opts = { compare: function(x, y) { return x.date - y.date; }};
-data.rustc = sortedList.create(opts);
-data.benchmarks = sortedList.create(opts);
-var summary = {};
-summary.rustc = [];
-summary.benchmarks = [];
+var data = {
+    rustc: sortedList.create(opts),
+    benchmarks: sortedlist.create(opts)
+};
+var summary = {
+    rustc: [],
+    benchmarks: []
+};
 var benchmarks = [];
 
 var last_date;
--- a/backend/perf.js
+++ b/backend/perf.js
@@ -203,7 +203,6 @@ function summarise_data(i, date, kind, summary_state) {
         let idx = Math.max(start_idx - j, 0);
         weeks.push(data[kind][idx].by_crate);
     }
-    // let week = data[kind][idx].by_crate;
 
     // Compute the difference as a percent for every value of data.
     for (let crate_name in weeks[0]) {
--- a/backend/perf.js
+++ b/backend/perf.js
@@ -16,25 +16,24 @@
 
 // On request for data, produce data in JSON format from our db.
 
-var http = require('http');
-var url = require('url');
-var async = require('async');
-var path = require("path");
-var fs = require('fs');
-var nodegit = require('nodegit');
-var sortedList = require('sortedlist');
+let http = require("http");
+let url = require("url");
+let path = require("path");
+let fs = require("fs");
+let nodegit = require("nodegit");
+let sortedList = require("sortedlist");
 
-var WEEKS_IN_SUMMARY = 12;
+const WEEKS_IN_SUMMARY = 12;
 
 // Data stored by test name and date, each date maps to a table storing the times by crate and by phase.
 // See make_data() for the object layout.
 var data = {};
-var opts = { compare: function(x, y) { return x.date - y.date; }}
-data['rustc'] = sortedList.create(opts);
-data['benchmarks'] = sortedList.create(opts);
+var opts = { compare: function(x, y) { return x.date - y.date; }};
+data.rustc = sortedList.create(opts);
+data.benchmarks = sortedList.create(opts);
 var summary = {};
-summary['rustc'] = [];
-summary['benchmarks'] = [];
+summary.rustc = [];
+summary.benchmarks = [];
 var benchmarks = [];
 
 var last_date;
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
Actually, now that I think of it, it does change the functionality--I changed `read_files` to blocking instead of asynchronous.. but I can change that if that's considered a bad thing.

backend/perf.js
```diff
@@ -148,7 +148,7 @@ function read_files(repo_loc, callback) {
 		}
 
 		for (let timing of times) {
-			for (let phase in timing.times)
+			for (let phase in timing.times) {
```
nrc: I'm a bit confused about what is going on here, you seem to add an unbalanced brace - is this just in the wrong commit?

Mark-Simulacrum: Yeah, I'm not sure how this snuck in the previous commit.

nrc:
I'm a bit confused by commit 3 - I don't see how that fixes the bug, perhaps the fix is in a different commit or something?

nrc:
Could you stick to spaces rather than hard tabs please? As well as trying to keep consistent, it is screwing up the indentation in GitHub :-s

backend/perf.js
```diff
@@ -16,25 +16,24 @@
...
-data['rustc'] = sortedList.create(opts);
-data['benchmarks'] = sortedList.create(opts);
+var opts = { compare: function(x, y) { return x.date - y.date; }};
+data.rustc = sortedList.create(opts);
+data.benchmarks = sortedList.create(opts);
 var summary = {};
-summary['rustc'] = [];
-summary['benchmarks'] = [];
+summary.rustc = [];
+summary.benchmarks = [];
```
nrc: This could be an object literal, rather than setting the fields outside the object

Mark-Simulacrum:
Commit 3 fixes it by introducing [these lines](https://github.com/rust-lang-nursery/rustc-perf/pull/20/commits/b5ecf2b621aaa2ad388219f16490387cec40cc98#diff-8a18e115a8fa64b81ca1f7a5167bf197R215) which `continue` when the phase isn't present in all three weeks.

Sure, I can change that. I'm guessing Sublime was feeling overzealous about spacing.

nrc:
Looks good, but is pretty hard to review because of the spaces -> tabs change.

Is there a reason to change read_files to blocking? I don't think it matters which way it is written, but the old way seems like more idiomatic node.js.

nrc:
Ah, and I guess you had previously unrolled the loop into https://github.com/rust-lang-nursery/rustc-perf/pull/20/commits/b5ecf2b621aaa2ad388219f16490387cec40cc98#diff-8a18e115a8fa64b81ca1f7a5167bf197R221 which finds the median by actually looking at three different entries rather than the same entry three times?

Mark-Simulacrum:
Fixed the spaces -> tabs (hopefully).

The read_files change was done to make sure that there is never a partial update state (whether on init or after a github webhook update). I'm not sure if it matters, but with the old code, if someone requested the page soon after the webhook came in it wouldn't display the right data (at least I think so, not actually certain about this--I'm not sure exactly how Node's event loop works).

init.sh
```diff
@@ -1,2 +1,2 @@
-sudo npm install nodegit http url async path fs sortedlist
+sudo npm install nodegit sortedlist
```
Mark-Simulacrum: These are all core node modules, no need to install them.

Mark-Simulacrum:
Yeah, I'm not 100% happy with the current median computation code, but I don't think the loop was ever in that code (see https://github.com/rust-lang-nursery/rustc-perf/pull/20/files#diff-8a18e115a8fa64b81ca1f7a5167bf197L197), I just changed it to 4 lines instead of one for readability.

backend/perf.js
```diff
@@ -80,275 +79,285 @@ function init(repo_loc, callback) {
...
+            continue;
+        }
+        let test_name = filename.substring(0, filename.indexOf("--"));
+
+        let date = new Date(header.date);
+        if (isNaN(date)) {
+            let date_components = [0, 0, 0, 0, 0, 0];
+            let actual = header.date.split("-").map(comp => parseInt(comp, 10));
+
+            for (let i = 0; i < actual.length; i++) {
```
nrc: Why change from `i in actual`?

backend/perf.js
```diff
@@ -80,275 +79,285 @@ function init(repo_loc, callback) {
...
+                            date_components[3],
+                            date_components[4],
+                            date_components[5]);
+        }
+
+        if (test_name == "rustc") {
+            c_bootstraps += 1;
+            data.rustc.insertOne(make_data(date, header, times, true));
+
+            for (let timing of times) {
```
nrc: I'm a bit confused here - this seems to be an inlined version of `record_crate`, but that seemed to be called much later, so I'm not sure this is a safe move (it might well be, I just can't quite tell what is going on). Could you tell me what is happening please?

backend/perf.js
```diff
@@ -80,275 +79,285 @@ function init(repo_loc, callback) {
...
-    for (var j = 0; j < 3; ++j) {
-        var idx = Math.max(start_idx - 1, 0);
-        week_data.push(data[kind][idx]);
+    let start_idx = mk_start_idx(date, kind);
+    let weeks = [];
+    for (let j = 0; j < 3; ++j) {
+        let idx = Math.max(start_idx - j, 0);
+        weeks.push(data[kind][idx].by_crate);
     }
+    // let week = data[kind][idx].by_crate;
```
nrc: should this still be here?

backend/perf.js
```diff
@@ -80,275 +79,285 @@ function init(repo_loc, callback) {
...
 
-    prev_summary[kind] = cur_summary[kind];
+    return by_crate;
 }
 
 
 function make_data(date, header, times, rustc_crate) {
-    if (!header || !times || times.length == 0) {
-        return {};
-    }
```
nrc: Why remove this?

nrc:
OK, I think this looks good, thanks. There's a few comments inline. I'm not totally confident that the changes won't change anything though, especially as we don't have tests (my fault, sorry). Did you manage to test this locally?

backend/perf.js
Mark-Simulacrum: Either works, but the above makes it clear that it is an array. I would've used for-of syntax, but we need the index to update `date_components`. Let me know if you prefer for-in. 

I also think that at least to some extent, the JS community is moving away from for-in syntax in preference to for-of, c-style for, and for-of with Object.keys. There is also [this](http://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea) to consider, although most of those concerns are probably not applicable in this specific case.

backend/perf.js
Mark-Simulacrum: Yes, this is an inlined version of record_crate. I moved it into here since I believe this works: no code after this modifies `times`, which is the only dependency of `crate_list` near as I could tell. I can move it back though if we're uncertain.

It's worth pointing out that I also moved `record_phase` as well (currently line 150-155).

Mark-Simulacrum:
From visual-based testing, everything seemed equivalent. If you'd like I can try to run wget and diff the results before and after?

backend/perf.js
Mark-Simulacrum: From what I can tell, none of these were ever possibly true. We check [and skip with `continue`] for `times.length == 0`; and both `header` and `times` are accessed previously, so they could never be false. If we feel this is a good sanity check though I can remove this change.

Mark-Simulacrum:
Pushed fixes to inline comments and linter warnings.

Mark-Simulacrum:
I'm not sure why I failed to notice this before (possibly because I was testing with an outdated version of the input data), but the changes to indexing result in a massive difference currently: 2.8% on the summary page at perf.rust-lang.org and 11.4% in this branch, at least from what I can tell.

backend/perf.js
Mark-Simulacrum: Upon further work with the Rust port, I've encountered that moving `record_phase` isn't actually viable: the `/info` path depends on the `phase_list` containing "total" (at least, it contains it, I'm not sure that it's necessary for it to). We can either insert "total" into the phase list prematurely since it's not technically a phase, or I can revert the phase_list changes.

nrc:
Closing in favour of #22 as discussed on irc

Issue #21, opened 2016-08-03
Change to calculation of compile time percent increase/decrease


Mark-Simulacrum:
Currently, `((previous - current) / current) * 100.0` is the calculation used to derive the percent change between the previous median time and the current median time for the summary page. [This](https://en.wikipedia.org/wiki/Relative_change_and_difference#Percentage_change) page on Wikipedia implies that the formula should instead be `((current - previous) / previous) * 100.0`.

Wanted to open this for discussion since this change the results in my current branch from 11.4% increase in compile times to a 9.4% decrease.

PR #22, opened 2016-08-03
Port NodeJS backend into Rust.

Mark-Simulacrum:
Removes JS file and replaces backend directory with Rust port.

Fixes: #21.

Running list to keep track of major future, either in this PR or separate, work (@nrc, feel free to add):
 - Use liblog rather than println for logging.
 - Track down a good way of parsing dates passed from the client.
 - Tests.
 - Think about and possibly replace error-chain.
 - Rewrite the make_times function to return a Timings struct which will keep the `total` phase separate, since it is a meta-phase. This is a huge project since this affects most of the codebase; as well as causing problems with the iteration over phases code: it probably expects the total phase to exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+data
+rs-backend/target
--- a/rs-backend/Cargo.lock
+++ b/rs-backend/Cargo.lock
@@ -1,0 +1,438 @@
+[root]
+name = "rustc-perf"
+version = "0.1.0"
+dependencies = [
+ "chrono 0.2.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "error-chain 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "iron 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "json 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "persistent 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "router 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace-sys 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "dbghelp-sys 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-demangle 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace-sys"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "gcc 0.3.32 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "cfg-if"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "chrono"
+version = "0.2.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "num 0.1.34 (registry+https://github.com/rust-lang/crates.io-index)",
+ "time 0.1.35 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "conduit-mime-types"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "rustc-serialize 0.3.19 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "cookie"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "time 0.1.35 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "dbghelp-sys"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "error"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "traitobject 0.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "typeable 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "error-chain"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "gcc"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "hpack"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "httparse"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "hyper"
+version = "0.9.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cookie 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "httparse 1.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "language-tags 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "mime 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 0.2.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-serialize 0.3.19 (registry+https://github.com/rust-lang/crates.io-index)",
+ "solicit 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "time 0.1.35 (registry+https://github.com/rust-lang/crates.io-index)",
+ "traitobject 0.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "typeable 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicase 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "idna"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "matches 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-bidi 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-normalization 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "iron"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "conduit-mime-types 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "error 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper 0.9.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "modifier 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 0.2.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "plugin 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "typemap 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "json"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "kernel32-sys"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "language-tags"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "lazy_static"
+version = "0.1.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "libc"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "log"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "matches"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "mime"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "modifier"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "num"
+version = "0.1.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "num-integer 0.1.32 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num-iter 0.1.32 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num-traits 0.1.34 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "num-integer"
+version = "0.1.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "num-traits 0.1.34 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "num-iter"
+version = "0.1.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "num-integer 0.1.32 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num-traits 0.1.34 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "num-traits"
+version = "0.1.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "num_cpus"
+version = "0.2.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "persistent"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "iron 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "plugin 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "plugin"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "typemap 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "route-recognizer"
+version = "0.1.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "router"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "iron 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "route-recognizer 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "rustc-demangle"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "rustc-serialize"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "rustc_version"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "semver 0.1.20 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "semver"
+version = "0.1.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "solicit"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "hpack 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "time"
+version = "0.1.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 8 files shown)


backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+authors = ["Mark-Simulacrum <mark.simulacrum@gmail.com>"]
```
Mark-Simulacrum: I'm not sure what to put in here--let me know and I can replace it/change it.

backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "rs-backend"
```
Mark-Simulacrum: Rename to `rustc-perf`?

nrc: +1

backend/src/errors.rs
```diff
@@ -0,0 +1,13 @@
+error_chain! {
+    links {
+    }
+
+    foreign_links {
+        ::std::io::Error, Io, "io error";
+        ::json::Error, Json, "json error";
+        ::chrono::ParseError, Chrono, "chrono parsing error";
```
Mark-Simulacrum: I'm not sure that something so powerful as `error-chain` is needed here, or that I'm using it right (the strings especially seem "extra" to me since they're not really adding information.

backend/Cargo.toml
nrc: `["Mark-Simulacrum <mark.simulacrum@gmail.com>", "Nicholas Cameron <ncameron@mozilla.com>", "The rustc-perf contributors"]`

backend/src/errors.rs
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    errors {
```
Mark-Simulacrum: One of the next steps would be to go through all the potential error cases and give them a good error message, I guess. This would resolve #6, too, if I follow what that's asking for. Though resolving that may need work on the client side as well.

backend/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[dependencies]
+chrono = "0.2.22"
+error-chain = "0.2.2"
+json = "0.10.1"
+iron = "0.4.0"
+router = "0.2.0"
+persistent = "0.2.0"
+
+[profile.release]
+debug = true
```
nrc: probably don't want this in, right?

Mark-Simulacrum: Well, until we optimize the `ServerData::from_fs` function it's very slow (55 seconds to read the `data/processed` in) to run it in debug mode, so I'm thinking it may be necessary for discovering where errors occurred in production. I'm not sure what the "correct" way to do this is though (I've written little Rust code with proper error handling before now).

nrc: Should still get all the errors without compiling in debug mode, this just means we'll get better info in stack traces, etc. I would hope we don't need that and we can add this if needed later. OTOH, I guess we can leave it in as long as we remember to take it out in the future.

nrc:
Just at a first glance, it would be nice, I think, to break out at least a couple of modules - ideally main.rs should be just the `extern crate`s and a `main` function. You should probably leave the JS version there for the moment - it will be easy to remove later, but it might be interesting to run both for testing. Could you add some more comments please, particularly around the major data structures.

backend/Cargo.toml
Mark-Simulacrum: I think it's probably best to take it out now and re-add it if needed.

Mark-Simulacrum:
Regarding the JS version: should I leave the ES6 version (upon which this work was based, at least in part: although I compared with both along the way to make sure I wasn't missing anything) or the version in master right now?

nrc:
I guess the old version, since that is still our working reference.

Mark-Simulacrum:
Pushed the modularization and documentation. Should be at least ready for another review pass.

backend/src/load.rs
```diff
@@ -0,0 +1,418 @@
...
+        }
+
+        let last_date = last_date.expect("No dates found");
+
+        data_rustc.sort();
+        data_benchmarks.sort();
+
+        // Post processing to generate the summary data.
+        let summary_rustc = Summary::new(&data_rustc, last_date);
+        let summary_benchmarks = Summary::new(&data_benchmarks, last_date);
```
Mark-Simulacrum: I just did a test, and both of these complete in less than a second (in debug mode). Should we move them into the server logic, rather than splitting the summary into two steps? I don't see many advantages to doing so, but wanted to leave a comment noting that here.

backend/src/load.rs
```diff
@@ -0,0 +1,418 @@
...
+        })
+    }
+
+    pub fn by_kind(&self, kind: &str) -> &[Data] {
+        if kind == "rustc" {
+            &self.data_rustc
+        } else if kind == "benchmarks" {
+            &self.data_benchmarks
+        } else {
+            panic!("Unknown kind: {}", kind);
```
Mark-Simulacrum: TODO?: Encode kind as an enum, avoiding this check, and possibly improving the codebase overall.

backend/src/load.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+impl Ord for Data {
+    fn cmp(&self, other: &Data) -> Ordering {
+        self.date.cmp(&other.date)
+    }
+}
+
+impl Data {
+    // TODO: Document what rustc_crate means.
```
Mark-Simulacrum: I've tried to figure out exactly what it's meant to represent, but so far I've been unable to.

backend/src/server.rs
```diff
@@ -0,0 +1,501 @@
...
+        };
+        per_bench.insert("bootstrap".to_string(), bootstrap.into());
+
+
+        per_bench.into()
+    }).collect::<Vec<JsonValue>>();
+
+    Ok(object! {
+        "total_summary" => summaries.remove(0),
+        "total_breakdown" => breakdown.remove(0),
```
Mark-Simulacrum: I'm a little worried that the order in summary_* isn't guaranteed. (Order in JS may have been always the same due to an implementation detail, but that may not be true here: or even if it is, we should not depend on it).

Mark-Simulacrum:
I've also been thinking about the median computation for the summary page, and I'm not sure that it's beneficial: can you explain to me what the intended benefits are? I feel like it doesn't help to show the true by-week results, and skews the data by potentially never showing a week (especially if the week is an outlier!). I may be misreading something though.

backend/src/load.rs
nrc: We might want to take a global look at what code is where later, but for now I think this is fine.

backend/src/load.rs
nrc: sgtm

Mark-Simulacrum: Should we implement this now, or leave this as a "future work" type thing? Thoughts on this for other things?

nrc:
re the medians, it is the median of three commits at apx the same point in the week, not three weeks, so we should never miss a week. The benefit is meant to be that there is some noise in the results and if we don't take the median, we are more likely to see a difference due to noise rather than signal, hopefully using a median gets rid of some of the noise.

backend/src/load.rs
nrc: Either way is fine, I'd be happy to leave it till later

Mark-Simulacrum:
> it is the median of three commits at apx the same point in the week, not three weeks

I'm not sure I follow you here: Is there a guarantee that there are >=3 commits per week? If not, then the (at least current) logic takes the previous 3 data points, not necessarily within the current week:

```rust
// For a given date we'll get the three most recent sets of data
// and take the the median for each value.
let start_idx = start_idx(data, &date);
assert!(start_idx >= 3, "Less than 3 days of data");
let mut weeks = Vec::new();
for idx in 0..3 {
    weeks.push(&data[start_idx - idx].by_crate);
}
```

nrc:
It's not guaranteed, and at the moment we've been struggling with perf issues (ha! irony), but we were averaging three runs per day, so it would be highly unlikely to ever miss a week. We could add a guard against this in the code.

Mark-Simulacrum:
A guard sounds good to me. (And this can be a future work type thing, I think).

This sound good as to a specification?: Assert that all three data points we just pulled are within the current week. Drop any data points that are outside the current week (and log a warning?). Possibly log another warning when 0 data points are now left and skip the week if no data points for it were found.

Mark-Simulacrum:
Also, I've been looking at the rust-lang/rust#34831 issue (as a case of a pre-existing URL pointing at perf.rust-lang.org), and upon testing it with the current code, things break. I don't know of a good way to "best effort parse this date we're throwing at you" in Rust (and even in JS, truth be told). Without that, we'll either need to try and list all the cases we can think of (attempting parsing with each, failing, and moving on to the next one), or think of something else. Perhaps input sanitization? Regex-based search and recombine the input into something more readily parseable?

Let me know if you think we shouldn't care about backwards compatibility with URLs (I think we should). I'm also not 100% certain how to generate the [URL](http://perf.rust-lang.org/graphs.html?start=Mon%20Jun%2013%202016%2019%3A51%3A42%20GMT%2B0000%20(UTC)&end=Wed%20Jul%2013%202016%2020%3A26%3A57%20GMT%2B0000%20(UTC)&kind=rustc&crates=total&phases=LLVM%20passes%2Ctranslation&group_by=phase) in that issue, my attempts don't seem to produce the `HH:MM:SS GMT+0000 (UTC)` part.

```
Error(Msg("while parsing Mon Jun 13 2016 19:51:42 GMT+0000 (UTC)"), (Some(ParseError(TooLong)), stack backtrace:
...
```

backend/src/errors.rs
nrc: why do you use error-chain? I'd have thought it is unneccessary and we can just use simple errors. But if there is good motivation, then I'm ok with using it.

backend/src/util.rs
```diff
@@ -0,0 +1,37 @@
+use load::Data;
+use chrono::NaiveDateTime;
+
+/// Return where the passed date is (Ok(usize)) or should go (Err(usize)) in
+/// the data slice.
+fn get_insert_location(data: &[Data], date: &NaiveDateTime) -> ::std::result::Result<usize, usize> {
```
nrc: Given the return type, this should probably be called `find` or something like that

backend/src/errors.rs
Mark-Simulacrum: I think something simpler should work; maybe `quick_error!` or a hand-crafted basic implementation of the `From<OtherError> -> MyError` logic. I used `error-chain` in part because I sort of expected there to be more here (and for the `errors` declaration below to fill out, as per my comment); and I thought that error-chain seemed to be the way of doing that. I don't feel strongly either way about it; but I feel like the ergonomics of being able to use try! on pretty much any crate's error/result type are nice.

backend/src/util.rs
```diff
@@ -0,0 +1,37 @@
...
+    data.binary_search_by(|probe| probe.date.cmp(date))
+}
+
+/// Return the start index for an iterator from the passed date to the index
+/// returned by the companion function, end_idx.
+pub fn start_idx(data: &[Data], date: &NaiveDateTime) -> usize {
+    let idx = get_insert_location(data, date);
+    if idx.is_ok() {
```
nrc: Prefer match here

backend/src/util.rs
```diff
@@ -0,0 +1,37 @@
...
+        } else {
+            0
+        }
+    }
+}
+
+/// Returns the end index for an iterator from the start_idx() to this date.
+pub fn end_idx(data: &[Data], date: &NaiveDateTime) -> usize {
+    let idx = get_insert_location(data, date);
+    let idx = if idx.is_ok() {
```
nrc: ditto

backend/src/util.rs
Mark-Simulacrum: I was tempted, but `find` [to me] implies returning `&Item` (`&Data`) in this case; `position` could work, but since this can return the position where something should go, not where it is, I wasn't too sure.

Also, do you think it makes sense to return Result<usize, usize> here? Or would returning usize and doing the if is_ok(), unwrap(), else, unwrap_err() logic in here be better? We perform that logic anyway below near as I can tell.

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+
+use util::start_idx;
+use errors::*;
+
+const WEEKS_IN_SUMMARY: usize = 12;
+
+/// Contains the data loaded from the file system.
+/// Useful to avoid passing around each individual field, instead passing the
+/// entire struct.
+pub struct ServerData {
```
nrc: SO is this the data we read in from the GH repo? If so, then maybe `InputData` or `RawData` is a better name

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// A set containing all crate names.
+    pub crate_list: HashSet<String>,
+
+    /// A set containing all phase names, across all crates.
+    pub phase_list: HashSet<String>,
+
+    /// TODO: Better docs. It's unknown how this is used on the client.  A set
+    /// of test names. Test names are found from the file path, everything
+    /// before the `--` is included.
+    pub benchmarks: HashSet<String>,
```
nrc: Is there a reason to use HashSets for these fields rather than just a Vec?

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub last_date: NaiveDateTime,
+
+    /// Private due to access being exposed through by_kind method.
+    /// Care must be taken to sort these after modifying them.
+    data_rustc: Vec<Data>,
+    data_benchmarks: Vec<Data>,
+}
+
+/// Allows storing ServerData in Iron's persistent data structure.
+impl ::iron::typemap::Key for ServerData { type Value = ServerData; }
```
nrc: nit: needs some newlines

backend/src/load.rs
Mark-Simulacrum: Sounds good. I'll go with `InputData` since we do processing before it goes in here (summary construction, parsing the JSON, etc.).

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+
+impl ServerData {
+    /// Initialize ServerData from the file system.
+    pub fn from_fs() -> Result<ServerData> {
+        use std::path::Path;
+        use std::fs::{self, File};
```
nrc: Prefer to put non-glob imports at the top level, rather than in functions

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+        let mut c_benchmarks = 0;
+        let mut c_benchmarks_add = 0;
+
+        let mut data_rustc = Vec::new();
+        let mut data_benchmarks = Vec::new();
+        let mut benchmarks = HashSet::new();
+
+        // Read all files from repo_loc/processed
+        let mut file_count = 0;
+        for entry in try!(fs::read_dir(repo_loc.join("processed"))) {
```
nrc: I'd prefer using `?` to `try!` - means using nightly, but that seems fine to me

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+            match try!(file.read_to_string(&mut file_contents)) {
+                // Skip files whose size is 0.
+                0 => {
+                    no_times += 1;
+                    continue;
+                },
+                _ => {}
+            }
+
+            let contents = try!(json::parse(&file_contents));
```
nrc: We should recover from this error by continuing

backend/src/load.rs
Mark-Simulacrum: Perhaps incorrectly (and this hurts us later since json-rust doesn't define a From<HashSet<String>> impl for it's value type) I've always tried to use a "Set" for "sets". This also saves us having to check if the value exists in the set before inserting it (an O(n) operation, where as inserting into a set is ~O(1)).

Doesn't matter much to me.

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+        use std::fs::{self, File};
+
+        // TODO: Read this at runtime, not hardcoded.
+        let repo_loc = Path::new("../data");
+
+        let mut last_date = None;
+        let mut crate_list = HashSet::new();
+        let mut phase_list = HashSet::new();
+        let mut no_times = 0;
+        let mut c_bootstraps = 0;
```
nrc: Is this the same as `data_rustc.len()`? Likewise for `c_benchmarks`.

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+            let entry = try!(entry);
+            if try!(entry.metadata()).is_dir() {
+                continue;
+            }
+            file_count += 1;
+
+            let filename = entry.file_name().into_string().unwrap();
+            let mut file = try!(File::open(entry.path()));
+            let mut file_contents = String::new();
+            match try!(file.read_to_string(&mut file_contents)) {
```
nrc: This should be an `if`

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+
+            let test_name = &filename[0..filename.find("--").unwrap()];
+
+            let times = &contents["times"];
+            if test_name == "rustc" {
+                c_bootstraps += 1;
+                data_rustc.push(Data::new(date, &header, times, true));
+
+                for timing in times.members() {
+                    let crate_name = timing["crate"].as_str().unwrap().to_string();
```
nrc: What is going on here: `.as_str().unwrap().to_string()` ?

backend/src/load.rs
Mark-Simulacrum: This is the same near as I can tell, but `c_benchmarks` isn't (unless we don't care about tracking `c_benchmarks_add` separately). I'll remote `c_bootstraps`, not sure if the tracking on c_benchmarks_add is important.

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        // Post processing to generate the summary data.
+        let summary_rustc = Summary::new(&data_rustc, last_date);
+        let summary_benchmarks = Summary::new(&data_benchmarks, last_date);
+
+        println!("read {} files", file_count);
+        println!("found {} files without times", no_times);
+        println!("{} bootstrap times", c_bootstraps);
+        println!("{} benchmarks times", c_benchmarks);
+        println!("{} benchmarks times (appended)", c_benchmarks_add);
```
nrc: we should use liblog rather than println for logging (could be a follow-up)

backend/src/load.rs
```diff
@@ -0,0 +1,404 @@
...
+        } else if kind == "benchmarks" {
+            &self.data_benchmarks
+        } else {
+            panic!("Unknown kind: {}", kind);
+        }
+    }
+}
+
+/// The data loaded for a single date, and all associated crates.
+pub struct Data {
```
nrc: This really needs a better name, although I admit I cannot think of one.

backend/src/load.rs
Mark-Simulacrum: timing["crate"] is a [`JsonValue`](http://terhix.com/doc/json/enum.JsonValue.html); so to convert it to a string we need to take a `&str` out, unwrap that (it could be something else) and then convert it to a `String`.

This is one of the places where Serde might help... I'm still split on which one is better (Serde's ergonomics are terrible IMO if you don't use the Serialize/Deserialize auto-impls).

backend/src/load.rs
Mark-Simulacrum: Completely agree. `DateData` and `RunData` are two I've thought of (but I like neither much). I really want to get rid of "Data"...

backend/src/server.rs
```diff
@@ -0,0 +1,501 @@
...
+/// handler passed in. Specifically, it parses JSON data from the request if
+/// the request had greater than 0 length, and hands that into the handler.
+/// Post-processing consists of applying access control headers and [potentially]
+/// printing the error message.
+fn get_handler<F>(handler: F, req: &mut Request) -> IronResult<Response>
+    where F: Fn(Option<JsonValue>, &ServerData) -> Result<JsonValue> {
+    use std::ops::Deref;
+    use iron::headers::{ContentType, AccessControlAllowOrigin};
+    use iron::mime::{Mime, TopLevel, SubLevel};
+    use iron::modifiers::Header;
```
Mark-Simulacrum: > Prefer to put non-glob imports at the top level, rather than in functions
Do we want to follow that rule? All of these are very specific to this function (other than Deref).

backend/src/load.rs
```diff
@@ -0,0 +1,407 @@
...
+                    crate_list.insert(crate_name);
+                }
+            } else {
+                benchmarks.insert(test_name.to_string());
+                let index = data_benchmarks.iter().position(|benchmark: &Data| benchmark.date == date);
+                if let Some(index) = index {
+                    c_benchmarks_add += 1;
+                    let crate_name = times[0]["crate"].as_str().unwrap();
+                    data_benchmarks[index].by_crate.insert(test_name.to_string(),
+                        make_times(times, false).remove(crate_name).unwrap());
```
Mark-Simulacrum: I'm not sure what the original intended purpose of this code was. However, I noticed that this means we overwrite existing data in these cases. None of these are recent, so it may not be too big a deal, but wanted to make sure we were aware of this, since it might lead to data loss in the future as well.
```
hyper.0.5.0 in hyper.0.5.0--2015-05-25-14-31-35.json
helloworld in helloworld--2015-05-25-14-31-35.json
html5ever-2015-05-15 in html5ever-2015-05-15--2015-05-25-14-31-35.json
regex.0.1.30 in regex.0.1.30--2015-05-25-14-31-35.json
rust-encoding.0.2.32 in rust-encoding.0.2.32--2015-05-25-15-45-57.json
regex-macros.0.1.30 in regex-macros.0.1.30--2015-05-25-14-31-35.json
```

Mark-Simulacrum:
I'd like to add tests, but I've been struggling to come up with good isolated portions of the code which can be tested. We can definitely do some form of overall diff-like testing though I think (create a `InputData` with predetermined state and call the server functions to generate output). That shouldn't be too difficult as the server functions don't depend on a request being passed, just the input data and [optionally] a body.

backend/src/server.rs
```diff
@@ -0,0 +1,506 @@
...
+fn get_stats(req_body: Option<JsonValue>, data: &InputData) -> Result<JsonValue> {
+    let body = assert_request_body_present(req_body)?;
+
+    let kind = body["kind"].as_str().unwrap();
+    if kind != "rustc" && kind != "benchmarks" {
+        return Err(format!("bad value kind: {}", kind).into());
+    }
+
+    if kind == "benchmarks" && body["crates"].contains("total") {
+        return Err(format!("bad value for crates {:?} with kind benchmarks", body["crates"]).into());
```
Mark-Simulacrum: Any ideas as to why a "benchmarks" kind can't have "total" in the crates? I'm a little confused by this.

backend/src/server.rs
nrc: The iron stuff you could probably leave here - it has to be really specific for me to keep it at function scope, but it is a grey area and comes down to taste (reasoning is that for most things, if it is really a good idea to import into a function, then it is usually a better idea to split up the function and create a new module - modules are cheap!)

Mark-Simulacrum: Honestly, I would create a module, but I feel like that wouldn't be much different from this since it would mostly just contain this one function. As a future work item, I'd like to consider rethinking the get_handler function. That may also come with the github integration since I suspect it might conflict with the current way state is handled.

Going to leave this as-is for now; if this function grows to the point of pulling in other functions we can move it out into a module.

nrc:
Yeah, how we test this is a big question for me. I think in the medium term we should refactor to make portions testable, one piece at a time, and add unit tests - test coverage is a big goal for me for the rewrite.

In the short-term, that is probably too much to do. Some diff testing with real data from the timing repo would be good to do, if it is not too complex.

backend/src/util.rs
```diff
@@ -0,0 +1,32 @@
...
+pub fn end_idx(data: &[Data], date: &NaiveDateTime) -> usize {
+    match get_insert_location(data, date) {
+        Ok(idx) => idx,
+        Err(idx) => if idx < data.len() {
+            idx
+        } else {
+            data.len() - 1
+        }
+    }
+}
```
Mark-Simulacrum: Both `start_idx` and `end_idx` basically just clamp the value into the bounds of 0..data.len(); is there something in the stdlib better than what we do now?

nrc: Not that I know about, might be worth asking on #rust though

Mark-Simulacrum:
I think I can do some dirty testing with real data. Unfortunately, due to a few changes in logic (the percent update is the main one), we can't compare every page. I think I'll wait on the tests until we're blocking on them for merge if I can't come up with some good way of doing them quickly and cheaply. My current slow and semi-painful idea is to `Save as...` each page from a run of the JS version and do the same here and then diff them; but this grows rather quickly due to the semi-large number of variants for each page: `kind`, `group_by`, etc.

nrc:
It might be worth separating out any corrections into a separate commit, so we can be sure that the old version matches the new version without fixes, then add the fixes back in. Not sure if that is easy enough to be worthwhile.

Mark-Simulacrum:
Hmm, yeah, I'm not sure how easy that would be either. I _think_ the only major difference should be on the summary page (and this has so far proven true from visual testing); but I may be wrong. There may also be subtle changes that are more difficult to detect/fix: for example, right now I believe the dates on the summary page are off by one compared to the dates on perf.rust-lang.org. I'm not certain why this is, although I have suspicions that it has to do with my updating the data directory on my machine while the official one seems to have stopped auto-updating.

nrc:
I should have thought of this earlier, but we should move either the Rust or node implementation into a separate directory so we know what is what.

backend/src/errors.rs
nrc: We can land it and see if it pulls its weight. I've not used it before so I'm interested to see how it pans out.

Mark-Simulacrum:
I see a couple of options for that, not sure which is better, but I prefer either the first or the third personally. Creating a directory for a single file feels odd to me.

 1. Moving Rust into a rs-backend directory
 1. Moving Node into a node-backend directory
 1. Moving the [single file] node `perf.js` to top level and rename to `node-backend.js`.

nrc:
I like 1 or 2, so 1 seems like a plan we can both get behind :-)

backend/src/load.rs
nrc: Do you know if it ever happens that we get duplicate entries? Or could we just assume we'll get a set without testing for an entry existing?

backend/src/load.rs
nrc: Hmm, I don't recall either. I wonder if it is meant to coalesce data rather than overwrite it?

backend/src/load.rs
Mark-Simulacrum: I'm certain that for all three we get duplicate entries: phases are added into whenever we encounter a phase across all files, crates are added into on encountering a crate [and we test the same crates], benchmarks are also added repeatedly.

We don't use these values on the backend side near as I can tell other than returning them in the `/info` getter.

backend/src/load.rs
nrc: `TestRun` or `PerRun` or `PerRunData`, any of those or simialar seems fine to me (and better than Data).

backend/src/load.rs
nrc: I think that if it is true, it means the data is from a run on the bootstrap (i.e., the compiler compiling itself), rather than a benchmark

backend/src/load.rs
nrc: OK, then seems good to leave as a Set for now.

backend/src/load.rs
Mark-Simulacrum: Do we need to keep this around? We use this to determine whether to place "totals" as a fake crate into the by_crate field. We could do that unconditionally, and remove the argument. Computing totals is cheap (so not a performance concern IMO), and at least right now we don't benchmark a "totals" crate.

backend/src/server.rs
```diff
@@ -0,0 +1,518 @@
...
+    router.get("/info", |r: &mut Request| get_handler(get_info, r));
+    router.post("/data", |r: &mut Request| get_handler(get_data, r));
+    router.post("/get_tabular", |r: &mut Request| get_handler(get_tabular, r));
+    router.post("/get", |r: &mut Request| get_handler(get_days, r));
+    router.post("/stats", |r: &mut Request| get_handler(get_stats, r));
+
+    let mut chain = Chain::new(router);
+    chain.link(State::<InputData>::both(data));
+
+    Iron::new(chain).http("0.0.0.0:2346").unwrap();
```
nrc: Can you pull the ip addr into a const please?

backend/src/load.rs
nrc: SGTM

backend/src/server.rs
Mark-Simulacrum: Crate-level or module-level? I'm torn either way: server-specific but also semi-global.

nrc: I'd probably stick at the top of the crate to make it as easy as possible to change. We'll probably read it from a config file at some point, I guess.

nrc:
Awesome work, thank you!

PR #31, opened 2016-08-05
Switch summary comparison calls to compare weeks in the correct order.

### LINKED ISSUE #30: Summary page inconsistency between Rust and JS versions
#31 still doesn't bring exact compatibility with the JS in terms of the results of the summary page.
I'm not sure why this is (I've tried changing the percent logic to what I thought it was before, and a few more things), but nothing seems to bring us back even close to the old JS numbers.

Also, comparing the JS results and the Rust results makes me think a few more of these bugs are hiding from us: while the graph of compile times has dropped significantly, the summary page from the Rust version insists that compile times have gone up by ~13% last I checked while the JS page displays a ~55% drop (which feels more realistic to me when looking at the graphs).

Mark-Simulacrum:
Previously, the code compared a more recent week with an older week, returning
incorrect results. The code has been updated to flip this comparison order.

Fixes #30.

### DIFF
--- a/rs-backend/src/load.rs
+++ b/rs-backend/src/load.rs
@@ -415,12 +415,16 @@ impl Summary {
         // the previous week.
         let mut weeks: Vec<SummarizedWeek> = Vec::new();
         for window in medians.windows(2) {
-            weeks.push(compare_weeks(&window[0], &window[1]));
+            // We want to compare week 1 to week 0; since we want the change from
+            // week 1 to week 0.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Added another commit that brings the Rust implementation into equivalence with the JS summary page. Turns out the `rustc_crate` was actually needed, though I'm still not sure why we can't gather/store/use the totals for the benchmarks...

Editing description to close #30 since these commits resolve it.

nrc:
Thank you!

PR #33, opened 2016-08-10
[WIP] Rustification of the codebase

Mark-Simulacrum:
- Rewrite codebase using Serde.
 - Move POST handlers into using a trait instead of being defined as functions
   - This hurts searchability to some extent, but improves ergonomics greatly in my opinion.

Future work:
 - Move GET handlers
 - More rustification as a result of the migration to Serde.


rs-backend/src/server.rs
```diff
@@ -60,309 +142,287 @@ fn get_summary(_req_body: Option<JsonValue>, data: &InputData) -> Result<JsonVal
...
-        "date" => day.date.format(JS_DATE_FORMAT).to_string(),
-        "commit" => day.commit.clone(),
-        "data" => data
-    }
+    ObjectBuilder::new()
+        .insert("date", day.date.format(JS_DATE_FORMAT).to_string())
+        .insert("commit", day.commit.clone())
+        .insert("data", data)
+        .build()
 }
```
Mark-Simulacrum: I feel like the logic in this function is simplifiable even more, but I'm missing some crucial detail that will get us there.

nrc:
This looks great!

PR #34, opened 2016-08-10
More rustification

Mark-Simulacrum:
Reviewing each commit separately may be easier. Most of them don't build on each other that much.

### DIFF
--- a/rs-backend/src/server.rs
+++ b/rs-backend/src/server.rs
@@ -401,22 +401,23 @@ impl PostHandler for Stats {
 }
 
 fn mk_stats(data: &[&TestRun], crate_name: &str, phases: &[String]) -> Value {
-    let mut count = 0;
-    let mut first = 0;
-    let skip_list = data.iter().enumerate().map(|(i, d)| {
-        if d.by_crate.contains_key(crate_name) && !d.by_crate[crate_name].is_empty() {
-            if count == 0 {
-                first = i;
-            }
-
-            count += 1;
-            false
+    let sums = data.iter()
+        .filter(|day| if let Some(krate) = day.by_crate.get(crate_name) {
+            !krate.is_empty()
         } else {
-            true
-        }
-    }).collect::<Vec<_>>();
+            false
+        })
+        .map(|day| {
+            let krate = &day.by_crate[crate_name];
+            let mut sum = 0.0;
+            for phase in phases {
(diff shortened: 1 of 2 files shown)


rs-backend/src/server.rs
```diff
@@ -189,42 +259,37 @@ fn get_data_for_date(day: &TestRun, crate_names: &[String], phases: &[String], g
...
+        }
+    }
+
     let crates = crate_names.into_iter().filter_map(|crate_name| {
         day.by_crate.get(crate_name).map(|krate| {
             (crate_name, krate)
         })
     }).collect::<Vec<_>>();
 
     let mut data = HashMap::new();
```
Mark-Simulacrum: The new variant of the grouping code is much shorter. However, I'm not sure if it's entirely a win since there is a loss in clarity: instead of inserting after creating we now update in-place, since we always iterate over the same things. The order in which the two for loops are in no longer matters either.

rs-backend/src/server.rs
```diff
@@ -1,39 +1,95 @@
...
 use SERVER_ADDRESS;
 use errors::*;
-use load::{SummarizedWeek, Kind, TestRun, InputData};
+use load::{SummarizedWeek, Kind, TestRun, InputData, Timing};
 use util::{start_idx, end_idx};
 
 const JS_DATE_FORMAT: &'static str = "%Y-%m-%dT%H:%M:%S.000Z";
 
+// Boilerplate for parsing and responding to both GET and POST requests.
+mod handler {
```
nrc: I'd move this into its own file

nrc:
A quick scan of this looks pretty good to me. I think I'll probably do a pass of the code myself after merging as a PR since that is probably easier for both of us.

Thanks for the PR!

Mark-Simulacrum:
> I think I'll probably do a pass of the code myself after merging as a PR since that is probably easier for both of us.

What do you mean by this? Are you saying that you'll go through the code and either commit directly to master or put up another PR? (I'm fine with either).

nrc:
> What do you mean by this? Are you saying that you'll go through the code and either commit directly to master or put up another PR? (I'm fine with either).

Yes. Probably a PR so I can get your feedback unless there are only very minor changes

Mark-Simulacrum:
Sounds good.

Issue #29, opened 2016-08-05
Automatic update upon Github webhook receival


Mark-Simulacrum:
- [x] Add code to pull from GitHub
 - [x] Add a `/on_push` path to `server.rs` (unknown if GET or POST) which will update the state.
       - For now, we can reinitialize the entire state. Eventually, we'll want incremental updates.

PR #36, opened 2016-08-12
Added git module to update the repo at the specified path.

daytonallen:


Mark-Simulacrum:
Is there reasoning behind using `git fetch` instead of `git pull`?

rs-backend/src/git.rs
```diff
@@ -0,0 +1,47 @@
+use std::env;
+use std::path::Path;
+use std::process::Command;
+
+/// Fetches repo path from supplied command line argument, panics otherwise.
+pub fn get_repo_path() -> String {
+    if env::args().count() < 2 {
+        panic!("No argument supplied, needs location of data repo.");
```
nrc: Could you log a warning here and return a Result, rather than panicking

rs-backend/src/git.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+pub fn update_repo(repo_path: String) {
+    let mut working_dir = Path::new(&repo_path);
+    if env::set_current_dir(working_dir).is_err() {
+        panic!("Invalid path specified");
+    }
+
+    let fetch = Command::new("git")
```
nrc: I would issue a `git checkout master` first, just to make sure we are in the correct state

rs-backend/src/git.rs
```diff
@@ -0,0 +1,47 @@
...
+        panic!("No argument supplied, needs location of data repo.");
+    }
+    env::args().nth(1).unwrap()
+}
+
+
+pub fn update_repo(repo_path: String) {
+    let mut working_dir = Path::new(&repo_path);
+    if env::set_current_dir(working_dir).is_err() {
+        panic!("Invalid path specified");
```
nrc: Likewise, I'd warn + return rather than panic here

daytonallen:
That's the way it was done in the nodejs version so I just replicated, would it be better to refactor to use git pull instead?

rs-backend/src/git.rs
```diff
@@ -0,0 +1,47 @@
...
+            .arg("master")
+            .arg("origin/master")
+            .status()
+            .unwrap();
+
+        if merge.success() {
+            let checkout = Command::new("git")
+                .arg("checkout")
+                .arg("head")
+                .arg("--force")
```
nrc: Could you pull out some of these strings into `const`s please? It is probably worth factoring out the `Command` uses into a function too, since that seems to be repetitive code.

Mark-Simulacrum:
That's true. I'm not sure why the node version was written as such, though it may have been to ensure that no matter what the checkout happened.

daytonallen:
Thanks for the quick feedback guys! I'll refactor and resubmit a pull request.

nrc:
The node version was written using a library which only provided lower-level operations (`git pull` on the command line is actually not a primitive git operation, and libgit (at least the node version) doesn't provide it). For this version, I think using `pull` would be better, since all command line git clients support it.

rs-backend/src/git.rs
```diff
@@ -0,0 +1,59 @@
+use std::env;
+use std::path::Path;
+use std::process::Command;
+
+const CHECKOUT_OPTION: &'static str = "master";
```
Mark-Simulacrum: `CHECKOUT_BRANCH` is clearer to me.

Mark-Simulacrum: Or maybe even just `BRANCH`.

rs-backend/src/git.rs
```diff
@@ -0,0 +1,59 @@
...
+        return Err("No argument supplied, needs location of data repo.");
+    }
+    Ok(env::args().nth(1).unwrap())
+}
+
+pub fn update_repo(repo_path: String) {
+    let working_dir = Path::new(&repo_path);
+
+    if checkout_master(working_dir).is_ok() {
+        if pull_updates(working_dir).is_ok() {
```
Mark-Simulacrum: Return a result from `update_repo` and make these `?`; that'll make it easier to handle any print any potential errors in more detail.

rs-backend/src/git.rs
```diff
@@ -0,0 +1,59 @@
...
+    if let Ok(checkout) = Command::new(GIT)
+        .current_dir(working_dir)
+        .arg(GIT_CHECKOUT)
+        .arg(CHECKOUT_OPTION)
+        .status() {
+        if checkout.success() {
+            return Ok(());
+        }
+    }
+    Err("Checkout failed.")
```
Mark-Simulacrum: We use error-chain already, so we can just `use errors::*` at the top of this module and return the actual error here. I think you'll need to add `mod git` to the `main.rs` file to import errors. Once you've imported the errors module the return type should be `Result<()>`.

rs-backend/src/git.rs
```diff
@@ -0,0 +1,59 @@
...
+    if let Ok(pull) = Command::new(GIT)
+        .current_dir(working_dir)
+        .arg(GIT_PULL)
+        .status() {
+        if pull.success() {
+            return Ok(());
+        }
+
+    }
+    Err("Pull failed.")
```
Mark-Simulacrum: Same goes for here (use error-chain).

rs-backend/src/git.rs
```diff
@@ -0,0 +1,41 @@
...
+use errors::*;
+
+const BRANCH: &'static str = "master";
+const GIT: &'static str = "git";
+const GIT_CHECKOUT: &'static str = "checkout";
+const GIT_PULL: &'static str = "pull";
+
+
+pub fn get_repo_path() -> Result<String> {
+    Ok(env::args().nth(1).unwrap())
```
Mark-Simulacrum: We can replace the unwrap with `.ok()?` I think.

rs-backend/src/git.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+fn checkout_master(working_dir: &Path) -> Result<()> {
+    Command::new(GIT)
+        .current_dir(working_dir)
+        .arg(GIT_CHECKOUT)
+        .arg(BRANCH)
+        .status()?;
+    Ok(())
+
```
Mark-Simulacrum: nit: extra line

daytonallen:
The code changes in the load.rs file are just formatting changes. I had rustfmt enabled in my text editor and it auto formatted when I saved the files.

Mark-Simulacrum:
Can you revert the changes to load.rs? I think you ran rustfmt; but those changes are out of scope for this PR; though I think another PR with them would be much appreciated.

rs-backend/src/git.rs
daytonallen: No field name ok for Option<String>

nrc: `ok_or` or `ok_or_else`

Mark-Simulacrum:
Looks good to me. Thanks! Once this is merged, #29 will be almost done: just needs integration with updating the server state.

Mark-Simulacrum:
Forgot one thing, can you squash these commits into one?

daytonallen:
Could you advise me as how to do so? I've been googling around but I haven't found a solution that works.

Mark-Simulacrum:
`git rebase -i 29cb837e6c3df3a4bd8d62db10b1c93997ec0e54` (that's the commit SHA of your first commit) should show you all of your commits, like this:
```
pick 82f2e52 Added git module to update the repo at the specified path.
pick dc2d977 Refactored git module with suggested changes. Also, spawned child processes to update repo at specified path rather than changing the entire program's CWD.
pick 3c33683 Renamed CHECKOUT_OPTION to BRANCH
pick cc3e724 Implemented suggested error handling updates.
pick ed567b8 Revert "Implemented suggested error handling updates."
pick 628563b Reverted changes to load.rs, Removed extra line
pick 4f54846 Minor code tweak.
pick 975b9bb Refactored 'get_repo_path' to use ok_or instead of unwrap.
```

If you change all of the commits except the first one to start with "s" instead of "pick" it will squash them into the previous commit; once you've done that, you can check the commit message to make sure it summarizes the changes here, and commit it. Once you've done that you'll need to force push (`git push -f`)

If you want more interactive help, feel free to ping me on IRC (I'm in #rust-tools on irc.mozilla.org).

nrc:
Look's great, thank you!

nrc:
When rebasing, you can use `f` instead of `s` to squash whilst ignoring the commit message, which is useful if you only want the first commit's message.

daytonallen:
No prob, appreciate the guidance and advice!

PR #38, opened 2016-08-13
Adds the onpush hook into the server.

Mark-Simulacrum:
Implemented via pulling from GitHub, then reading all data from the
filesystem, and subsequently write-locking the data store, and updating
it. Locking the data store for as minimal a time as possible was the
primary intent; this should cause minimal downtime for users (since the
data store should only be write locked for a short time).

I'm not entirely happy with the clear everything out and reload, but wondering if it's worth working on changing that right now. Let me know if that's something we want implemented before this can move forward.

This also moves handler into its own file but that isn't directly related to the changes here; I can revert that change if needed.

### DIFF
--- a/rs-backend/src/handler.rs
+++ b/rs-backend/src/handler.rs
@@ -1,0 +1,64 @@
+//! Boilerplate for parsing and responding to both GET and POST requests.
+
+use std::ops::Deref;
+use std::io::Read;
+
+use serde;
+use serde_json::{self, Value};
+use iron::prelude::*;
+use iron::status;
+use persistent::State;
+
+use load::InputData;
+use errors::*;
+
+pub fn respond(res: Result<Value>) -> IronResult<Response> {
+    use iron::headers::{ContentType, AccessControlAllowOrigin};
+    use iron::mime::{Mime, TopLevel, SubLevel};
+    use iron::modifiers::Header;
+
+    let mut resp = match res {
+        Ok(json) => {
+            let mut resp = Response::with((status::Ok, serde_json::to_string(&json).unwrap()));
+            resp.set_mut(Header(ContentType(Mime(TopLevel::Application, SubLevel::Json, vec![]))));
+            resp
+        },
+        Err(err) => {
+            // TODO: Print to stderr
+            println!("An error occurred: {:?}", err);
+            Response::with((status::InternalServerError, err.to_string()))
+        }
+    };
+    resp.set_mut(Header(AccessControlAllowOrigin::Any));
+    Ok(resp)
+}
+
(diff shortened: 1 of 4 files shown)


nrc:
> I'm not entirely happy with the clear everything out and reload

I think the proper way to fix this is by moving to use a DB for the data.

rs-backend/src/main.rs
```diff
@@ -12,6 +12,7 @@ extern crate serde_json;
 
 pub const SERVER_ADDRESS: &'static str = "0.0.0.0:2346";
 
+mod handler;
```
nrc: this mod should have a more descriptive name - handler is a bit generic. hook_handler? or notification_handler or something.

rs-backend/src/server.rs
```diff
@@ -522,6 +456,30 @@ fn mk_stats(data: &[&TestRun], crate_name: &str, phases: &[String]) -> Value {
...
 }
 
+fn on_push(req: &mut Request) -> IronResult<Response> {
+    // FIXME we are throwing everything away and starting again. It would be
+    // better to read just the added files. These should be available in the
+    // body of the request.
+
+    println!("received onpush hook");
+
+    let mut responder = move || {
```
nrc: what is being moved here?

rs-backend/src/server.rs
```diff
@@ -522,6 +456,30 @@ fn mk_stats(data: &[&TestRun], crate_name: &str, phases: &[String]) -> Value {
...
+        git::update_repo(git::get_repo_path()?)?;
+
+        println!("updating from filesystem...");
+        let new_data = InputData::from_fs()?;
+
+        let rwlock = req.get::<State<InputData>>().unwrap();
```
nrc: I'm a bit confused here (in part I think because the names are all very generic - would be good to identify with in the name or a comment what data is being manipulated here). It looks like `data` is the main backing data for the program, in which case why do we `get` it from the `Request`?

rs-backend/src/handler.rs
```diff
@@ -0,0 +1,64 @@
...
+use iron::status;
+use persistent::State;
+
+use load::InputData;
+use errors::*;
+
+pub fn respond(res: Result<Value>) -> IronResult<Response> {
+    use iron::headers::{ContentType, AccessControlAllowOrigin};
+    use iron::mime::{Mime, TopLevel, SubLevel};
+    use iron::modifiers::Header;
```
nrc: Now we're in our own module, I'd move these imports out to module scope

nrc:
Thanks for the changes!

PR #39, opened 2016-08-14
Remove trailing whitespace in the HTML and JS files.

Mark-Simulacrum:
This is done to aid possible future edits (since they may accidentally introduce these unrelated changes).


nrc:
Thanks!

PR #41, opened 2016-08-15
Polishing the Rust backend

nrc:
Most changes are fairly minor. The only things that might be a bit controversial are replacing the Get/PutHandler traits by functions. I think this is an improvement since we get rid of some dummy structs and because the traits don't seem to add much to me.

r? @Mark-Simulacrum


rs-backend/src/load.rs
```diff
@@ -100,29 +109,12 @@ impl InputData {
...
-                let date_str = &filename[(filename.find("--").unwrap() + 2)..filename.find(".json").unwrap()];
-
-                // Ignore seconds: they aren't key to the data processing, and
-                // parsing them requires logic to replace them with 0 when
-                // they are absent.
-                NaiveDateTime::parse_from_str(date_str, "%Y-%m-%d-%H-%M")
-                    .map_err(|err| err.into())
-            }
-
             let header = InputHeader {
```
Mark-Simulacrum: Since we're polishing, this struct is actually entirely useless. I think keeping the commit and date as standalone variables makes more sense to me. If the struct is kept then `TestRun::new` should just take the header instead of the date and commit separately.

Mark-Simulacrum:
I think it might actually be beneficial to refactor the `from_fs` function into the filesystem gathering part and then a `new` function which takes `data_rustc` and `data_benchmarks`. It's not necessary, but I think it might make the code cleaner, and the changes shouldn't be too extensive.

PR #48, opened 2016-08-15
Fix errors and warnings introduced during addition of integration tests.

Mark-Simulacrum:
Sorry. No idea how this slipped past me.


nrc:
AH, but the sweet joy of having this caught by CI more than makes up for it :-)

Mark-Simulacrum:
That's very true.

PR #50, opened 2016-08-16
Code cleanup; debug logging.

Mark-Simulacrum:
Documents in code reason for #28; we can close that probably... I don't think there's really an issue with how the code works.

PR #51, opened 2016-08-16
Use UTC dates instead of NaiveDateTime.

Mark-Simulacrum:
Based on the other PR; I'm not sure GitHub will understand what's happening here. I tried to rebase, but there's conflicts there that I'd rather not fix [potentially] twice.

Note that this *changes the test cases* but all change are just dates shifting around due to using timezones.


nrc:
r+, but waiting on #50 before merging.

Mark-Simulacrum:
Will need test changes due to removing merging from #50.

PR #55, opened 2016-08-18
Declare structs for entire API in api.rs.

Mark-Simulacrum:
Excludes on_push handler due to unknown true request.

Modules for each API endpoint are created; declaring the
Request/Response relationship. Both requests and responses are
serializable and deserializable to allow tests to serialize requests.
Responses are currently never deserialized, but the derive is included
for consistency.

Also, let me know where/what you think needs more documentation.... I think I documented everything I thought was worthy of it, so if you think something needs more let me know.

r? @nrc

### DIFF
--- a/rs-backend/src/api.rs
+++ b/rs-backend/src/api.rs
@@ -1,0 +1,212 @@
+// Copyright 2016 The rustc-perf Project Developers. See the COPYRIGHT
+// file at the top-level directory.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Each API endpoint has its own module. The modules contain Request and/or
+//! Response structs; these contain the specifications for how to interact
+//! with the API.
+//!
+//! The responses are calculated in the server.rs file.
+
+pub mod summary {
+    use std::collections::HashMap;
+    use serde::{self, Serialize, Deserialize};
+
+    /// One decimal place rounded percent
+    #[derive(Debug, Copy, Clone)]
+    pub struct Percent(pub f64);
+
+    impl Deserialize for Percent {
+        fn deserialize<D>(deserializer: &mut D) -> ::std::result::Result<Percent, D::Error>
+            where D: serde::de::Deserializer
+        {
+            struct PercentVisitor;
+
+            impl serde::de::Visitor for PercentVisitor {
+                type Value = Percent;
+
+                fn visit_str<E>(&mut self, value: &str) -> ::std::result::Result<Percent, E>
+                    where E: serde::de::Error
+                {
+                    match value.parse() {
+                        Ok(value) => Ok(Percent(value)),
+                        Err(_) => Err(serde::de::Error::invalid_value(value)),
+                    }
+                }
+            }
+
+            deserializer.deserialize(PercentVisitor)
+        }
+    }
+
+    impl Serialize for Percent {
+        fn serialize<S>(&self, serializer: &mut S) -> ::std::result::Result<(), S::Error>
+            where S: serde::ser::Serializer
+        {
+            serializer.serialize_str(&format!("{:.1}", self.0))
+        }
+    }
+
+    // TODO: Deduplicate bootstrap counting for twice as much text; rephrase.
+    // TODO: Deduplicate 'First week in vector is the current week'.
+    // TODO: Come up with, document, and use terminology for referring to 13
+    //       weeks ago and 0 weeks ago.
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct Response {
+        /// The average summary across all benchmarks and bootstrap.
+        /// Bootstrap counts for twice as much that all other benchmarks do.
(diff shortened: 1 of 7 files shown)


Mark-Simulacrum:
Oh.. uh, the build is failing because `aster` is broken; not sure how much we can do about that... I fixed it locally.

Mark-Simulacrum:
Aster should be fixed after serde-rs/aster#101 is merged.

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
+// Types implement both serialize and deserialize despite conceptually needing
```
nrc: Could you add the copyright blurb please?

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
...
+// only one for testing.
```
nrc: I don't understand this comment - it doesn't seem to be true?

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// One decimal place rounded percent
+    type Percent = String;
+
+    // TODO: Deduplicate bootstrap counting for twice as much text; rephrase.
+    // TODO: Deduplicate 'First week in vector is the current week'.
+    // TODO: Come up with, document, and use terminology for referring to 13
+    //       weeks ago and 0 weeks ago.
+
+    #[derive(Serialize)]
```
nrc: structs should probably all derive Debug and Clone too

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
+// Types implement both serialize and deserialize despite conceptually needing
+// only one for testing.
+
+pub mod summary {
+    use std::collections::HashMap;
+
+    /// One decimal place rounded percent
+    type Percent = String;
```
nrc: Maybe worth using a newtype integer with a Display impl rather than a String alias

rs-backend/src/api.rs
Mark-Simulacrum: Well, it's mainly just poorly phrased: should really be "Type implement both serialize and deserialize despite only conceptually needing only one due to testing requiring serialization of requests". Not sure if it's really needed... I can edit it into a more thorough explanation though, or just remove it.

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
+// Types implement both serialize and deserialize despite conceptually needing
+// only one for testing.
+
```
nrc: Would be good to have a doc comment explaining the one mod per API request layout, etc.

rs-backend/src/api.rs
nrc: Oh, I see it is for mods other than Summary

rs-backend/src/api.rs
```diff
@@ -0,0 +1,167 @@
...
+        /// Which crates to return data for
+        pub crates: Vec<String>,
+
+        /// Which phases to return data for
+        pub phases: Vec<String>,
+    }
+
+    #[derive(Serialize, Deserialize)]
+    pub struct Response(pub Vec<Value>);
```
nrc: Why use Value here?

rs-backend/src/api.rs
Mark-Simulacrum: I think a display impl isn't enough: we'd have to implement serialize/deserialize ourselves. That's the main reason for doing this; aside from minor concerns related to having to export it (making summaries unique in that aspect of exporting something other than Request/Response).

nrc:
Looks great!

rs-backend/src/api.rs
Mark-Simulacrum: The main reason is that the helper function `get_data_for_date` returns a value currently. I plan to rewrite it at some point in the future (read: soon) to not do that, and then both of the value instances here will become nicer.

Mark-Simulacrum:
I'll fix the bugs/nits tomorrow (in my timezone), might even pull in the Value in a separate commit.

Can you respond to [this](https://github.com/rust-lang-nursery/rustc-perf/pull/55#discussion_r75249723) though?

rs-backend/src/api.rs
nrc: Yeah, I'd just remove it

rs-backend/src/api.rs
nrc: Hmm, yeah, of course. Still, I think it would be nice to implement deserialise so that we can verify that it fits our conception of a percentage - using a String seems kinda risky.

rs-backend/src/api.rs
nrc: ok, sounds good

oli-obk:
aster has been fixed

rs-backend/src/date.rs
```diff
@@ -0,0 +1,39 @@
+use chrono::{UTC, DateTime};
+use serde::{self, Serialize, Deserialize};
+
+#[derive(Debug, Copy, Clone)]
+pub struct Date(pub DateTime<UTC>);
+
+// TODO: Deprecate and replace with RFC3339 dates.
```
Mark-Simulacrum: Leaving this as a separate change/for a future PR since it requires changing the tests and checking all the frontend pages; with possible changes to frontend code.

Mark-Simulacrum:
Tests will not pass until Serde is updated to use latest aster. Otherwise, I believe this is ready for a review.

PR #58, opened 2016-08-19
Various changes throughout the codebase

Mark-Simulacrum:
Based on #55 as before; first actual commit on this branch is f85b860.

The following things have changed:
  - We pass and expect RFC 3339 dates from the client. The previous format is still supported (but considered deprecated).
  - There is no longer a distinction between finding the start index or the end index, see 02831f5's commit message for justification.
  - There is now a helper function for extracting ranges given a start and end date for a given kind.

r? @nrc


rs-backend/src/date.rs
```diff
@@ -4,14 +4,28 @@ use serde::{self, Serialize, Deserialize};
...
+                match DateTime::parse_from_rfc3339(s) {
+                    Ok(value) => Ok(Date(value.with_timezone(&UTC))),
+                    Err(_) => Err(()),
+                }
+            }
+        }
+    }
+}
+
 // TODO: Deprecate and replace with RFC3339 dates.
```
nrc: can remove this TODO, I guess?

nrc:
r+ with the TODO removed/updated

nrc:
Thank you!

PR #62, opened 2016-08-19
Add commit SHA into tooltips in graph display.

### LINKED ISSUE #61: Expose SHA1 hashes
Currently, if you hover on a data point, you get the date and time, but not the SHA1 hash of the commit. The SHA1 hash would be much more precise, and make it easy to determine if (e.g.) a given change is also on beta. For bonus points, it could be linked to the "diff" on GH versus the last bullet point, so one can easily browse the set of changes that took place in between.

Mark-Simulacrum:
Included an image of how it looks. Unfortunately, I don't believe there is any good way to allow selecting text from the tooltip.

image

Helps with #61, but an incomplete fix (no diff or link, not possible to click).

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -72,7 +72,7 @@ Chart.types.Line.extend({
             var steppedLabels = [];
             for (var i = 0; i < data.labels.length; ++i) {
                 if (i % step == 0) {
-                    steppedLabels.push(data.labels[i]);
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Compatible with existing JS and new Rust version.

PR #63, opened 2016-08-21
Show difference between commits when clicking on a point.

### LINKED ISSUE #52: Add units to graphs
Both http://perf.rust-lang.org/graphs.html and http://perf.rust-lang.org/mem.html lack units for the graphs.

### LINKED ISSUE #57: Graph display is misleading (8% regression is seen as much larger)
From IRC: https://botbot.me/mozilla/rustc/2016-08-18/?msg=71561193&page=3

### LINKED ISSUE #61: Expose SHA1 hashes

Mark-Simulacrum:
- Replace Chart.js with Highcharts. Highcharts is easier to work with, and is also used on http://rusty-dash.com/ (though that wasn't a big part of the decision).
 - Add label to y-axis. Fixes #52.
 - Always start y-axis at 0. Fixes #57.
 - The legend is now built-in to the graphs and can be manually toggled from there. I think it may be worth looking into moving the checkboxes directly into there, but there's problems (too many options) with that approach.
 - The new graph is x-axis zoomable (select a range with cursor), though I have not made this change the actual data queried from the server, nor update the URL.
 - The new graph more accurately represents the lack of data until recently.
 - Completely compatible with both old JS backend and new Rust backend.
 - Colors are no longer randomly chosen.

Fixes #61: clicking on a point on either the graphs or memory page will open a page on Github like [this](https://github.com/rust-lang/rust/compare/d3c3de8abe63f738113874267dad3b92a1965ecd...e64f68817d850ccbe642d7f067083bc655115d84).


Mark-Simulacrum:
New graph look:
image

Old graph look:
image

A couple things should be noted.
 - The legend is now built-in to the graphs and can be manually toggled from there. I think it may be worth looking into moving the checkboxes directly into there, but there's problems (too many options) with that approach.
 - The new graph is x-axis zoomable (select a range with cursor), though I have not made this change the actual data queried from the server, nor update the URL.
 - The new graph more accurately represents the lack of data until recently.
 - Completely compatible with both old JS backend and new Rust backend.

nrc:
Hmm, I think this is uglier, but more readable and functional, so basically a win.

PR #71, opened 2016-08-23
Frontend code cleanup

Mark-Simulacrum:
# Unused scripts:
Missed these in the UI update, didn't expect a page without a graph to load the graphing scripts.

It's possible they were there for caching purposes, if so I can re-add and change the graphing library to load the correct one. I don't see much point though, the site most definitely will not work without internet (since we dynamically load the data).

# JS looking up element with wrong ID:
Slight code duplication, but I felt that was better at least for now, as opposed to passing in the IDs


nrc:
Thank you!

PR #80, opened 2016-08-30
Use one base URL in shared.js for fetching

### LINKED ISSUE #77: Isolate the root URL for fetches
So we can change between the Rust and JS backends by changing one string rather than 7 in 7 different files.

Mark-Simulacrum:
Minimizes difference when using different backend servers.

Ideally, we'd abstract the usage of BASE_URL away (custom fetch function, perhaps) but that requires more work than this relatively simple PR.

Fixes #77.


nrc:
Thanks!

PR #81, opened 2016-08-30
Add 'Updated as of' text to the bottom of each page

### LINKED ISSUE #65: Show when last update of data occurred
Include a link to the data repository as a [manual] way to check if there has been updates since when the page was loaded/the server believes the latest update happened.

Mark-Simulacrum:
No effort is made to link to where the data is normally loaded from, or
to note who to notify if the reader believes there to be a problem.
While doing so is now simple, I don't believe it necessary to include in
the as-of text. This should instead be on something akin to an about
page.

With old JS backend, this will display "Updated as of: Invalid Date" instead of the correct text. This can probably be changed to something like "failed to load last date" if we want, but since we're working on phasing out the JS backend I'm not sure it's necessary.

This also sometimes makes a needless network request to get the `/info` endpoint, but optimization of network usage is not yet a priority goal of this project.

Fixes #65.


Mark-Simulacrum:
I guess technically this will show incorrect data if a tab is left open for long enough, but that doesn't seem like too big a problem. Easy to fix though if we want to.

shared.js
```diff
@@ -144,6 +144,22 @@ function make_settings(callback, total_label) {
     });
 }
 
+function make_as_of(callback) {
+    return fetch(BASE_URL + "/info", {}).then(function(response) {
+        response.json().then(function(data) {
+            document.getElementById("as-of").innerHTML = "Updated as of: " + (new Date(data.as_of)).toLocaleString();
+
+            if (callback) {
```
nrc: what is the callback for? It doesn't seem to be used

Mark-Simulacrum: Hmm, that's true. I kept it from the other function, make_settings, but we don't use it. Will remove.

nrc:
Thanks!

PR #82, opened 2016-08-30
Replace println! with liblog

### LINKED ISSUE #23: Add proper logging
`liblog` has been brought up by @nrc.

Mark-Simulacrum:
printlns were converted to reasonable log levels, but no new logging was
added (other than printing a message on server start since there's no
longer a wall of text by default).

More detailed logging is out-of-scope, and can be introduced as the codebase is worked on/later.

Fixes #23.

### DIFF
--- a/rs-backend/src/main.rs
+++ b/rs-backend/src/main.rs
@@ -7,11 +7,17 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+extern crate env_logger;
 extern crate rustc_perf;
 
 use rustc_perf::{load, server, util};
 
 fn main() {
+    env_logger::init().unwrap();
+
     let data = load::InputData::from_fs(&util::get_repo_path().unwrap()).unwrap();
+
+    println!("Starting server!");
+
     server::start(data);
 }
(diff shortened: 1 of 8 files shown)


rs-backend/src/main.rs
```diff
@@ -7,11 +7,17 @@
...
 extern crate rustc_perf;
 
 use rustc_perf::{load, server, util};
 
 fn main() {
+    env_logger::init().unwrap();
+
     let data = load::InputData::from_fs(&util::get_repo_path().unwrap()).unwrap();
+
+    println!("Starting server!");
```
nrc: should this be info?

Mark-Simulacrum: I wanted to print when we are starting the server even when no logging is explicitly enabled. info is nice, but is less than warn and then we're back to wall of output.

nrc: I usually prefer that silence => success, but I agree it is a useful notification

nrc:
thanks!

rs-backend/src/main.rs
Mark-Simulacrum: I prefer silence being success too, but due to the long boot up, I've found it's nice to have the "finished" notification (I see ~2 minute long loads).

PR #84, opened 2016-09-03
Serde update and frontend overhaul

Mark-Simulacrum:
Reviewing is likely best done by commit. I can split into multiple PRs if that helps.

Fixes #7 and #54; this is what the comparison view now looks like:
image

Also changes placement of "Submit":
Before:
image

After:
image

Stats view updated to be quite similar to the comparison view:
Before:
image
(and then some, couldn't fit it all on one screen)

After:
image

PR #86, opened 2016-09-05
Merge/overwrite runs by commit instead of by date

### LINKED ISSUE #53: Stop merging multiple runs with the same datetime into one
When loading runs from disk, we can find more than one run has occurred on a given date. When this happens, we ideally want to keep all of the data. I propose that we continue to merge data, but instead of doing so by date, we should do so by the associated commit. The commit is what gives each run "uniqueness"; the date is associated metadata. We should still sort both `data_` vectors by the date, in order to ease display/manipulation and allow binary search functionality.

See #50 and #28 for previous discussion.

Mark-Simulacrum:
Fixes #53.

Based on #84; don't merge before that one.


nrc:
r+. Needs a rebase, but feel free to merge with that

Mark-Simulacrum:
Merging in a different PR since cherry picking turned out to be easier.

PR #94, opened 2016-09-07
Ensure stable order on summary page

Mark-Simulacrum:


Mark-Simulacrum:
Merging myself since the change is relatively trivial, we can undo or change sort order later, though I think the default is reasonable (alphabetic).

Issue #60, opened 2016-08-19
Make it easy to go from front page to detailed info


nikomatsakis:
When you see a regression, it'd be great to be able to click on the test name and see more detailed data for just that test -- for example, the graph for total time on that benchmark and also the graphs for all individual compiler passes on that benchmark. This would probably be a simple step that would address #56 in an easier way. That is, if we could just click on the regression, we'd easily see if one particular compiler pass seems to have changed dramatically.

Mark-Simulacrum:
Only major item left here is being able to click on the total column, but this is mainly an issue in principle.

Discussion on IRC also brought up that determining which phase is which in the graph is difficult, filed #104 to fix that.

PR #95, opened 2016-09-07
Make cells on summary page link to time graph

Mark-Simulacrum:
Only the time graph is currently supported, since most of the time, it
is the one the user cares about.

The total column is not supported since it is rare that someone will
care about improvement over the past 13 weeks, but it should be
relatively simple to add it if requested; though the changes will not be
trivial.

The implementation currently queries the info endpoint for the phase
list to pass to the graphs page, but the ideal way to do this would be a
meta "all_phases" phase name we could pass instead.

r? @nrc

cc @nikomatsakis

Potentially fixes this issue: #60, potentially some other issues as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,23 +15,27 @@
     <script src="shared.js"></script>
     <script>
     make_as_of();
+    var info_data;
     fetch(BASE_URL + "/summary").then(function(response) {
         response.json().then(function(data) {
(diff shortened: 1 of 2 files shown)


nrc:
Looks good, thank you!

nikomatsakis:
Yay!

PR #97, opened 2016-09-07
Replace strings including \" with template string literals

Mark-Simulacrum:
This cuts down on line length and allows for string interpolation.

All modern browsers support template string literals, excluding IE 11
and and Safari 7.1 and 8. This is considered sufficient by the author.


nrc:
r+, but needs a rebase

PR #103, opened 2016-09-13
Remove the NodeJS backend

Mark-Simulacrum:
The Rust backend is now considered stable enough for official use.

Let me know if we want to rename the `rs-backend` directory to `backend`.


nrc:
> Let me know if we want to rename the rs-backend directory to backend.

Yeah, let's do that

nrc:
Awesome!

PR #107, opened 2016-09-14
Link to graph page for total column

### LINKED ISSUE #60: Make it easy to go from front page to detailed info

Mark-Simulacrum:
All columns now link to the graph page.

Fixes #60.

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,9 +62,10 @@
         for (var b of bench_keys) {
             html += "<tr><th>" + b + "</th>"
             for (var i = week_count - 1; i >= 0; --i) {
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Huzzah! Nice. 👍

PR #108, opened 2016-09-14
Move out tests into own files

### LINKED ISSUE #25: Move integration tests' expected outputs out into files
Edit: Integration and unit tests are now added for at least some of the functionality, main priority now is to move the integration tests to their own files. (This will ease checking that what we think changed changed).

----

Extracted conversation; primary point being that we'd like unit tests, but for them to be possible, modularization of the code is necessary.

---
Mark-Simulacrum:
> I'd like to add tests, but I've been struggling to come up with good isolated portions of the code which can be tested. We can definitely do some form of overall diff-like testing though I think (create a InputData with predetermined state and call the server functions to generate output). That shouldn't be too difficult as the server functions don't depend on a request being passed, just the input data and [optionally] a body.

nrc:
> Yeah, how we test this is a big question for me. I think in the medium term we should refactor to make portions testable, one piece at a time, and add unit tests - test coverage is a big goal for me for the rewrite.
> In the short-term, that is probably too much to do. Some diff testing with real data from the timing repo would be good to do, if it is not too complex.

Mark-Simulacrum:
Fixes #25.


Mark-Simulacrum:
Okay, I have a working theory: `26.70299727520436` and `26.702997275204358` will always parse/serialize to the first (might be the latter, not 100% sure), but the number we actually want is `26.702997275204355`... I'm not sure exactly why (in terms of underlying mechanics) this happens, though.

PR #112, opened 2016-09-19
Change tests to only use 2016 data

### LINKED ISSUE #110: Write a tool that auto-updates all of the integration tests given a set of changes
This is not required, but helps greatly when working on backend changes.

Mark-Simulacrum:
The basic code I tried to use to only acquire the 2016 data from July onwards (to keep the size down/equivalent to the old size, 40MB), after cd into the processed data directory; note that this may be incomplete, since I extracted it from bash's history log:
```bash
ls | grep -oP -- '.*--\d+-\d+-\d+' | sort | uniq | xargs -n1 -I{} sh -c "find . -name '{}*' | head -n1" > /tmp/tl2
find -size -500c | xargs -n1 rm
find . -name '*--2016-01-*' | xargs rm
find . -name '*--2016-02-*' | xargs rm
find . -name '*--2016-03-*' | xargs rm
find . -name '*--2016-04-*' | xargs rm
find . -name '*--2016-05-*' | xargs rm
find . -name '*--2016-06-*' | xargs rm
```

This currently fails on Travis for unknown reasons.

Fixes #110.


nrc:
How easy would it be to always round of floats to a smaller number of significant figures wherever they occur in the tests?

Mark-Simulacrum:
I believe that the only way to do this would be to wrap all our numbers in a unit struct and implement Serialize/Deserialize on that struct; there's no way to override number serialization otherwise.

An alternative might be to serialize into serde_json's `Value` struct and write/find code that will recursively descend through it, at which point we can change each number to the precision we want.

I've talked to one of the Serde devs, but we did not arrive at a good solution, though erickt hasn't yet gotten back to me.
```
[16:37:17] <erickt> simulacrum: float parsing is... hard
[16:39:02] <simulacrum> erickt: I know, there's an issue open for it already (by me); I just need a way to figure out what I should have
[16:39:29] <simulacrum> I actually still don't quite understand how this is happening because these numbers are being generated by serde
[16:39:57] <erickt> can that number be concretely represented by binary floats?
[16:40:54] <simulacrum> No idea; I don't know what the number is. Frankly, I don't really need that much precision; but I don't think I can tell Serde to "stop at 2 decimal points" or something like that
[16:44:04] <simulacrum> I think my problem is that I'm comparing f64 -> Value -> string -> serde_json::Value with f64 -> serde_json::Value
[16:48:21] <erickt> simulacrum: it's up to the serializer on how to serialize floats
[16:48:35] <erickt> there's no support for customizing decimal points
[16:49:18] <simulacrum> erickt: Thought so. Oh well, I guess I'll have to try harder. Is there a possibility that floats are serialized differently when using to_string vs. to_string_pretty?
[16:49:40] <simulacrum> Because I never had these problems before I started using pretty
[16:49:44] <erickt> that should be the same
[16:49:46] <erickt> hm.
[16:49:51] <simulacrum> Might just be me though
[16:50:21] <simulacrum> Let me test it, give me a minute...
[16:52:48] <simulacrum> Yeah I think it was just me. Is it possible the parsing is different? That would be odd..
[16:57:38] <simulacrum> erickt: Is there anyway to debug the serialization code to give me the actual number I should be putting into my source so that it parses back and forth correctly?
[17:07:12] <erickt> simulacrum: if you're not trying to validate that numbers are parsing correctly, you should probably stick to numbers that can be easily parsed
[17:07:45] <erickt> so like 2.5 can be precisely represented in f32/f64
[17:08:19] <simulacrum> Right, but I'm not sure how I can do that. Can I "do" something that does something to make the number easily parseable and serializable?
[17:09:07] <erickt> is this a test? can you point me to your code?
[17:09:49] <simulacrum> Give me a minute, yes
[17:11:41] <simulacrum> This is the branch with the broken tests: https://github.com/Mark-Simulacrum/rustc-perf/tree/new-tests
```

Mark-Simulacrum:
One thing that I still don't understand is how I was able to correct tests with 100% accuracy before we transitioned to loading them from separate files. I copy/pasted the output from Serde into a string, and it "just worked." I'm unable to come up with a theory as to how this is possible, since erickt said that serialization/deserialization doesn't care if the output is prettified or not.

dtolnay:
I see this in the test failure on Travis:

```diff
- "percent": 3.040540540540542,
+ "percent": 3.0405405405405419,
```

I see this in [backend/tests/data/processed/rustc--2016-09-13-09-24-31.json](https://github.com/Mark-Simulacrum/rustc-perf/blob/5a5a67bd69e06ea35cc30284547fe47385c8cb87/backend/tests/data/processed/rustc--2016-09-13-09-24-31.json#L162):

```json
"percent": 3.0405405405405412,
```

I see this in [backend/tests/expected_results/tabular_rustc.json](https://github.com/Mark-Simulacrum/rustc-perf/blob/5a5a67bd69e06ea35cc30284547fe47385c8cb87/backend/tests/expected_results/tabular_rustc.json#L16):

```json
"percent": 3.0405405405405419,
```

To show what is happening:

```rust
extern crate serde_json;

fn main() {
    let mut s = "3.0405405405405412".to_string();
    println!("{:?}", s);

    for _ in 0..3 {
        let f: f64 = serde_json::from_str(&s).unwrap();
        println!(" {}", f);

        s = serde_json::to_string(&f).unwrap();
        println!("{:?}", s);
    }
}
```

```
"3.0405405405405412"
 3.0405405405405412
"3.0405405405405414"
 3.0405405405405417
"3.0405405405405419"
 3.040540540540542
"3.040540540540542"
```

So (for whatever reason) the "expected" output expects that the data will be round-tripped twice but it is actually getting round-tripped three times. If you just want to fix the tests, you should be able to fix them by reading in the file with serde_json and writing it back out.

The float parsing bug you linked to is next on my list after Macros 1.1 stuff is settled.

> An alternative might be to serialize into serde_json's `Value` struct and write/find code that will recursively descend through it, at which point we can change each number to the precision we want. I don't know if Serde will lose precision if we go directly from our struct to Value. I suspect it might.

It will not lose precision there. The f64 in your struct and the f64 in the Value will be identical.

Mark-Simulacrum:
@dtolnay Thank you! That explanation also helped me understand why the previous copy/paste method worked: I wasn't serializing the server's responses again, because the results weren't pretty.

---

I don't understand why the tests are currently failing, though. Locally, I see them succeed.... going to do more testing.

Mark-Simulacrum:
Frankly, I'm baffled right now. The latest results show the following diff between the Travis' server response and our expected response; everything I've done so far cannot reproduce this. Perhaps something is wrong with Travis's cache? Can someone else test locally?
```diff
--- rustc-perf/backend/tests/expected_results/info.json	2016-09-23 21:16:12.000000000 -0600
+++ travis-info.json	2016-09-23 21:53:11.000000000 -0600
@@ -12 +11,0 @@
-    "piston-image-0.10.3",
@@ -15,2 +13,0 @@
-    "rust-encoding-0.3.0",
-    "rust-encoding.0.2.32",
```

Mark-Simulacrum:
Clearing caches entirely on Travis did not solve the issue, same diff result. Marking this PR as help wanted since I don't know what else to try--@dtolnay's local test seems to prove that it likely not a local element of my environment that is causing the difference.

Inspecting the test failure itself, I don't understand why or how those specific crates could be missing from Travis.

cc @nrc in case you have any thoughts on how to fix this, and to let you know that it's stuck at the moment

nrc:
So there is a difference in what crates are being pulled in on travis vs locally? Is this a separate issue from the floating point precision.

For the crates, it is worth checking if there are Cargo lock files checked in or being checked where they shouldn't be. It might be worth trying with the lock files checked in. This looks a bit like a difference between Cargo build on a clean directory vs Cargo update, then build on a non-fresh directory. I.e., the equivalent of Cargo update being run at different times. That should be addressed by checking in the lock files, but that has downsides. Perhaps we can change the scripts (on both sides) to do `cargo update && cargo build` or even add a `cargo clean` too.

Mark-Simulacrum:
This is almost(99%) certainly a different issue from floating point precision.

I'm not sure I follow you: the diff I posted above is what the difference between what Travis is apparently seeing in the test data vs. what we see locally, which shouldn't be affected what dependencies rustc-perf code itself has, and as such `cargo {clean,update}` won't matter in that case.

I already tried clearing out the Travis cache, so Travis *should* have had a clean build.

nrc:
> the diff I posted above is what the difference between what Travis is apparently seeing in the test data vs. what we see locally

OK, then I don't understand :-) This is the *input* to the test data? Or the result of testing?

Mark-Simulacrum:
@nrc I believe this is ready for review, everything looks good other than having ignored the `/info` test for now. I'll file a PR changing that once this is merged and ask for @alexcrichton's help there.

nrc:
Thanks for struggling though this :-)

PR #114, opened 2016-09-26
Display only the current series in graph tooltips

### LINKED ISSUE #104: Show which line is which phase/crate in the graph views
Perhaps by sorting the "current" one to the top and bolding it in the tooltip.

Mark-Simulacrum:
It is hard to tell which data point is the current one when many series
are displayed, so we should only display one series in the graph
tooltip. Relative comparisons more so than exact comparisons are
generally made on the graph page, and exact data often varies from
run-to-run even without changes to the compiler, so this change should
not hurt usability overall, and in fact should improve it.

Potential future work could involve finding a way to query the previous
and next series in a given "column" on the graph, and display the series
data for those two data points.

Before:
image

After:
image

Fixes #104

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -71,16 +71,12 @@ function init_graph(data, series, field, total_label, yAxisLabel) {
         },
         series: datasets,
         tooltip: {
-            shared: true,
             crosshairs: [true],
             formatter: function () {
                 var date = new Date(this.x);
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
cc @nikomatsakis

PR #115, opened 2016-09-29
De-ignore info integration test

Mark-Simulacrum:
This *should* fail Travis.

Follow up of #112.

cc @alexcrichton, @nrc says that you are familiar with Travis, can you help us debug this failure? It doesn't fail locally.

### DIFF
--- a/backend/src/api.rs
+++ b/backend/src/api.rs
@@ -66,17 +66,18 @@ pub mod summary {
 
 pub mod info {
     use date::Date;
+    use std::collections::BTreeSet;
 
     #[derive(Debug, Clone, Serialize, Deserialize)]
     pub struct Response {
         /// Sorted vector of crate names (all compiler crates)
-        pub crates: Vec<String>,
+        pub crates: BTreeSet<String>,
 
         /// Sorted vector of phase names.
-        pub phases: Vec<String>,
+        pub phases: BTreeSet<String>,
 
         /// Sorted vector of benchmark (non-compiler crates) crate names.
-        pub benchmarks: Vec<String>,
+        pub benchmarks: BTreeSet<String>,
 
         /// Chronologically last loaded run date.
         pub as_of: Date,
--- a/backend/src/load.rs
+++ b/backend/src/load.rs
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, BTreeSet};
 use std::cmp::{Ordering, max};
 use std::fs::{self, File};
 use std::path::PathBuf;
@@ -30,15 +30,15 @@ pub struct InputData {
     pub summary_benchmarks: Summary,
 
     /// A set containing all crate names of the bootstrap kind.
-    pub crate_list: HashSet<String>,
+    pub crate_list: BTreeSet<String>,
 
     /// A set containing all phase names, across all crates.
-    pub phase_list: HashSet<String>,
+    pub phase_list: BTreeSet<String>,
 
     /// A set of test names; only non-bootstrap benchmarks are included. Test
     /// names are found from the file path, everything before the `--` is
     /// included.
-    pub benchmarks: HashSet<String>,
+    pub benchmarks: BTreeSet<String>,
 
     /// The last date that was seen while loading files. The DateTime variant is
     /// used here since the date may or may not contain a time. Since the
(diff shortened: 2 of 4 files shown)


Mark-Simulacrum:
Okay, this has failed Travis. Re-pasting diff:
```diff
--- rustc-perf/backend/tests/expected_results/info.json 2016-09-23 21:16:12.000000000 -0600
+++ travis-info.json    2016-09-23 21:53:11.000000000 -0600
@@ -12 +11,0 @@
-    "piston-image-0.10.3",
@@ -15,2 +13,0 @@
-    "rust-encoding-0.3.0",
-    "rust-encoding.0.2.32",
```

alexcrichton:
Hm sorry I don't have a lot of context here, but some ideas off the top of my head:

* Does a fresh clone pass tests for you locally?
* Maybe a Cargo.lock isn't checked in and something's changing?
* Perhaps a dirty build directory?

Mark-Simulacrum:
> Does a fresh clone pass tests for you locally?

Yes, and @dtolnay also verified that this passes locally.

> Maybe a Cargo.lock isn't checked in and something's changing?

Cargo.lock should not affect this, but either way, independent verification and my own testing within docker containers should make this impossible.

> Perhaps a dirty build directory?
Checked in docker (with no mounted volumes, just cloning), so unlikely.

alexcrichton:
I was able to reproduce this locally with the latest nightly build on a fresh checkout, but it still doesn't reproduce locally for you?

dtolnay:
I tried it again. In a fresh checkout of 339f1b5b1f433961589192f006abd587ed0e1507, the Travis build steps:

```sh
cd backend
cargo fmt -- --write-mode=diff
cargo build --verbose
cargo test --verbose
```

do not reproduce the failure for me.

alexcrichton:
Sorry I may not be much help, I'm not sure I have enough context to help out here :(

Mark-Simulacrum:
@alexcrichton Okay, I just reconfirmed, and I am also unable to reproduce the Travis failure locally, version information included below. Can you tell me what your environment is? I'm confused why Travis and you are seeing failures, but both @dtolnay and myself are not.

Also, I don't know if this even has potential to work, but post-clone:
```
cd backend
cargo clean
cargo fmt -- --write-mode=diff
cargo build
cargo test
cd ..
tar cjvf backend.tar.bz2 backend
md5sum backend.tar.bz2
=> d66ffa68b1ea8dd8ae635749a6002df0  backend.tar.bz2
```

Is this md5 equivalent for you as well?

```
rustc 1.14.0-nightly (289f3a4ca 2016-09-29)
binary: rustc
commit-hash: 289f3a4ca79916d6445b452fc19a18a1e42a879a
commit-date: 2016-09-29
host: x86_64-unknown-linux-gnu
release: 1.14.0-nightly
```

alexcrichton:
[This container](https://gist.github.com/alexcrichton/7ac96e5b6b3233e2ddca89540c3b57cf) will reproduce the problem.

dtolnay:
Nope, does not reproduce it for me. `docker build` downloads everything, clones the repo, builds, and the tests pass.

Mark-Simulacrum:
@alexcrichton Unfortunately not; my results for that container are still that all tests pass. This is one odd bug!

nrc:
Could this be platform-specific? I see that @Mark-Simulacrum is on Linux from one of the comments, I assume @alexcrichton is on OS X? Did Travis fail on all platforms?

dtolnay:
I think Travis only ran (and failed) on Linux. I am using Linux and it passed.

The most confusing thing so far is how the same Dockerfile fails for Alex but passes for Mark and me. I am only involved to the extent that serde_json is involved so I have not looked at what the failing test is doing, but is it downloading data from somewhere? Is the server sending us different data? What else could cause it to behave differently in Docker? The Dockerfile is not copying in anything from the host.

dtolnay:
I built the docker file on a mac laptop and it worked.

Mark-Simulacrum:
re: @dtolnay 
> I have not looked at what the failing test is doing, but is it downloading data from somewhere? Is the server sending us different data? What else could cause it to behave differently in Docker?

Data is not being downloaded, the test fails because the server is sending us different data--but we spin up that server (it's the backend that we're testing). I don't see why it would behave differently within Docker.

> Could this be platform-specific?

It's possible, but as far as I know Travis is linux-based as well; we have only configured Travis to test on linux.

I have also built the docker container on a mac laptop (running sierra, in case it matters), all tests pass.

Mark-Simulacrum:
After ~28 failed builds, this has now finally become green once more.

PR #119, opened 2016-10-06
Display all known benchmarks on summary page regardless of build status.

Mark-Simulacrum:
My belief is that this state is not ideal; these changes make it so that we display crates we've replaced, like piston-image-0.3.11, on the summary page. These haven't built in a long time, but we have data on them somewhere, so we display them.

new summary page

I think merging this as-is isn't a good idea, but I'm uncertain how to fix these problems correctly; largely putting this up for async discussion.

In fixing this, I believe I now know why we aren't seeing syntex or syntex-incr-clean on perf.rlo's summary page. The summary is as such: we need at least two weeks of data before we start showing things. Specifically, because we summarize weeks by iterating over week0's crate keys, we never see the syntex benchmark because we don't have 2 weeks of data on it (yet).

This is mostly also true for the incr-clean variant, except there we haven't deleted data, the data is simply not being processed.

This is illustrated by all of these files being empty (month of October), despite data being in the raw variants of these files. I have not yet looked into exactly why the processing fails.
```
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-05-00-45-25.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-04-18-05-37.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-04-11-25-48.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-04-04-45-58.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-03-22-06-31.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-03-15-26-55.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-03-08-44-48.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-03-02-03-40.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-02-19-23-41.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-02-12-43-43.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-02-06-03-30.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-01-23-23-33.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-01-16-43-30.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-01-10-04-03.json
-rw-rw-r-- 1 mark mark 0 Oct  5 09:27 syntex-0.42.2-incr-clean--2016-10-01-03-06-19.json
-rw-rw-r-- 1 mark mark 0 Oct  5 19:51 syntex-0.42.2-incr-clean--2016-10-05-14-05-44.json
-rw-rw-r-- 1 mark mark 0 Oct  5 19:51 syntex-0.42.2-incr-clean--2016-10-05-07-25-36.json
```


nikomatsakis:
I definitely want to see all the rows, even if they don't have a lot of data -- but it *would* be nice to hide things that have been intentionally replaced. Can we extend the data w/ something to indicate rows that we should intentionally hide?

nikomatsakis:
To elaborate on why I'd prefer to see all rows: I find the current display pretty confusing in that it doesn't give much indication about the status of newly added tests. It also seems very easy for something to stop building and kind of silently disappear, though this may have been addressed separately. But basically I feel like using "no data" as a heuristic for "should be hidden" feels suboptimal. We should decide explicitly to retire things, and for everything else I would at least expect to see a row so that I know the benchmark *is* "registered".

Mark-Simulacrum:
One option is to add something like a `config.json` which would allow specifying which crates to ignore (and potentially more than that). This file could be checked in to git. Another option would be to, when retiring/updating a test, move its data to rustc-timing-archive.

I think both options have their merits. The first option is slightly worse, I think, since it requires us to maintain configuration for something that shouldn't really need configuration (at least not yet).

Mark-Simulacrum:
See #120 about some of my more wide-sweeping thoughts on this.

nrc:
My feeling is that we should keep the summary page as small as focussed as possible. I like the big number more than I like the table and the table is already a little bit fiddly. Perhaps we could keep the front summary page small, and add another page which is the summary table but with all possible rows? We'd then have a page where we could add more summary-style data without crowding the front page.

nikomatsakis:
> Perhaps we could keep the front summary page small, and add another page which is the summary table but with all possible rows? 

I agree about keeping the front page focused, but I'm not quite sure what tests you are envisioning excluding on a permanent basis. Is it the deprecated tests, or something else too?

I see three categories of things missing (and I could definitely be wrong on these points):

- broken tests seem to silently vanish sometimes
- new tests without much data yet
- deprecated tests

I agree that the third ones don't belong on the front page, if we need them anywhere at all, but not the first two.

Mark-Simulacrum:
With latest changes:

image

backend/src/load.rs
```diff
@@ -407,99 +407,114 @@ fn make_times(timings: &[Value], is_rustc: bool) -> HashMap<String, HashMap<Stri
...
-            let start_idx = index_in(data, date);
-            let mut weeks = Vec::new();
-            for idx in 0..3 {
-                if let Some(idx) = start_idx.checked_sub(idx) {
-                    if let Some(ref day) = data.get(idx) {
-                        weeks.push(&day.by_crate);
-                    }
-                }
-            }
+        // let mut medians: Vec<MedianRun> = Vec::new();
```
nrc: should prob be removed

nrc:
r=me with the comment removed

PR #126, opened 2017-04-17
New infrastructure

Mark-Simulacrum:
This pull request is impossible to review as a whole, and even the individual commits are quite large. I'll summarize what is happening, though:

 - Changed from using iron to hyper's new tokio-based implementation.
 - The hyper/iron server now serves all of the HTTP requests, not just the API's, but listens on the same port. This may mean that the proxy pass configuration needs updating to pass all requests through to the backend (essentially just removing the `/perf` filter, probably, though I'm not well versed in that area).
 - Updated loading code to at least be compatible with the new collection infrastructure (which still lives out-of-tree and in Mark-Simulacrum/rustc-perf-collector, but will be moved over here or into another nursery repository eventually).
 - Updated front end:
   - We now display time delta on the summary page, which currently means that it's practically empty (no significant changes over the past ~12 weeks).
   - Various pages have been cleaned up to coincide with the fact that we no longer collect bootstrap timings, so that distinction is irrelevant, and as such has now been removed.

With these changes, I've also changed how the repository itself is structured. We now have a traditional Cargo project layout, with `src` being in the root of the repository, instead of in `backend`. `static` holds all the JS, HTML, CSS, and font files.

Let me know if I can explain anything!


Mark-Simulacrum:
I think that merging this now seems like a good idea. We can rework portions later. The primary concerns that have been brought up so far are around keeping the old data alive, but I don't consider it entirely necessary and think that it might be better to write a script to convert the old data to new data instead. I also want to avoid having everything new simply pile into here (since that doesn't bring any advantages anyway, and only makes those changes effectively unreviewable).

Issue #132, opened 2017-05-16
Add endpoint for viewing performance change from one nightly to another


jonhoo:
Following a [tweet](https://twitter.com/nick_r_cameron/status/864327531935154177) by @nrc, there are now a number of Twitter bots tweeting out information about the latest nightly. While I was writing [one myself](https://twitter.com/rust_at_sunrise), I realized that it'd be really cool to also integrate it with perf.rust-lang.org.

A good first step would be the ability to look up the perf change for a given nightly. Something like http://perf.rust-lang.org/compare.html, except where you can provide `?nightly=<date>` in the URL to show the diff between that nightly and the previous benchmark result. I initially thought that `compare.html` could already do this, but it appears that that's not the case (changing the query parameters does not do anything). It'd also be fine to use revision instead of date in the URL, though it would be slightly more painful to work with.

If we wanted to be even fancier, we could also expose a JSON API endpoint that gives summary information about the performance change. That way it could be directly included in these announcements. Though I realize that that'd be a bunch more work.

-----

Following some digging around for such an endpoint, it appears I may have crashed perf.r-l.o :/
I did find http://perf.rust-lang.org/perf/get in the process though, which looks like it is a little bit like one of these endpoints. But it's not clear how to generate the right arguments to it?

Mark-Simulacrum:
The query parameters should work. I thought they did; that's a separate issue -- I'll look into it. /perf/get will take the arguments given here: https://github.com/rust-lang-nursery/rustc-perf/blob/844995579d0c3023f5b7754f3f0d58ba57756132/src/api.rs#L153-L170; let me know if there's anything wrong with that. Don't worry if you crash perf.rlo, it's easy to restart it. Just ping acrichto or myself in IRC (#rust-internals or #rust-infra is fine).

jonhoo:
Ouch, that's not a great endpoint. Having to JSON encode every parameter :/ But I guess it's fine as a starting point, thanks!

Mark-Simulacrum:
Yeah, it's possible that we'll soon/eventually restructure this into using non-JSON parameters, but I'm uncertain if/when that'll happen.

jonhoo:
I ended up just directly using `rustc_perf`, which works pretty well: https://github.com/jonhoo/rust-at-sunrise/blob/master/src/main.rs#L164. Won't be quite as easy for non-Rust clients though.

pnkfelix:
This seems like it might be mostly resolved now?

[...]

PR #133, opened 2017-05-17
Expose fields on Recording

jonhoo:
For consumers of the perf API, the fields of `Recording` are vital to actually extract any results (e.g., #132).

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -97,6 +97,14 @@ impl Recording {
             rss: 0,
         }
     }
+    
+    pub fn get_time(&self) -> f64 {
+        self.time
+    }
+    
+    pub fn get_rss(&self) -> u64 {
+        self.rss
+    }
 
     fn record(&mut self, phase: Option<&Pass>) {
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Could you instead just make the fields `pub`? I don't really see the point of the wrapper structs in this case. Also, be aware that depending on perf.rlo is inherently unstable -- the API and internals may change at any time.

jonhoo:
I decided to add accessors instead of making the fields `pub` here because this is one of the few structs in the codebase (as far as I could tell) where you seem to want people to update the fields using the `record` method, and not by directly updating the fields. Happy to make them `pub` instead though.

As for API stability, that's fine. It would be nice to eventually expose a relatively stable API (even if it's just a JSON API with a stable format), but for now I don't mind keeping track.

Mark-Simulacrum:
Thanks! Yeah, the API for this struct isn't very well formulated, pub I think makes sense.

PR #143, opened 2017-08-27
Changes for the `perf` branch of rustc-perf-collector

alexcrichton:
This is what's currently deployed at https://immense-ocean-97066.herokuapp.com/ and wanted to make sure I put this up!

The code currently basically just duplicates a few API paths so this isn't landable, but I'm curious to hear your thoughts on how to handle the old pass-based infrastructure. Delete it? Keep it? Migrate everything?


Mark-Simulacrum:
Since we're moving off of EC2 for measurement, I think it makes sense to delete all old data and remove time-passes as a method of collection. It's not really all that useful today, and has a significant contribution to code complexity. I believe currently that perf.rlo's primary purpose is to inform that a regression exists, and then have humans diagnose what the problem is specifically. This requires local work anyway, so time-passes isn't all that helpful; especially since in most cases we can already blame a specific PR immediately.

alexcrichton:
Hm yeah excellent points!

Mark-Simulacrum:
Okay, I've merged this into the perf branch manually and also imported collector into this repo (under that branch). That way we can more uniformly work on future patches without having separate PRs to each repo.

PR #153, opened 2017-09-29
 Add benchmarks of regex in release mode

### LINKED ISSUE #140: Bench optimized builds
Looking at this repo and rustc-benchmarks, it doesn't look like there is any benchmarks done of optimized/release mode builds. (Or am I missing something?)

Wouldn't it make sense to check the performance of this as well?

oyvindln:
This adds benchmarks of regex using the --release flag.
I also did some small changes to the documentation as I hit some snags.
Let's see if I did this right..

Fixes #140


Mark-Simulacrum:
Instead of `regex-opt-version-X` can you rename the directories to `regex-opt-X`? And then squash the two regex-related commits together? I think we'll be good to go then.

oyvindln:
@Mark-Simulacrum Updated as you requested.

Mark-Simulacrum:
Thanks!

michaelwoerister:
🎉

PR #154, opened 2017-10-02
Use CARGO_INCREMENTAL instead of RUSTFLAGS for regex benchmarks.

michaelwoerister:
I think this is the more realistic test case.

cc @nikomatsakis 
r? @Mark-Simulacrum

### DIFF
--- a/collector/benchmarks/regex-0.1.80/makefile
+++ b/collector/benchmarks/regex-0.1.80/makefile
@@ -1,72 +1,68 @@
-.PHONY: all \
-	    all@010-baseline \
-	    all@020-incr-from-scratch \
-	    all@030-incr-change-private-fn
-
-INCREMENTAL_FLAGS=-Z incremental=incr
+.PHONY: all@010-baseline \
+        all@020-incr-from-scratch \
+        all@030-compile_one \
+        all@040-is_valid_cap_letter \
+        all@050-expand \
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
I'm not opposed to doing this, though it will likely introduce a noticeable jump in data collected after this (and we can't really back-collect since CARGO_INCREMENTAL hasn't always existed).

r=me if we're fine with not back-collecting the data, otherwise we can discuss further.

cc @alexcrichton

michaelwoerister:
Since CARGO_INCREMENTAL was introduced in January, back-collecting a few months should not be a problem, right?

Mark-Simulacrum:
Wasn't aware it was introduced so far back. Mind converting the other benchmarks as well, or is there a particular reason only regex is being converted?

nikomatsakis:
I don't have a strong opinion about this. It seems to me we probably ought to measure what users see.

michaelwoerister:
The tokio tests are already based on CARGO_INCREMENTAL. The syntex case is a bit different, since the big crates come from crates.io and therefore would not be compile incrementally when using CARGO_INCREMENTAL.

Mark-Simulacrum:
Alright, sounds good to me. I'll merge when I get a chance to delete data for back collection purposes.

PR #157, opened 2017-10-05
Collect wall-time statistics

alexcrichton:
I thought we were already doing this, but apparently not!

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::process::Command;
+use std::time::{Instant, Duration};
 
 fn main() {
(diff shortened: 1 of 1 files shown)


collector/src/bin/rustc-fake.rs
```diff
@@ -71,6 +75,10 @@ fn print_memory() {
     }
 }
 
+fn print_time(dur: Duration) {
+    println!("{}.{:09};;wall-time;4;100.00", dur.as_secs(), dur.subsec_nanos());
```
Mark-Simulacrum: Could we add some documentation here? Especially around what the `4` represents. A link (since I assume this is the standard perf output) to documentation somewhere else would also be fine.

Issue #165, opened 2017-10-26
site/tests/lib.rs is outdated


kennytm:
It was still assuming the API where the `site` crate was called `rustc_perf` 😑 The whole test should be rewritten.

Mark-Simulacrum:
We've since given up on testing like this, primarily because I don't think it ever caught any actual bugs and was painful to maintain.

PR #166, opened 2017-10-27
Reenable Travis CI.

kennytm:
1. Merged the two original `.travis.yml` inside the subpackages to the top level.
2. Fixed several issues which prevented Travis from passing:

    1. Disabled `site/tests/lib.rs` due to #165. (Thus currently `cargo test` does absolutely nothing 🤷‍)
    2. Fixed the workspace issue when trying to run `collector/check-benchmarks.sh`.
    3. Repurpose `collector/check-benchmarks.sh` for Travis checking only (removed `set -x` and added some `travis_fold`). I think for development most people would run the collector directly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: rust
+sudo: false
+
+cache: cargo
+
+rust:
+  - nightly
+
+env:
+  - CHECK='cd collector && ./check-benchmarks.sh'
+  - CHECK='cargo build --verbose && cargo test --verbose'
+
+script:
+  - sh -x -c "$CHECK"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,6 @@
 [workspace]
-members = [
-    "collector",
-    "site",
-]
+members = ["site"] # do not add "collector", otherwise the exclude won't work.
+exclude = ["collector/benchmarks"]
 
 [profile.release]
 debug = true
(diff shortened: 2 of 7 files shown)


Mark-Simulacrum:
Will collector still be part of the workspace with the current changes?

kennytm:
Yes. (I think this may be a misfeature of Cargo.)

https://travis-ci.org/rust-lang-nursery/rustc-perf/jobs/293550642#L861

Mark-Simulacrum:
So we'd be depending on a bug, sort of?

kennytm:
@Mark-Simulacrum Yeah... the bug seems to be rust-lang/cargo#4089. There is a fix at rust-lang/cargo#4297 but it has been sitting there for 3 months? Perhaps we could hold this off until that cargo PR is available on nightly.

Mark-Simulacrum:
Hm, it looks like maybe it was fixed by https://github.com/rust-lang/cargo/pull/4594?

Mark-Simulacrum:
Hm, it looks like maybe it was fixed by https://github.com/rust-lang/cargo/pull/4594?

kennytm:
@Mark-Simulacrum Nope, the description states

> This is the first step of the fix, that's only a codemod to put together the relevant parts : workspace.members, workspace.excludes, and the root_dir (where members and excludes are relative to). **There's no logic change in this PR** to keep review easier.

PR #168, opened 2017-11-02
Add patches for the style benchmark 

SimonSapin:


collector/benchmarks/style-2f3bc0de49/makefile
```diff
@@ -19,20 +19,18 @@ all@020-clean:
 	    $(CARGO_RUSTC_OPTS) -Z incremental-info
 
 all@030-commit1:
-	patch -Np1 -i 030-commit-b5762295.patch
+	patch -Np1 -i 030-debugging-println.patch
 	touch components/style/lib.rs
 	CARGO_INCREMENTAL=1 \
 	    $(CARGO) rustc $(CARGO_OPTS) -- \
 	    $(CARGO_RUSTC_OPTS) -Z incremental-info
```
michaelwoerister: Would you mind removing `-Z incremental-info` here and everywhere else in this file? It might skew the timings since it causes the compiler to do more work than it has to.

SimonSapin: Done.

Mark-Simulacrum:
Could you rebase this and change the merge target into https://github.com/rust-lang-nursery/rustc-perf/tree/future? It should be relatively simple -- the primary change is removing the Makefile and renaming patches into `0-xxx.patch` `1-xxx.patch`, etc. instead of `000-xxx.patch`. You can have a look at the [config file](https://github.com/rust-lang-nursery/rustc-perf/blob/future/collector/benchmarks/style-servo/perf-config.json) for this crate for an example if you're adding an alternate crate, though you should be able to just reuse it.

SimonSapin:
I don’t think I can change the target branch of a PR. I’ll open new ones.

SimonSapin:
Never mind, I found how :)

Mark-Simulacrum:
As far as I can tell this was merged manually at some point but GitHub failed to recognize that, sorry for the wait!

PR #169, opened 2017-11-02
Add a benchmark for Servo’s script crate

SimonSapin:
This crate is not only big and takes a long time to compile, but also tends to sometimes expose pathological edge cases of the compiler.

This benchmark uses unstable language features, which is a concern. However I think this concern can be alleviated:

* I volunteer to fix this benchmark when a new compiler version breaks it. (We’ll need to figure out a way to notify me when that happens, ideally automatically.)
* To do so I will backport minimal changes, not update the whole thing to servo master. Changing the benchmark’s code makes comparing results less meaningful, but I’m hoping we can keep the changes small enough.
* This PR includes a number of changes that I’ve made to Servo lately to make breakage less likely to happen in the first place:
  * Usage of unstable feature that is not strictly necessary (for example optimizations that have a fallback path) is under an `unstable` Cargo feature that is enabled by default when building Servo, but is not enabled in this benchmark.
  * The `script_plugin` crate is an “old-style” compiler plugin that uses internal rustc APIs (such as traversing HIR) to implement a custom lint. In this benchmark, the lint is disabled and its code is not compiled. A bare minimum of the crate remains to whitelist a couple of attributes used by the lint. (It still uses `#[plugin_registrar]`, `rustc_plugin::Registry`, and `syntax::feature_gate::AttributeType::Whitelisted`.) In our experience maintaining this crate over time, that subset of APIs changes much less frequently that say details of HIR.

Thoughts?


Mark-Simulacrum:
I think Travis should be fixed now, though you may need to rebase. Once that's green (it should test benchmarks are successful) I'm happy to merge this.

SimonSapin:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

In another repo I work around this with the wrapper script below. What do you think?

```python
#!/usr/bin/python
import time
import subprocess
import sys

def main(args):
    process = subprocess.Popen(args)
    while 1:
        for _ in range(5 * 60):
            result = process.poll()
            if result is not None:
                sys.exit(result)
            time.sleep(1)
        print("(Dummy output to convince Travis we're still working)")

if __name__ == "__main__":
    main(sys.argv[1:])
```

Mark-Simulacrum:
Sounds good to me.

michaelwoerister:
> I volunteer to fix this benchmark when a new compiler version breaks it

Awesome, thank you such much!

Mark-Simulacrum:
Same as with https://github.com/rust-lang-nursery/rustc-perf/pull/168#issuecomment-342244707 -- could you change the target here and rebase?

SimonSapin:
Rebased as requested. However there is now no easy way to wrap the `cargo` call in a `travis_wait.py` wrapper script, so Travis will likely kill this job again.

PR #173, opened 2017-12-10
Add test case for rust-lang/rust#46449

kennytm:


Mark-Simulacrum:
Looks good. Will deploy once I get access to the machine (alexcrichton is at all hands so hasn't had a chance yet).

PR #185, opened 2018-03-01
Add benchmark for deeply nested types

ishitatsuyuki:
Request per https://github.com/rust-lang/rust/pull/48296#issuecomment-367858548.

Test case taken from https://github.com/rust-lang/rust/issues/38528#issuecomment-348748109.

### DIFF
--- a/collector/benchmarks/deeply-nested/Cargo.lock
+++ b/collector/benchmarks/deeply-nested/Cargo.lock
@@ -1,0 +1,4 @@
+[[package]]
+name = "deeply-nested"
+version = "0.1.0"
+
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Hm, I wonder if we could name this something like `deeply-nested` and then add a note to the Cargo.toml about which issue it specifically addresses -- that would help clarify what's going on at a cursory look at perf.rlo if there's regressions here I think.

Otherwise though this looks good.

Mark-Simulacrum:
Looks good to me. I'll deploy soon.

PR #199, opened 2018-03-28
Allow rescaling the y-axis when hiding a series.

kennytm:
Useful for checking non-NLL performance.

Requested in https://mozilla.logbot.info/rust-infra/20180323#c14509141-c14509144.

### DIFF
--- a/site/static/index.html
+++ b/site/static/index.html
@@ -155,8 +155,6 @@
                         enabled: false,
                     } : {},
                     min: 0,
-                    max: max * 1.05,
-                    softMax: max * 1.05,
                     ceiling: max * 1.05,
                     floor: 0,
                 } : {
--- a/site/static/index.html
+++ b/site/static/index.html
@@ -148,12 +148,9 @@
                     type: "datetime",
                 },
                 yAxis: absolute ? {
-                    title: crate_name.includes("-opt") ? { text: "" } : {
+                    title: (crate_name.includes("-opt") || crate_name.includes("-check")) ? { text: "" } : {
                         text: crate_name.startsWith("Summary") ? summaryYAxis : yAxis,
                     },
-                    labels: crate_name.includes("-opt") ? {
-                        enabled: false,
-                    } : {},
                     min: 0,
                     ceiling: max * 1.05,
                     floor: 0,


Mark-Simulacrum:
Does this change the default "look" when all graphs are shown where all three graphs horizontally have the same "max"?

kennytm:
@Mark-Simulacrum 🤔 Yes that would be changed. Perhaps better merge all three into the same canvas of 3 graphs?

Mark-Simulacrum:
Hm, so I'm actually not sure having the same Y axis is helpful since the data is drastically different in scale. Maybe @nikomatsakis / @michaelwoerister want to weigh in as to what would be preferable? As I see it, the options are:
 - Equivalent y-axis bounds on all three graphs, hiding line has no effect (today)
 - Equivalent y-axis bounds on all three graphs, hiding line on one hides on all and rescales to continue equivalency in y-axis.
 - Remove the y-axis bound setting, let the graph library auto-decide it. This means different bounds across all three graphs.

michaelwoerister:
Letting the graph auto-adjust is fine by me if the y-axis is properly labeled.

site/static/index.html
```diff
@@ -154,8 +154,6 @@
                         enabled: false,
```
Mark-Simulacrum: So we'll want to remove this logic here since we're no longer consistent across all three graphs as to the maximum.

nnethercote:
> Equivalent y-axis bounds on all three graphs, hiding line on one hides on all and rescales to continue equivalency in y-axis.
> Remove the y-axis bound setting, let the graph library auto-decide it. This means different bounds across all three graphs.

Which one of these is better depends on the data. E.g. for deep-vector there is not much difference between -check and -opt, so locking the three y-axes is good. But for piston-image there's a huge difference and you can barely read the -check graph. Ideally there would be a "lock y-axes" checkbox for each benchmark that lets you switch as necessary.

Having said that, either of the two quoted approaches would still be an improvement over the status quo.

nnethercote:
Thinking some more: being able to read each graph individually is the most important thing. And in some cases the locking of the y-axis across the three graphs makes the "-check" graph almost unreadable. In other words, while being able to lock the three graphs together can be nice, that's less important than having each graph scaled to use the full space available.

nnethercote:
All the graphs are readable now, yay!

PR #200, opened 2018-04-04
Remove the benchmarks/README.md file.

nnethercote:
It's almost entirely wrong -- make isn't used any more, collect.py isn't
used any more. The correct information would only be one or two
sentences, so I figured it was better to just remove the whole thing.


nnethercote:
@Mark-Simulacrum: the Travis failure looks like it's due to the existing code not matching what rustfmt would produce, mostly due to the ordering of `mod` and `use` items. Should it be ignored?

PR #207, opened 2018-04-06
Fix the 'bench_local' subcommand.

nnethercote:
Because it is currently pretty much unusable.

This patch changes its invocation to this:

  collector bench_local <ID> --cargo <CARGO> --rustc <RUSTC>

<ID> is an identifier used in the output file's name and contents, in the same
places that the commit SHA is used when benchmarking a particular commit.
<CARGO> and <RUSTC> are mandatory options specifying local versions of Cargo
and rustc.

Output is now produced in much the same way as the other subcommands,
with the exception that the directory named by --output-repo doesn't
have to be an actual rustc-timings repo; any directory will do. `site`
can be used to view the data in the output directory.

Details of note:

- The patch fixes up the README.md and the usage docs a bit.

- Sysroot is no longer used in bench_commit() or execute.rs. Instead we pass
  `rustc_path`, `cargo_path`, and (for bench_commit() only) `triple`, because
  that's all that is needed.

- Sysroot::command() is no longer used by CargoProcess::base_command().
  It was only contributing a tiny bit of functionality (e.g. the CARGO
  env var), which CargoProcess::base_command() can do itself.

- Repo::open() gains an additional parameter that tells it not to complain if
  the 'retries' file is missing; this is necessary now that ordinary
  directories can be used for output by 'bench_local'.


collector/README.md
```diff
@@ -1,39 +1,61 @@
...
 
-$RUSTC should point to the compiled rustc in stage1 or stage2, i.e.
-`./build/x86_64-unknown-linux-gnu/stage2/bin/rustc`. Stage 2 is likely a little better for
-benchmarking purposes, but not critical. The rustc binary has to be in a subdirectory of the
-`rustc-perf` root directory for the benchmark run to succeed.
+`$OUTPUT_DIR` should point to a directory, in which the timing data will be
+placed.
+
+`$RUSTC` should point to a rustc executable. For benchmarking purposes a stage
+2 build is a bit better than a stage 1 build, but not critical.
```
Mark-Simulacrum: We might want to mention that not all benchmarks will compile with a stage 1 build as they might need proc macros which won't work today at stage 1.

collector/README.md
```diff
@@ -1,39 +1,61 @@
...
+
+`$CARGO` should point to a Cargo executable.
```
Mark-Simulacrum: I think it's worth clarifying whether we expect these to be relative or absolute paths here.

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -88,8 +87,9 @@ pub struct Benchmark {
     config: BenchmarkConfig,
 }
 
-struct CargoProcess<'sysroot> {
-    sysroot: &'sysroot Sysroot,
+struct CargoProcess<'a> {
+    rustc_path: &'a PathBuf,
```
Mark-Simulacrum: hm, these should probably be `&'a Path` instead.

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -102,13 +102,16 @@ struct CargoProcess<'sysroot> {
 #[derive(Copy, Clone, Debug)]
 struct Incremental(bool);
 
-impl<'sysroot> CargoProcess<'sysroot> {
+impl<'a> CargoProcess<'a> {
     fn base_command(&self, cwd: &Path, subcommand: &str) -> Command {
-        let mut cmd = self.sysroot.command("cargo");
+        let mut cmd = Command::new(Path::new("cargo"));
         cmd
-            .env("SHELL", env::var_os("SHELL").unwrap_or_default())
```
Mark-Simulacrum: I've had problems with some benchmarks in the past if the SHELL environment variable isn't passed down, but if this works, I'm fine with landing it...

collector/src/bin/rustc-perf-collector/main.rs
```diff
@@ -240,7 +244,7 @@ fn main_result() -> Result<i32, Error> {
     )?;
     let use_remote = matches.is_present("sync_git");
     let out_repo = PathBuf::from(matches.value_of_os("output_repo").unwrap());
-    let mut out_repo = outrepo::Repo::open(out_repo, use_remote)?;
+    let mut out_repo = outrepo::Repo::open(out_repo, true, use_remote)?;
```
Mark-Simulacrum: Could we condition the `true` being passed here to be only if we're running locally? I'd like to avoid accidentally clearing out some git directory on the perf collector...

collector/src/bin/rustc-perf-collector/execute.rs
nnethercote: To be fair, I've only tested `helloworld` and `parser`! I will run all the benchmarks to see if SHELL or anything else is needed.

collector/src/bin/rustc-perf-collector/execute.rs
nnethercote: I've done that, and likewise for a bunch of the `rustc_path`/`cargo_path` function parameters. It was much easier than I expected... there must be some kind of automatic `PathBuf` to `Path` coercions happening.

collector/README.md
nnethercote: Ok.

collector/README.md
nnethercote: Ok.

collector/src/bin/rustc-perf-collector/main.rs
nnethercote: Sure.

PR #209, opened 2018-04-11
Implement profiling subcommands

nnethercote:
perf-record, Cachegrind, and DHAT are the profilers I've used in the past on rustc. This branch adds support for running each of them on the rustc-benchmarks.

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -4,38 +4,37 @@ use std::time::{Duration, Instant};
 
 fn main() {
     let mut args = env::args_os().skip(1).collect::<Vec<_>>();
-
     let rustc = env::var_os("RUSTC_REAL").unwrap();
-    let mut cmd = Command::new(&rustc);
 
-    // -is-final-crate is a special arg indicating that this is the final crate
-    // being compiled, i.e. we should measure it. We strip it out so the real
-    // rustc doesn't see it.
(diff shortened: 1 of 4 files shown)


collector/src/bin/rustc-fake.rs
```diff
@@ -13,24 +13,74 @@ fn main() {
...
+                assert!(cmd.status().expect("failed to spawn").success());
+            }
+
+            "cachegrind" => {
+                let mut cmd = Command::new("valgrind");
+                let has_valgrind = cmd.output().is_ok();
+                assert!(has_valgrind);
+                cmd.arg("--tool=cachegrind")
+                    .arg("--cache-sim=no")
+                    .arg("--branch-sim=no")
```
Mark-Simulacrum: I see "Note that you cannot specify --cache-sim=no and --branch-sim=no together, as that would leave Cachegrind with no information to collect." in the manpage for valgrind; is that information wrong/misleading?

nnethercote:
Well spotted on the Cachegrind docs! That info is indeed wrong, with --cache-sim=no and --branch-sim=no it still collects instruction counts. That's by far the most useful stat, and disabling the cache-sim and branch-sim make it run faster. I've added a comment about this, and I told the Valgrind maintainer about the doc error.

I also changed the `out_file` closure to use `format!` instead of multiple `push_str()` calls.

PR #212, opened 2018-04-18
Use separators when printing large numbers.

nnethercote:
Because 94,266,333,477.00 is much easier to read than 94266333477.00.

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -43,7 +43,7 @@
     function add_datum_fields(datum) {
         let html = "";
         if (datum) {
-            html += "<td>" + datum.toFixed(2) + "</td>";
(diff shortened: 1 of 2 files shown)


site/static/compare.html
```diff
@@ -43,7 +43,7 @@ <h3>Commits</h3>
     function add_datum_fields(datum) {
         let html = "";
         if (datum) {
-            html += "<td>" + datum.toFixed(2) + "</td>";
+            html += "<td>" + datum.toLocaleString('en-US', {minimumFractionDigits: 2, maximumFractionDigits: 2}) + "</td>";
```
Mark-Simulacrum: Is this supported across at least Chrome and Firefox?

nnethercote: Yes. I checked that it works on current release versions of Firefox, Chrome, and Safari.

PR #213, opened 2018-04-18
Add `profile_time_passes` subcommand.

nnethercote:
This subcommand profiles with -Ztime-passes.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -79,6 +79,8 @@ RUST_LOG=info ./target/release/collector --output-repo $OUTPUT_DIR \
 
 All this is the same as for the `bench_local` subcommand, except that
 `$PROFILE_CMD` is one of the following.
+- `profile_time_passes`: Profile with rustc's `-Ztime-passes`. Output is
+  written to files with a `Ztp` prefix.
 - `profile_perf_record`: Profile with `perf-record`. Output is written to
   files with a `perf` prefix. Those files can be read with `perf-report` and
   other similar `perf` commands.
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
I'm personally not opposed, but @nikomatsakis has told me in the past that time-passes shouldn't be used; instead we should use https://forge.rust-lang.org/profile-queries.html. Could you try that and see if it's more useful from your perspective?

nnethercote:
I just tried the profile-queries stuff. I got empty output files, perhaps because my rustc build doesn't have debug assertions enabled.

One strange thing about the profile-queries is that you apparently have to run rustc twice, once with `-Z profile-queries -Z incremental=cache`, and then again with `-Z time-passes`, which doesn't fit into the existing collector code all that well.

nnethercote:
I just tried again, using a rustc build in which debug assertions were enabled. Again, the output files are empty:
```
-rw-r--r-- 1 njn njn    0 May  7 16:41 profile_queries.ack
-rw-r--r-- 1 njn njn    0 May  7 16:41 profile_queries.counts.txt
-rw-r--r-- 1 njn njn 1333 May  7 16:41 profile_queries.html
-rw-r--r-- 1 njn njn    0 May  7 16:41 profile_queries.log.txt
```
(profile_queries.html has 1333 bytes of CSS and other stuff, but no actual data.)

@nikomatsakis: is the profile query stuff known to be broken?

nnethercote:
@Mark-Simulacrum: @nmatsakis has confirmed (in #50780) that profile-queries is broken. Until a better thing comes along, I think this subcommand can be useful, and the implementation is very simple. Sound ok to land?

Mark-Simulacrum:
Sounds good to me, thanks for confirming.

PR #214, opened 2018-04-19
Don't measure NLL for coercions or tuple-stress.

nnethercote:
Because NLL is *much* slower than non-NLL for those benchmarks.


Mark-Simulacrum:
Cc @nikomatsakis @michaelwoerister

PR #215, opened 2018-04-30
Add support for profiling with Callgrind.

nnethercote:
This is very similar to the support for Cachegrind.


nnethercote:
Cachegrind records instruction counts (and possibly cache/branch counts) for every line and every function. Callgrind does the same, but also records additional information about what functions call what other functions. So: Callgrind gathers more information but runs slower as a result. (There's an argument to be made that the two tools could be merged.)

PR #216, opened 2018-05-04
Add support for Massif

nnethercote:
Because it's useful for memory profiling.


Mark-Simulacrum:
Looks reasonable, thanks!

PR #217, opened 2018-05-04
Timeout git commands

Mark-Simulacrum:
@alexcrichton Could you perhaps take a look at the tokio/futures code here and tell me if I've made some mistake? I don't feel very confident about my use of select.


alexcrichton:
I think the process management is still blocking here, right? In that sense I'm not sure that this'll work right if it's actually blocking

Mark-Simulacrum:
My intent was that tokio would run the timer code on a separate thread, but maybe that's not the case. I'll try an alternative approach.

PR #222, opened 2018-05-11
Only use 1 iteration for `bench_local`.

nnethercote:
Because it's good enough and even 1 iteration takes plenty of time.


Mark-Simulacrum:
Seems reasonable. I think the variance might be higher -- especially for timing runs. Otherwise though this is good enough.

nnethercote:
I've been running locally with this setting for weeks. Apart from the "stylo-servo clean incremental" run, instruction counts variability is low enough. (I also disable NLL on all my local runs because it's so slow.)

PR #223, opened 2018-05-11
Improve docs on local benchmarking.

### LINKED ISSUE #221: Collector Documentation Issues
Some of the documentation on collector is a bit confusing to me, using it for the first time.

Two of the primary issues, one, because it uses the same `--output-repo` flag for committed and local builds, but the description is different. For the local build, should `--output-repo` still be a clone, or is it unrelated? Additionally, should the described `$CARGO` be a locally built cargo as well? I can't find a cargo in my build outputs.

I've been running it locally with apparently no output, several times, so the documentation (and maybe program feedback) isn't quite self-explanatory enough for me.

```
$ ./target/release/collector --output-repo ../rustc-timing/ profile_dhat rust0 --rustc ../rust0/build/x86_64-unknown-linux-gnu/stage2/bin/rustc --cargo /home/calebjon/.cargo/bin/cargo
```

nnethercote:
This fixes #221.

cc @porglezomp

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -21,19 +21,29 @@ output data will be placed and committed.
 
 To benchmark local builds:
 ```
-./target/release/collector --output-repo $OUTPUT_DIR bench_local \
-    --rustc $RUSTC --cargo $CARGO $ID
+RUST_LOG=info ./target/release/collector --output-repo $OUTPUT_DIR \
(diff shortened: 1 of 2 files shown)


porglezomp:
Thanks, these changes look very helpful.

PR #225, opened 2018-05-17
Remove `parser`.

### LINKED ISSUE #220: Remove parser
Most of the benchmarks in rustc-perf are either real code (e.g. clap-rs, crates.io, futures) or stress tests (e.g. deeply-nested, tuple-stress, unused-warnings). The two exceptions are helloworld and parser, which are toy programs.

I think it's reasonable for helloworld to be in the suite, to provide a lower bound on compilation time. But I think parser should be removed.
- It's a toy program, written (I think) by someone learning Rust.
- If you look at the code, it's actually exactly the same as helloworld, functionality-wise (there's extra code that isn't run).
- It's only about 700 lines of code.
- Profiles for it are quite similar to profiles for helloworld.

When I'm looking at profiles I almost always ignore it. I think its presence is a waste of human time and machine time.

@Mark-Simulacrum, what do you think?

nnethercote:
Fixes #220.

I can also use the `remove_benchmark` command to prepare a PR for rustc-timings that removes the `parser` data, if that's appropriate.


Mark-Simulacrum:
I believe we shouldn't need to run remove_benchmark, the site should handle the removal relatively well. If that's not the case we can always run it later.

Thanks!

PR #226, opened 2018-05-18
Add `profile_eprintln` subcommand.

nnethercote:
This subcommand captures stderr. This is useful for profiling a rustc
that has had ad hoc `eprintln!` statements added to it.

(This might seem like a weird one, but this kind of profiling is something I do frequently. Until now I was using the `profile_dhat` subcommand because it also captures stderr. But that is (a) super slow, because DHAT is super slow, and (b) includes all of DHAT's output mixed in with the `eprintln!` output. So I think it's worth having as a separate subcommand. I have already used it once today since I implemented it.)


Mark-Simulacrum:
Looks good to me, thanks! Pending travis, r=me

nnethercote:
Travis is happy. Ready to land?

PR #227, opened 2018-05-19
Switch to messagepack for POST response

ishitatsuyuki:
Switched the encoding of large responses to MsgPack so we can save data primarily on floats. Encoding is also faster as we don't need to make them decimal.

Additionally, changed the response in graph data from double precision floats to single precision, as we can't observe the difference with eye.

Results:

Response size: 3.8MB -> 1.8MB
Computing time: 1164ms -> 724ms

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,6 +703,14 @@ dependencies = [
 
 [[package]]
 name = "num-traits"
+version = "0.1.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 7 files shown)


Mark-Simulacrum:
Could you check if there's a significant difference in computing time on release mode?

Switching to f32 over f64 I think is reasonable; we should land that change regardless of the message pack switch. Could you extract it into a separate commit?

Otherwise this seems fine: I'm pleasantly surprised with how simple the change was to implement.

ishitatsuyuki:
Done.

PR #231, opened 2018-05-23
Unbreak compare.html.

nnethercote:
PR #227 added a dependence on msgpack.min.js to index.html, but failed
to add it to compare.html.


nnethercote:
Oh, #230 fixed this.

Issue #192, opened 2018-03-15
Create initial version of a "dashboard" that shows long-term performance trends


michaelwoerister:
This should implement https://github.com/rust-lang/rust/issues/48750, but for starters just containing the FROM-SCRATCH and SMALL-CHANGE usage scenarios.

I imagine this to be a separate page on perf.rlo, showing just two graphs, one for each usage scenario.

Each graph should contain a data point for each stable release (starting a few releases back, maybe with 1.17), the current beta, and the current nightly. The score should be calculated as the geometric mean of all build times for the given usage scenario, as listed in https://github.com/rust-lang/rust/issues/48750.

The main tasks to accomplish this are:
 - Add the missing benchmarks
   - [ ] ripgrep
   - [ ] webrender
   - [ ] cargo
   - [ ] winapi
   - [ ] stm32f103xx
   - [ ] encoding-rs (`cargo test --lib --no-run`)
   - [ ] clap-rs (`cargo test --no-run`)
   - [ ] regex (`cargo test --lib --no-run`)
   - [ ] syn (`cargo test --no-run`)
   - [ ] futures (`cargo test --test=all --no-run`)
 - Allow for running benchmarks in a Rust version specific way (so we don't crash when trying to specify `-Cincremental` on versions that don't support it yet)
 - Setup the new page (e.g. under perf.rlo/dashboard.html)

That page should make sure that all benchmark results for computing a given aggregate result are available, otherwise the result would be misleadingly low. 

The new benchmarks should also be displayed on the regular perf.rlo page. Some open questions:
 - Do we need to start looking into running tests in parallel on multiple machines (because they are becoming more and more)?
 - What to do with the `cargo test` benchmarks? Probably just treat them as separate benchmarks, that is, they would add another row instead of another column: `encoding-rs` and `encoding-rs (test)`, for example, would not be tied together in any way (although it would be nice if they could keep sharing the same source directory)

@Mark-Simulacrum, @nikomatsakis, thoughts?

nikomatsakis:
This sounds great!

Mark-Simulacrum:
> Do we need to start looking into running tests in parallel on multiple machines (because they are becoming more and more)?

[...]

PR #233, opened 2018-05-24
Add Cargo, Webrender, Ripgrep

Mark-Simulacrum:
Incrementally adding a smaller portion of the benchmarks listed in #192 to see
if we can afford to do this on current hardware.

cc @michaelwoerister


Mark-Simulacrum:
We should also find some way long-term to do this without ~200,000 line additions, but that's a different problem we can solve later.

nnethercote:
When I try to benchmark or profile `cargo` locally I get this error:
```
stderr=error: extra arguments to `rustc` can only be passed to one target, consider filtering
the package by passing e.g. `--lib` or `--bin NAME` to specify a single target
```
A normal `cargo build` in the `cargo` directory works, though.

nnethercote:
I updated my rustc and the `cargo` benchmark is now working.

nnethercote:
Looking at http://perf.rust-lang.org, ripgrep is showing up on both the graphs page and the compare page, but cargo and webrender are not.

PR #235, opened 2018-05-24
Mark measurements that have high variation.

nnethercote:
A couple of the rustc-perf benchmarks have high variation for particular
runs, which can lead to misleading results.

This patch adds a mechanism for marking such results. Here's an example of what the output looks like.
```
              style-servo          avg: 3.8%?         min: -0.4%?  max: 19.3%?
                    clean  327,201,649,786.00  327,079,069,169.00        -0.0%
     baseline incremental  580,869,925,861.00  581,642,436,771.00         0.1%
        clean incremental  107,184,307,330.00  127,902,127,040.00      19.3%??
     patched incremental:
        debugging println  336,986,389,937.00  335,518,589,160.00        -0.4%
     patched incremental:
 b9b3e592dd cherry picked  470,140,412,082.00  470,324,586,916.00         0.0%
```
Unreliable measurements are marked with "??". Values derived from unreliable
measurements are marked with "?". Furthermore, all marked measurements have a
"title" element explaining how they are unreliable, so you get a tool-tip if
you hover over them in a desktop browser.

The particular runs marked as unreliable are both of the "patched incremental"
runs for coercions, and the "clean incremental" run for style-servo. These are
based on my experience.

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -8,6 +8,11 @@
 <body class="container">
     <div>&gt; <a href="index.html">graphs</a>, <a href="compare.html">compare</a>.</div>
     <div id="content" style="display: none"></div>
+    <div style=''>
+        <p>Warning: although measurements known to have high variation are marked with
+           '?'/'??', this does not mean that unmarked measurements are guaranteed to have
+           low variation.</p>
+    </div>
     <div id="settings">
         <div id="commits" class="settings" style="text-align:left;">
             <h3>Commits</h3>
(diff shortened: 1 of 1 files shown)


nnethercote:
Here's a screenshot of a run I had lying around. The top four results are unreliable, as are five of the top 10! In practice the two versions of the compiler used in this run were very similar, but you can't tell that without these markers.
dodgy

Mark-Simulacrum:
I'm somewhat worried that by hard-coding these we'll create the impression that other benchmarks have low (or nonexistent) variance -- which I'm not certain is true. Is there a chance we could compute these variances for crates we conduct multiple runs of in the collector and save that data?

That's a larger change, but we could then indicate variances across the board which I think would be more useful. If you'd be willing to implement that, great, if not, could you add a note to the bottom (or somewhere in the UI) that benchmarks not marked with `?` might also be unreliable?

Mark-Simulacrum:
I'll probably deploy ~tomorrow.

nnethercote:
Thank you for the fast feedback and merge.

PR #237, opened 2018-05-25
Remove fetch polyfill

ishitatsuyuki:
Given that we already use arrow functions inside the JS code, it's clear that we don't support IE. Removing the fetch polyfill as well as it's implemented in all evergreen browsers.


Mark-Simulacrum:
Thanks!

PR #240, opened 2018-05-29
Change style-servo to avoid spurious rebuilds

Mark-Simulacrum:
The old code generation generated different code on rebuilds due to
depending on HashSet iteration order; this patch patches that problem by
replacing the HashSet with a BTreeSet. --pretty=expanded output is now
identical across builds.

cc https://github.com/rust-lang/rust/issues/48184, @SimonSapin @michaelwoerister 

I'm not actually sure this fixes the entire problem, but it's probably a good start.


michaelwoerister:
Thanks, @Mark-Simulacrum!

PR #242, opened 2018-06-08
Allow build/run kind selection in `bench_local`.

nnethercote:
This makes it more flexible, like `profile`.

This patch makes `Benchmark::measure()` and `Benchmark::measure()` more
similar to each other. This will make them easier to merge in the
future.

Also, there are now multiple mechanisms for specifying which runs to do;
they should be streamlined in the future.


Mark-Simulacrum:
This looks good to me, thanks!

PR #244, opened 2018-06-19
Use a linear y-axis for the dashboard graphs.

nnethercote:
A log y-axis is appropriate when the data covers multiple orders of
magnitude and/or is exponential in nature. In this case neither of those
two things are true. Furthermore, a log scale distorts the data,
exaggerating compile time reductions and minimizing compile time
regressions, which is undesirable.

So this patch changes to a log scale. It also ensures the y-axis starts
at zero, which is a good choice unless there is a clear reason not to do
so (and in this case there is not).

### DIFF
--- a/site/static/dashboard.html
+++ b/site/static/dashboard.html
@@ -31,7 +31,7 @@
             },
             yAxis: {
                 title: { text: "Seconds" },
-                type: "logarithmic",
+                min: 0,
             },
             xAxis: {
                 categories: versions,


nnethercote:
Here's a screenshot with the log scale:
log
And here's the same data with the linear scale:
linear
The main differences:
- No seasickness when looking at the y-axis with the linear scale.
- The "worst-case" regression when going from 1.23 to 1.24 isn't unduly compressed with the linear scale.

Mark-Simulacrum:
I'm happy to merge this (don't have much of an opinion here). cc @aturon

nnethercote:
A shorter summary of my position: a linear scale starting at 0 should be used for a graph unless there is a compelling reason not to, and in this case there is no compelling reason.

nnethercote:
One final point: http://perf.rust-lang.org/index.html uses a linear scale for all its graphs.

Mark-Simulacrum:
I believe @scottmcm was the one to originally argue for log-scale on Discord so it's likely they have thoughts here as well.

I am unopposed to merging or not merging: I don't really understand either side of the argument well enough, I think.

I'll probably merge this mid-next week if there's no further discussion here.

PR #247, opened 2018-06-22
Merge `Benchmark::measure()` and `Benchmark::profile()`.

nnethercote:
These two functions have a lot of overlap. This patch introduces the
`Processor` trait, which abstracts away the differences between the two
functions. There are two implementations of the new trait,
`MeasureProcessor` (which implements the `measure()`-specific
functionality) and `ProfileProcessor` (which implements the
`profile()`-specific functionality).

As a result `profile()` is no longer necessary, and `measure()` is
passed a `Processor` trait object.

Apologies for the large, hard-to-read patch; this change didn't lend itself
well to being broken up into smaller pieces.

### DIFF
--- a/collector/src/bin/rustc-perf-collector/execute.rs
+++ b/collector/src/bin/rustc-perf-collector/execute.rs
@@ -142,21 +142,17 @@ impl Profiler {
 struct CargoProcess<'a> {
     rustc_path: &'a Path,
     cargo_path: &'a Path,
+    cwd: &'a Path,
     build_kind: BuildKind,
-    profiler: Option<Profiler>,
     incremental: bool,
     nll: bool,
+    processor_etc: Option<(&'a mut dyn Processor, &'a str, RunKind, &'a str, Option<&'a Patch>)>,
     manifest_path: String,
     cargo_args: Vec<String>,
     rustc_args: Vec<String>,
 }
 
 impl<'a> CargoProcess<'a> {
-    fn profiler(mut self, profiler: Profiler) -> Self {
-        self.profiler = Some(profiler);
-        self
-    }
-
     fn incremental(mut self, incremental: bool) -> Self {
         self.incremental = incremental;
         self
@@ -164,6 +160,12 @@ impl<'a> CargoProcess<'a> {
 
     fn nll(mut self, nll: bool) -> Self {
         self.nll = nll;
+        self
+    }
+
+    fn processor(mut self, processor: &'a mut dyn Processor, name: &'a str,
+                 run_kind: RunKind, run_kind_str: &'a str, patch: Option<&'a Patch>) -> Self {
+        self.processor_etc = Some((processor, name, run_kind, run_kind_str, patch));
         self
     }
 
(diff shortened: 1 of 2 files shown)


collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -262,6 +258,265 @@ lazy_static! {
...
+}
+
+impl Processor for MeasureProcessor {
+    fn profiler(&self) -> Profiler {
+        Profiler::PerfStat
+    }
+
+    fn process_output(&mut self, _name: &str, _cwd: &Path, _build_kind: BuildKind,
+                      run_kind: RunKind, _run_kind_str: &str, patch: Option<&Patch>,
+                      output: process::Output) -> Result<bool, Error> {
```
Mark-Simulacrum: Can we refactor all of these arguments into some struct? Not sure what it would be called, but I think that would aid readability especially since we don't use many of them.

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -262,6 +258,265 @@ lazy_static! {
     };
 }
 
+pub trait Processor {
```
Mark-Simulacrum: Please add some fairly short documentation describing what each method here is intended to do.

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -262,6 +258,265 @@ lazy_static! {
...
+                    }
+                }
+                Ok(true)
+            }
+	    Err(DeserializeStatError::NoOutput(output)) => {
+		warn!(
+		    "failed to deserialize stats, retrying; output: {:?}",
+		    output
+		);
+                Ok(false)   // retry
```
Mark-Simulacrum: Could we change this bool to an enum? Not quite sure what to call it, but maybe for now just `Retry::{Yes, No}`. I guess to an extent it feels like we should be returning `Result<(), SomeError>` from this and retry would be indicated in the error enum/struct -- I think failure should allow this fairly easily.

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -746,7 +751,7 @@ fn process_perf_stat_output(output: process::Output) -> Result<Vec<Stat>, Deseri
     Ok(stats)
 }
 
-fn process_stats(build_kind: BuildKind, state: BenchmarkState, runs: Vec<Vec<Stat>>) -> Run {
+fn process_stats(build_kind: BuildKind, state: BenchmarkState, runs: &Vec<Vec<Stat>>) -> Run {
```
Mark-Simulacrum: `&[Vec<Stat>]`

collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -348,346 +605,94 @@ impl Benchmark {
         }
     }
 
-    /// Run a specific benchmark under perf-stat.
+    /// Run a specific benchmark under a processor + profiler combination.
     pub fn measure(
         &self,
+        processor: &mut dyn Processor,
```
Mark-Simulacrum: Let's refactor this to take `Processor` generically, that is this would be `&mut T where T: Processor`. (If I understand the callsites correctly this should be a trivial change, if it isn't then we don't need to do this).

collector/src/bin/rustc-perf-collector/main.rs
```diff
@@ -174,16 +174,18 @@ fn bench_commit(
...
-        let result = benchmark.measure(build_kinds, run_kinds, rustc_path, cargo_path, iterations);
-
-        if let Err(ref s) = result {
-            info!("failed to benchmark {}, recorded: {}", benchmark.name, s);
-        }
+        let mut processor = execute::MeasureProcessor::new(&benchmark.name);
+        let result = benchmark.measure(&mut processor, build_kinds, run_kinds, rustc_path,
+                                       cargo_path, iterations);
+        let result = match result {
+            Ok(()) => Ok(processor.collected),
```
Mark-Simulacrum: This feels quite odd: normally, I'd expect collected to be in the result itself, but I think changing that's a larger refactor than is necessary, and is rather hard. Let's leave it for now.

collector/src/bin/rustc-perf-collector/execute.rs
nnethercote: The change works, but since I store a `dyn Processor` reference elsewhere, I used `dyn Processor` here for consistency. Seem reasonable?

Mark-Simulacrum: Makes sense, I had missed that before.

PR #251, opened 2018-06-27
Improve dashboard's presentation of data series

nnethercote:
This PR renames the existing data series, adds the "clean" series, and improves the transition between the pre- and post-incremental compilation eras.

Here is a screenshot showing the current dashboard:
dashboard0

Here is a screenshot showing the dashboard with these patches applied:
dashboard

I admit that the code in the second patch that identifies pre-incremental versions is hacky; I'm happy to hear alternative ideas.

### DIFF
--- a/site/src/server.rs
+++ b/site/src/server.rs
@@ -102,7 +102,7 @@ pub fn handle_nll_dashboard(
         .map(|bench| {
             let nll = bench.runs.iter().find(|r| r.check && r.is_nll())
                 .and_then(|r| r.get_stat(&body.stat));
-            let clean = bench.runs.iter().find(|r| r.check && r.is_non_incremental_clean())
+            let clean = bench.runs.iter().find(|r| r.check && r.is_clean())
                 .and_then(|r| r.get_stat(&body.stat));
 
             nll_dashboard::Point {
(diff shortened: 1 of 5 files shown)


site/src/server.rs
```diff
@@ -203,13 +203,18 @@ pub fn handle_dashboard(data: &InputData) -> dashboard::Response {
...
-            extend!(check_clean_incr_points, r, r.is_clean_incr() && r.check);
-            extend!(check_println_incr_points, r, r.is_println_incr() && r.check);
             extend!(debug_clean_points, r, r.is_clean() && !r.check && !r.release);
-            extend!(debug_base_incr_points, r, r.is_base_incr() && !r.check && !r.release);
-            extend!(debug_clean_incr_points, r, r.is_clean_incr() && !r.check && !r.release);
-            extend!(debug_println_incr_points, r, r.is_println_incr() && !r.check && !r.release);
+            // Skip these incremental measurements in versions prior to 1.24.0,
+            // because they didn't support incremental compilation.
+            if version != "1.19.0" && version != "1.20.0" && version != "1.21.0"  &&
+               version != "1.22.0" && version != "1.23.0" {
```
Mark-Simulacrum: I think we have code in the collector that does this -- we can parse the version as `semver::Version` and then compare as less than `1.24.0`. I think that should look cleaner.

PR #253, opened 2018-06-29
Data extrapolation/intrapolation

Mark-Simulacrum:
Fixes https://github.com/rust-lang-nursery/rustc-perf/issues/249.

Extrapolates on a run-based level across history.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -243,7 +243,7 @@ pub struct CommitData {
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct DeltaTime(#[serde(with = "round_float")] pub f64);
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Hash, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Date(pub DateTime<Utc>);
 
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -14,6 +14,7 @@ use std::hash;
 use std::str::FromStr;
 use std::path::{Path, PathBuf};
 use std::process::{self, Stdio};
+use std::borrow::Cow;
 
 use chrono::{DateTime, Datelike, Duration, TimeZone, Utc};
 use chrono::naive::NaiveDate;
@@ -61,11 +62,25 @@ impl Ord for Commit {
     }
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Patch {
     index: usize,
     pub name: String,
     path: PathBuf,
+}
+
+impl PartialEq for Patch {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+    }
+}
+
+impl Eq for Patch {}
+
+impl hash::Hash for Patch {
+    fn hash<H: hash::Hasher>(&self, h: &mut H) {
+        self.name.hash(h);
+    }
 }
 
 impl Patch {
@@ -138,14 +153,14 @@ impl BenchmarkState {
         }
     }
 
-    pub fn name(&self) -> String {
+    pub fn name(&self) -> Cow<'static, str> {
         match *self {
-            BenchmarkState::Clean => format!("clean"),
-            BenchmarkState::Nll => format!("nll"),
-            BenchmarkState::IncrementalStart => format!("baseline incremental"),
-            BenchmarkState::IncrementalClean => format!("clean incremental"),
+            BenchmarkState::Clean => "clean".into(),
+            BenchmarkState::Nll => "nll".into(),
+            BenchmarkState::IncrementalStart => "baseline incremental".into(),
+            BenchmarkState::IncrementalClean => "clean incremental".into(),
             BenchmarkState::IncrementalPatched(ref patch) => {
-                format!("patched incremental: {}", patch.name)
+                format!("patched incremental: {}", patch.name).into()
             }
         }
     }
@@ -176,13 +191,55 @@ pub struct Stat {
     pub cnt: f64,
 }
 
-#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Run {
     pub stats: Vec<Stat>,
     #[serde(default)]
     pub check: bool,
     pub release: bool,
     pub state: BenchmarkState,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct RunId {
+    check: bool,
+    release: bool,
+    state: BenchmarkState,
+}
+
+impl RunId {
+    pub fn name(&self) -> String {
+        self.to_string()
+    }
+}
+
+impl fmt::Display for RunId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let opt = if self.release {
+            "-opt"
+        } else if self.check {
+            "-check"
+        } else {
+            ""
+        };
+        write!(f, "{}{}", self.state.name(), opt)
+    }
+}
+
+impl PartialEq for Run {
+    fn eq(&self, other: &Self) -> bool {
+        self.release == other.release &&
+        self.check == other.check &&
+        self.state == other.state
+    }
+}
+
+impl PartialEq<RunId> for Run {
+    fn eq(&self, other: &RunId) -> bool {
+        self.release == other.release &&
+        self.check == other.check &&
+        self.state == other.state
+    }
 }
 
 impl Run {
@@ -209,15 +266,18 @@ impl Run {
         false
     }
 
+    pub fn id(&self) -> RunId {
+        let state = self.state.clone();
+        let state = state.erase_path();
+        RunId {
+            check: self.check,
+            release: self.release,
+            state: state,
+        }
+    }
+
     pub fn name(&self) -> String {
-        let opt = if self.release {
-            "-opt"
-        } else if self.check {
-            "-check"
-        } else {
-            ""
-        };
-        self.state.name() + opt
+        self.id().name()
     }
 
     pub fn get_stat(&self, stat: &str) -> Option<f64> {
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -91,7 +91,7 @@ pub mod data {
     use server::DateData;
     use collector::Bound;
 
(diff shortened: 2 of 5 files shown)


nnethercote:
Apologies: I'm having trouble imagining the problems you've described above. I find
this stuff a lot easier to think about with concrete examples.

Consider the following.
- A window of 20 commits, labeled 1..20.
- Four benchmarks: `cargo`, `futures`, `parser`, `serde`
  - `cargo` is added partway through, `parser` is removed partway through, `serde`
    breaks intermittently (completely or just for NLL)
- Two BuildKind+RunKind combinations: Check/Clean, Check/NLL
  - I don't think it's worth worrying about addition/removal of
    BuildKind+RunKind combinations. One day NLL will not be needed but then we
    can just remove it wholesale.
- Just the "instructions" metric, measured in billions.
  - I don't think it's possible to gather a subset of the metrics -- we either  
    get all of them or none of them -- so we don't have to worry about some
    being present and some not.

```
-----------------------------------------------------------------------------
commit  cargo           futures         parser          serde   
        Clean, NLL      Clean, NLL      Clean, NLL      Clean, NLL
-----------------------------------------------------------------------------
1       X,    X         5.0, 6.0        2.0, 3.0        8.0, 10.0
2       X,    X         5.0, 6.0        2.0, 3.0        8.0, 10.0
3       X,    X         5.0, 6.0        2.0, 3.0        I,   I
4       X,    X         5.0, 6.0        2.0, 3.0        I,   I
5       X,    X         5.0, 6.0        2.0, 3.0        8.0, 10.0
6       X,    X         5.0, 6.0        2.0, 3.0        8.0, 10.0
7       X,    X         5.0, 6.0        X,   X          8.0, 10.0
8       X,    X         5.0, 6.0        X,   X          8.0, 10.0
9       X,    X         5.0, 6.0        X,   X          8.0, 10.0
10      X,    X         5.0, 6.0        X,   X          I,   I
11      X,    X         5.0, 6.0        X,   X          I,   I
12      X,    X         5.0, 6.0        X,   X          I,   I
13      X,    X         5.0, 6.0        X,   X          8.0, 10.0
14      X,    X         5.0, 6.0        X,   X          8.0, 10.0
15      30.0, 40.0      5.0, 6.0        X,   X          8.0, 10.0
16      30.0, 40.0      5.0, 6.0        X,   X          8.0, 10.0
17      30.0, 40.0      5.0, 6.0        X,   X          8.0, I   
18      30.0, 40.0      5.0, 6.0        X,   X          8.0, I   
19      30.0, 40.0      5.0, 6.0        X,   X          8.0, 10.0
20      30.0, 40.0      5.0, 6.0        X,   X          8.0, 10.0
-----------------------------------------------------------------------------
```

For simplicity, I haven't shown any variation in the measurements for each
column.

So, overall we will have a single graph with two data series (Clean & Nll). Each 
data point will be a summary (i.e. average) of 4 values: 1..4 measured values,
and 0..3 interpolated/extrapolated values.

- Extrapolations (X):
  - cargo: copy the 30.0,40.0 from commit 15 back through commits 1..14
  - parser: copy the 2.0,3.0 from commit 6 forward through commits 7..20

- Interpolations (I):
  - serde: interpolate both Clean and NLL values for commits 3..4 and 10..12
  - serde: interpolate just the NLL value for commits 17..18

I don't see any particular difficulties but I might be overlooking something.

@Mark-Simulacrum: can you map your concerns onto this example data?
Feel free to modify the example as necessary to aid your explanation. Thanks!

nnethercote:
Also, I'd be interested to see screenshots of the new summary graphs. How do they compare to the existing ones? How do they compare to the per-benchmark graphs?

Mark-Simulacrum:
Yes, that all makes sense. I'll look at implementing the within-benchmark extrapolation and colored lines soon-ish.

Another question that came up while I was looking at this is if we'd like a toggle on the site which would enable/disable this -- that should be fairly easy to implement with the current setup (though it'd require a page reload). Not sure if it would be useful, though.

nnethercote:
I don't think a toggle would be useful, except possibly for short-term testing purposes.

Mark-Simulacrum:
Okay, looks like we have an initial version which I believe should work pretty much as expected; these are some selected screenshots; I am not ready to merge this yet as the startup performance in abysmal (that's when we do the computation for this) -- nearly 2 minutes.

image
image
image
image

nnethercote:
Cool! Some observations and questions.
- The summary graphs look promising. Nice flat lines :)
- I'm surprised to see some non-pink dots for the NLL data series in the "Summary-debug" graph -- that suggests that all the benchmarks compiled successfully, but NLL is explicitly disabled for some of them.
- What's happening with the upwards-curving pink on the left side of the `coercions` graphs? If data is missing for that entire period, I would have expected a flat line extending to the left of the first data point.
- How does `html5ever` have an (all-pink) NLL line, when it's not enabled for NLL?
- Will this extend easily to the dashboard view?

Mark-Simulacrum:
> I'm surprised to see some non-pink dots for the NLL data series in the "Summary-debug" graph -- that suggests that all the benchmarks compiled successfully, but NLL is explicitly disabled for some of them.

This PR didn't change that behavior (seems somewhat out of scope) -- we currently generate summary graphs/lines independent of whether all benchmarks actually ran for any commit. I also am uncertain that we have a clear expectation of what we *should* be doing here; we should discuss and implement the results.

> What's happening with the upwards-curving pink on the left side of the coercions graphs? If data is missing for that entire period, I would have expected a flat line extending to the left of the first data point.

We don't limit interpolation/extrapolation to the visible window, so that is actually a smaller section of the larger curve. The reason that this data looks somewhat non-linear is subtle: our interpolation works on a commit basis, not on a time bases, but the X-axis on these graphs is time-based. Therefore, if the commits were more spaced out then it results in a less linear looking slope, though the data is linear on a commit-based x-axis.

image

> How does html5ever have an (all-pink) NLL line, when it's not enabled for NLL?

I believe it used to be enabled for a short period of time, and then was disabled. The extrapolation/interpolation code does not currently limit itself to some time/commit range so it "reaches" into the past far enough to find those points. I see at least two points around 2018-03-22. The behavior here might not be ideal.

> Will this extend easily to the dashboard view?

It should. I haven't updated its code yet in this PR for the visual effects, though. I have also not updated the comparison, NLL, or status pages -- I'd like suggestions if you have any on the correct approach for the dashboard (should we interpolate/extrapolate still? stop showing lines as in summary graphs?), as well as the proper way to indicate data interpolation on the comparison page.

Specifically -- should we "color" the whole section somehow so it stands out in the overall view? Maybe add a Unicode warning sign? Something less fancy?

On a slightly less important note, do you think the currently spontaneously chosen color is fine? Should we go for something less visually jarring?

nnethercote:
> We don't limit interpolation/extrapolation to the visible window,

Should we? That's what I was expecting.

> to indicate data interpolation on the comparison page

I don't think we need to to interpolation on the comparison page. It shows data for two commits and doesn't do any summarizing. When data is missing it's already obvious.

> Something less fancy?

I'm not really sure. "Jarring" is definitely a good word for the current pink :)  It probably is worth distinguishing the following two cases, though:
- This data point was missing and we have filled it in (for individual benchmark graphs).
- This data point is a mix of real data and interpolated data (for summary graphs and the dashboard).

Mark-Simulacrum:
> > We don't limit interpolation/extrapolation to the visible window,
>
> Should we? That's what I was expecting.

This is somewhat harder -- currently we just run the extrapolation once when we load data and then cache that. Doing so on each request is certainly possible but will probably make the UI slower. I'd rather leave it as-is today, especially since that also means that changing the view you're looking at does not affect the data: this seems like a good property to have, overall.

Mark-Simulacrum:
I've rebased and I believe I'm ready to merge this -- the main unimplemented features are:
 - dashboard gives no indication that data may be interpolated
  - doesn't seem all that important, can only affect current master anyway
 - Interpolate within visible range
   - I don't think this is something I'd want to do (yet, at least)

@nnethercote If you think the above are non-blockers then we can merge and deploy this whenever.

nnethercote:
Apologies for the slow response.

Overall, it's great to see this land:
- It makes it obvious when individual benchmarks were broken for extended periods of time. E.g. `serde` and `tokio-webpush-simple` both were broken in June.
- The summary graphs are *much* more meaningful now.

There are still some rough edges, though.

NLL had a big jump on the `summary-debug` graph around July 2. I can't really see why, though it might be related to some missing extrapolations (see below).

Extrapolation is missing on the RHS of graphs, e.g. the NLL results for `cargo-debug` (and several others). It's also missing from the LHS in places where it's missing from the RHS, e.g. the NLL results for `crates.io-debug`.

Why can interpolation only affect master on the dashboard? I agree that it's much more likely to affect master (because master is most likely to have bustage), but that doesn't seem guaranteed. I remember when @aturon was first doing these measurements he had a failure for one of the benchmarks on one old release.

Anyway, I think it would be useful to have the pink indications on the dashboard.

On the graphs view, there is no explanation or legend explaining what the pink means. It's probably mysterious to everyone but you and me!

The new shade of pink is better, BTW :+1:  It would still be good two have two different colours, though, one meaning "partial data" (for the summary graphs and dashboard) and one meaning "no data" (for the individual benchmarks).

Mark-Simulacrum:
> Extrapolation is missing on the RHS of graphs, e.g. the NLL results for cargo-debug (and several others). It's also missing from the LHS in places where it's missing from the RHS, e.g. the NLL results for crates.io-debug.

This might be due to the fact that we do not extrapolate runs/benchmarks that are not in the last 20 commits; that behavior is what allows benchmarks to eventually "go away" if they're removed, as with NLL for *-debug and *-opt.

> Why can interpolation only affect master on the dashboard?

Well, to clarify, interpolation can only affect master *today* because it's not implemented to be generic over the dataset (i.e., only works with commits) but I agree that in theory we could extrapolate into artifact that too. For now though I think this'll have to be out of scope, ideally we'd have data for all benchmarks across all artifacts -- it might be worth adding some basic checkboxes below the dashboard graphs that confirm that.

> On the graphs view, there is no explanation or legend explaining what the pink means. It's probably mysterious to everyone but you and me!

This is generally speaking the case -- do you have any ideas about where to put documentation like this?

> It would still be good two have two different colours, though, one meaning "partial data" (for the summary graphs and dashboard) and one meaning "no data" (for the individual benchmarks).

It seems like that distinction is apparent by the location; i.e. on the summary graphs it's ~always partial data and on the individual benchmarks it's always no data, so I'm not quite sure of the usefulness here. Could you perhaps elaborate?

nnethercote:
The ideal spot for the explanation of the pink dots would be in the graphs themselves, as part of the legend for the other colored dots. I don't know if that's possible. If not, just a note at the bottom of the page is better than nothing.

> It seems like that distinction is apparent by the location; i.e. on the summary graphs it's ~always partial data and on the individual benchmarks it's always no data

That is true, but it's a non-obvious distinction. If we end up with per-graph indications of what the pink dots mean, then a single color is less problematic. But if the pink dots are explained at the bottom of the page, distinguishing the two cases would be a lot clearer.

PR #255, opened 2018-07-04
Enable two-phase-borrows in NLL

Mark-Simulacrum:
cc @nnethercote @nikomatsakis -- I think we'll see a bump in compile times


nnethercote:
By "bump" you mean an increase? :(

Mark-Simulacrum:
Yes, I assume so -- but could be wrong. I don't know how two phase borrows are implement under the hood but presume that there's additional code run over just MIR borrowck.

Mark-Simulacrum:
I've deployed this so we should see the results within ~5 hours or so.

PR #257, opened 2018-07-09
Enable NLL for html5ever.

nnethercote:
NLL causes a 20x slowdown, but that's similar to the `tuple-stress`
slowdown, and much better than the `tuple-stress` slowdown used to be.


Mark-Simulacrum:
Has the memory usage problem been fixed sufficiently for this to be feasible?

nnethercote:
Ah, I overlooked that. Probably not yet... https://github.com/rust-lang/rust/issues/52028 has some ideas, but no PRs filed for those ideas.

PR #265, opened 2018-07-30
Add keccak crate

RalfJung:
Quoting @oli-obk:
> The keccak crate is a great benchmark for ridiculous numbers of locals and basic blocks. We should add it to the list of perf tests.

So, let's do that. :) This adds an old version, one that still actually has "ridiculous numbers of locals and basic blocks". Later versions reduced that, also to help with compile-times -- but the old version is a good stress-test, so I think it is the better benchmark.
Concretely, this is an exact copy of https://github.com/RustCrypto/sponges/tree/a417e97a41776a2feee20f98e1d97917b7ad35f0/keccak.

Is there anything else needed, other than adding it in this directory?

### DIFF
--- a/collector/benchmarks/keccak/Cargo.lock
+++ b/collector/benchmarks/keccak/Cargo.lock
@@ -1,0 +1,4 @@
+[[package]]
+name = "keccak"
+version = "0.1.0"
+
--- a/collector/benchmarks/keccak/Cargo.toml
+++ b/collector/benchmarks/keccak/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "keccak"
+version = "0.1.0"
+authors = ["RustCrypto Developers"]
+license = "CC0-1.0"
+description = "Keccak-f sponge function"
+documentation = "https://docs.rs/keccak"
+repository = "https://github.com/RustCrypto/sponges"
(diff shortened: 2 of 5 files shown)


RalfJung:
I wanted to also add the benchmark @oli-obk and me developed for https://github.com/rust-lang/rust/issues/52849#issuecomment-408926242, but given that it is currently really slow that might not be a good idea yet?

Mark-Simulacrum:
Pending CI passing this looks good.

I think I'd be fine to add the other benchmark if it's actually 30 seconds -- many of our benchmarks are ~5-10 minutes long so a 30 second or minute long build won't do all that much :)

RalfJung:
It's 30 seconds after reducing complexity.^^ The original one (which takes <1min on a nightly from a week ago) did not complete compilation in 7min, then I stopped it...

Mark-Simulacrum:
Ah, okay, then yeah, let's hold off on that. Looks like CI here has passed so merging and deploying shortly.

RalfJung:
FWIW, keccak regressed 2% since 2018-07-07. Doesn't seem big enough to track separately, I guess. Also maybe fixing https://github.com/rust-lang/rust/issues/52849 will fix that as well.

PR #269, opened 2018-08-10
add packed_simd benchmark

### LINKED ISSUE #268: add packed_simd as a benchmark
Building the tests of https://github.com/rust-lang-nursery/packed_simd (`cargo build --tests`) takes a while to build. It uses a lot of memory, it has dozens (probably hundreds) of modules, hundreds of trait implementations, lots of macros, etc. 

I think it would be a very interesting benchmark for rustc-perf

gnzlbg:
This adds packed_simd to the benchmarks/ directory. Is there anything else I have to update here?

Closes #268


Mark-Simulacrum:
@gnzlbg Sorry for taking so long on this -- it looks like there's a CI failure due to missing Cargo.lock file; could you fix that?

Modulo CI passing though I think we should be able to merge this

gnzlbg:
@Mark-Simulacrum i've added the missing lock files but as mentioned i worry a bit that not using a git submodule will make these hard to manage down the road, and as mentioned I have no idea how rustc-perf is supposed to know how to run packed simd benchmarks (i've only added it to the collectors directory).

Mark-Simulacrum:
perf.rlo does not run any sort of runtime benchmarks today. Submodules currently don't quite work since we often need a perf-config file, though I've wanted to rework that eventually, somehow - perhaps pointing at a fork of the project.

gnzlbg:
> perhaps pointing at a fork of the project.

It also can be a branch in the project repo but a fork would be ok - submodules let us do whatever we want here. 

> and as mentioned I have no idea how rustc-perf is supposed to know how to run packed simd benchmarks

damn this PR was so long ago I thought we were talking about `lolbench` - `rustc-perf` should only bench how long do things take to compile, duh.

gnzlbg:
@Mark-Simulacrum so I've updated `packed_simd` and now it builds correctly. I've opened: https://github.com/rust-lang-nursery/rustc-perf/issues/295 to track that currently CI still passes even if some benchmarks fail to build =/

Mark-Simulacrum:
Hm, that's odd -- perhaps this needs to be rebased? I thought I had fixed that

gnzlbg:
I just rebased, but since I already fixed the issue, nothing will show up :D 

The ci script contains: `set -x;` , I often use `set -ex`, maybe that is the reason?

Mark-Simulacrum:
I'll merge and deploy this sometime today; thank you!

PR #272, opened 2018-08-17
Tweak the `perf-record` invocation.

nnethercote:
`--freq=99` gives much smaller samples than the default (e.g. 37x
smaller for the case I just tried; 5.3MB rather than 197MB), and it's
what lots of `perf` tutorials suggest.

The patch also removes a stray comment that was left behind in a
previous commit.


Mark-Simulacrum:
Hm, so to clarify, doesn't this mean we're reducing accuracy? I've never quite understood the perf frequency option...

Mark-Simulacrum:
Hm, well, seems largely fine to do. Probably doesn't actually end up mattering too much. Thanks!

PR #278, opened 2018-08-24
Add ucd crate

lqd:
As seen in https://github.com/rust-lang/rust/issues/53643, this crate has big statics similar to `html5ever`, but different enough that it's still an NLL stress test.

@nikomatsakis mentioned we should add it to perf [in this comment](https://github.com/rust-lang/rust/issues/53643#issuecomment-415720157).


Mark-Simulacrum:
Can you add a perf-config.json file with these contents? That way it'll take up less time on the benchmark machine.

```json
{
    "runs": 1
}
```

PR #279, opened 2018-08-26
Trim GitHub commit link from commit hashes

varkor:
It's convenient simply to copy a GitHub link for a commit instead of extracting the commit hash, so let's just ignore that when parsing a message.


Mark-Simulacrum:
Thanks! Merging and deploying shortly.

PR #281, opened 2018-08-31
Update markers on high variance benchmarks.

nnethercote:
These can be verified by eyeballing the graphs on the front page of
perf.rust-lang.org.


Mark-Simulacrum:
Thanks!

PR #284, opened 2018-09-03
Extend the docs some more.

nnethercote:
Specifically:

- Everything is now in one of two sections: Benchmarking or Profiling.

- A bit more introductory text in each section about why you would want
  to be doing this.

- Build docs changed to build `site` as well, which is necessary to view
  local benchmarking results.

- Added more detail about how @bors/@rust-timer benchmark runs are done.

- Added info about the kinds of measurements taken when benchmarking.

- Added info about locally comparing two versions of the compiler.

- Information about server-specific benchmarking commands has been moved
  after the stuff about local commands, because it's of interest to
  fewer people.

- A few other minor details added.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -1,25 +1,48 @@
-# Rust Compiler Performance Benchmarking
+# Rust Compiler Performance Benchmarking and Profiling
 
 ## How to build
 
+Before doing anything else, you should build `collector` (for running the
+benchmarks) and `site` (for viewing the results):
 ```
-cargo build -p collector --release
-```
-
-## Benchmarking committed builds
-
-To benchmark builds from a rustc repository:
-```
-./target/release/collector --output-repo $RUSTC_TIMING process
+cargo +nightly build --release
 ```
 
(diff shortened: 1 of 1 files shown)


collector/README.md
```diff
@@ -1,25 +1,48 @@
...
+privileges. Ping `simulacrum` on IRC as a starting point.
+
+That person will enter `@bors try` as a comment in the PR. This queues a normal
+try build on Travis, which takes some time.
+
+Once the try build has completed, that person will enter `@rust-timer build
+$MERGE` as a comment in the PR, where `$MERGE` is the full 40 character merge
+revision ID from the try build. This queues a benchmarking run, which also
+takes some time. Upon completion, a URL is posted in the PR at which the
+results can be viewed.
```
Mark-Simulacrum: Might be good to clarify here that the URL is posted instantly; results are generally available in ~6 hours or so I think

nnethercote:
@Mark-Simulacrum: nit has been fixed.

PR #288, opened 2018-10-05
Change a `cp` command.

nnethercote:
The existing command used is `cp -r -T -- src_dir dst_dir`. But
-T doesn't work on Mac.

This commit changes it to `cp -r src_dir/. dst_dir`.

- Appending the `.` has the same effect as -T: copy the contents of
  `src_dir` without copy `src_dir` itself.

- Also, -R is POSIX but -r is not.

### DIFF
--- a/collector/src/bin/rustc-perf-collector/execute.rs
+++ b/collector/src/bin/rustc-perf-collector/execute.rs
@@ -577,12 +577,14 @@ impl Benchmark {
     }
 
     fn make_temp_dir(&self, base: &Path) -> Result<TempDir, Error> {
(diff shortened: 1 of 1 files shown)


nnethercote:
This came up in https://github.com/rust-lang/rust/pull/53793#issuecomment-426875071, where @toidiu is trying to profile on Mac. For me, it helped, but Cachegrind doesn't run rustc properly and so I end up getting another error further along. Still, seems worth landing.

Mark-Simulacrum:
Thanks! Can't say I'm too familiar with the details myself but this seems fine; we can reimplement the copy in rust if needed.

PR #293, opened 2018-10-10
Clarify the required positions of the `bench_local` options.

### LINKED ISSUE #290: README for running collector unclear on where flags must go
The README at https://github.com/rust-lang-nursery/rustc-perf/tree/master/collector#benchmarking-options documents a bunch of flags, but it forgets to mention that some of them must be put *before* the subcommand (`bench_local` in my case), while others must be put *after* it. So it has to be `--filter syn  bench_local --builds Opt` and not any of the other possible mutations.

nnethercote:
Fixes #290.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -74,25 +74,24 @@ depending on the speed of your machine.
 
 ### Benchmarking options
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Thanks!

PR #294, opened 2018-10-25
Delete package.json and package-lock.json from crates.io

kennytm:
Stops GitHub from sending the spurious vulnerable alert all the time.


Mark-Simulacrum:
Interestingly I don't think I received that alert but seems fine either way.

PR #325, opened 2019-01-14
Fix some "error: missing documentation for macro" errors.

nnethercote:
The compiler must have got stricter about these recently.

This gets `futures`, `hyper`, and `style-servo` compiling again.
`script-servo` is still broken due to a different, trickier problem.


varkor:
I think this is due to https://github.com/rust-lang/rust/pull/56407.

PR #343, opened 2019-02-17
Add wg-grammar benchmark

matthewjasper:
cc rust-lang/rust#58178


Mark-Simulacrum:
If you could rebase that should fix CI -- well, make it run at least -- and then we can merge, I think.

nnethercote:
How do the `instructions` and `max-rss` numbers compare against the other benchmarks, for both non-NLL and NLL builds? It looks like you've just copied the full `wg-grammar` repo verbatim, and NLL builds were taking over an hour on my machine for that. We don't want this benchmark to be drastically larger than the others. ([Here](https://github.com/rust-lang/rust/issues/58178#issuecomment-460898187) are instructions on how to reduce input size.)

matthewjasper:
The compile time is much better on the latest nightly. Both with and without NLL take slightly over a minute for a clean build.

matthewjasper:
I'm getting "extra arguments to `rustc` can only be passed to one target, consider filtering\\nthe package by passing e.g. `--lib` or `--bin NAME` to specify a single target" when I try to run the benchmarks locally, do I need to remove the binary target?

Mark-Simulacrum:
Either that or add a perf-config.json file like https://github.com/rust-lang-nursery/rustc-perf/blob/master/collector/benchmarks/cargo/perf-config.json (probably without the runs: 1 bit).

nnethercote:
> The compile time is much better on the latest nightly. Both with and without NLL take slightly over a minute for a clean build.

I'm glad NLL is much faster, but this is still much too long. [Current benchmarks](https://perf.rust-lang.org/nll-dashboard.html?commit=&stat=wall-time) are all less than 30 seconds on CI, and most are less than 5 seconds. I suggest aiming for around 1-3 seconds on CI.

I care about this partly because of CI time/cost, but mostly because I often run the benchmarks locally under profilers like Cachegrind and DHAT that slow execution by up to 10x. I already have `--exclude script-servo` as a permanent feature of my scripts because it's too long-running (though it's currently [broken](https://github.com/rust-lang-nursery/rustc-perf/issues/289#issuecomment-458139167)). In general I'd like to avoid unnecessary increases to benchmarking times.

PR #349, opened 2019-02-19
Add a new const eval perf problem to the test suite

oli-obk:
This takes around two seconds on my machine, but with a few optimizations I've been able to push it below one second (https://github.com/rust-lang/rust/pull/58556) and I still see many more avenues to optimize it. We can lower the exponent to decrease the compilation time further.

Since this entire test used to take around 16 seconds, I hope the additional 2 seconds are ok for now.


Mark-Simulacrum:
2 seconds seems fine; we try to stay under a hard limit of 30.

I'm not so sure about modifying an existing benchmark though - it'll show up as a regression despite no changes to the compiler. @nnethercote, what do you think?

oli-obk:
could we rename the benchmark? so the old name just stops having data and the new one starts fresh?

Mark-Simulacrum:
Yeah, thinking on this more, that seems like the best option. It might be that we want to long-term add some form of versioning to benchmarks (so we can update whenever and graphs are properly disjoint).

nnethercote:
I agree that renaming is a good idea. Simply adding a '-2' suffix might suffice?

oli-obk:
I've renamed the benchmark

Mark-Simulacrum:
I will try to deploy this later today, thanks!

collector/benchmarks/ctfe-stress-2/src/lib.rs
```diff
@@ -60,3 +60,8 @@ expensive_static!(OPS: i32 = ((((10 >> 1) + 3) * 7) / 2 - 12) << 4; [4 16 16 16
 expensive_static!(RELOCATIONS : &'static str = "hello"; [8 16 16 16 16]);
 expensive_static!(UNSIZE_SLICE: &'static [u8] = b"foo"; [4 16 16 16 16 16]);
 expensive_static!(UNSIZE_TRAIT: &'static Trait = &42u32; [4 16 16 16 16 16]);
+
+// copying all these zeros and the corresponding definedness bits can be expensive and is probably
+// prone to regressions.
+// 24 is an exponent that makes the repeat expression take less than two seconds to compute
+const FOO: [i32; 1 << 24] = [0; 1 << 24];
```
RalfJung: I tried hard to give more ore less reasonable names to the constants above; I think we can do better than `FOO` ;)

PR #361, opened 2019-03-02
Inline rust_sysroot dependency

Mark-Simulacrum:
This moves the sysroot and git installation code into this project to
make it easier to modify and change.


Mark-Simulacrum:
This is a precursor to supporting/downloading Windows/wasm targets and benchmarking those.

PR #375, opened 2019-04-05
Add support for OProfile.

nnethercote:
OProfile uses the same kernel system as perf, but I find its output a lot easier to read.

### DIFF
--- a/collector/src/bin/rustc-perf-collector/execute.rs
+++ b/collector/src/bin/rustc-perf-collector/execute.rs
@@ -5,7 +5,6 @@ use std::collections::{HashMap, HashSet};
 use std::env;
 use std::f64;
 use std::fs::{self, File};
-use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::{self, Command};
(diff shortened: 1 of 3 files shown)


collector/src/bin/rustc-perf-collector/execute.rs
```diff
@@ -471,6 +474,13 @@ impl<'a> Processor for ProfileProcessor<'a> {
             path
         };
 
+        // Write text output to a file.
+        let write_text_to_file = |file, text| {
```
Mark-Simulacrum: Can probably just use https://doc.rust-lang.org/std/fs/fn.write.html.

nnethercote: Hmm, I recall the `flush` being necessary... but local testing suggests that it isn't. Ok, I'll switch to `fs::write`.

nnethercote:
I updated to use `fs::write`. I also added a note to the docs about how OProfile sometimes fails with a broken pipe.

PR #423, opened 2019-07-03
Add link to last run that measured noise

gnzlbg:


gnzlbg:
300 dependencies, so I haven't tested that this builds - I'll let CI do that.

Mark-Simulacrum:
I would put this on the comparison page itself, I believe the source is in `site/templates/compare.html`

gnzlbg:
My HTML is a bit rusty, but that should do.

PR #500, opened 2019-10-07
Bump team repo

pietroalbini:
This PR bumps the team repo to the latest commit. This is needed because a recent change in the repo added a new kind of team, and unfortunately the enum representing it was not exaustive. The bump both adds support for the new kind of team, and makes the enum not exaustive to avoid future issues like this.

r? @Mark-Simulacrum


Mark-Simulacrum:
I take it I'll need to redeploy the site once this lands?

PR #552, opened 2019-11-12
Migrate ctfe-stress-test-3 to ctfe-stress-test-4

davidhewitt:
If rust-lang/rust#66294 lands then evaluation of `const fn` with no arguments will be memoized. This breaks the purpose of the tests in ctfe-stress-3.

This PR changes to a new ctfe-stress-test-4 which is identical save that the const functions now take a dummy u32 argument to avoid memoization.

r? @oli-obk


Mark-Simulacrum:
Does it make sense to keep around a memoized case (though maybe smaller) so we can detect if memoization regresses in the future?

oli-obk:
We already test the memoization in the rustc test suite by causing cycles on recursion

Mark-Simulacrum:
Okay, great!

PR #573, opened 2019-12-09
Remove query cache hits column and update invocations title

### LINKED ISSUE #529: detailed (comparison) view: The `cache misses` column should be removed, `invocations` should be renamed
Since we don't record cache misses for perf.rlo, it does not make sense to show the number of cache misses (it will always be the same as the "# of invocations" right next to it). 

When recording cache misses *and* cache hits, then `invocations` is `hits + misses`. When not recording cache hits, then `invocations` is misleading because it only counts the cache misses. 

On [zulip we settled on "full executions"](https://rust-lang.zulipchat.com/#narrow/stream/187831-t-compiler.2Fwg-self-profile/topic/.22Finish.20the.20MVP.22.20meeting/near/179222392) as the new title for the "invocations" column. This subsumes "query *provider* invocations" and regular (thus uncached) function invocations.

wesleywiser:
Fixes #529

### DIFF
--- a/site/static/detailed-query.html
+++ b/site/static/detailed-query.html
@@ -57,9 +57,7 @@
                     <th data-sort-idx="2" data-default-sort-dir="-1">Time</th>
                     <th data-sort-idx="10" data-default-sort-dir="-1">% Total Time</th>
                     <th class="delta">Δ</th>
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Gah! Totally meant to merge this a while back.

PR #605, opened 2020-01-28
[perf] Avoid copying data on compare page unless it is really necessary

panstromek:
This saves about 700ms on initial load and all subsequent rerenders that have no filters (with the overhead of Chrome profiler).

The best way you can see it is by ticking checkboxes for individual runs (make sure filters are empty) It used to take about a second to rerender after ticking it, now it is almost instant.

There is definitely an opportunity to rewrite this without copying the data at all, which would speed up even filters, but this was literally the simplest low hanging fruit I could do at the moment.

Before:
image

After:
image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -115,15 +115,15 @@
         {
             let filter = document.querySelector("#filter").value;
             let test_names = unique([...Object.keys(data.a.data), ...Object.keys(data.b.data)]);
-            let isCopied = false
+            let is_copied = false;
             for (let name of test_names) {
                 if (!name.includes(filter)) {
-                    if(!isCopied){
+                    if (!is_copied) {
                         // Copying data is very expensive
                         // This branch ensures that we avoid copying
                         // if there are no filters (which is the most common case)
                         data = JSON.parse(JSON.stringify(DATA)); // deep copy
-                        isCopied = true
+                        is_copied = true;
                     }
                     delete data.a.data[name];
                     delete data.b.data[name];
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -115,8 +115,16 @@
         {
             let filter = document.querySelector("#filter").value;
             let test_names = unique([...Object.keys(data.a.data), ...Object.keys(data.b.data)]);
+            let isCopied = false
             for (let name of test_names) {
                 if (!name.includes(filter)) {
+                    if(!isCopied){
+                        // Copying data is very expensive
+                        // This branch ensures that we avoid copying
+                        // if there are no filters (which is the most common case)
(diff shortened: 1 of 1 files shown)


site/static/compare.html
```diff
@@ -115,8 +115,16 @@ <h3>Brief explanation</h3>
...
             let test_names = unique([...Object.keys(data.a.data), ...Object.keys(data.b.data)]);
+            let isCopied = false
             for (let name of test_names) {
                 if (!name.includes(filter)) {
+                    if(!isCopied){
+                        // Copying data is very expensive
+                        // This branch ensures that we avoid copying
+                        // if there are no filters (which is the most common case)
+                        data = JSON.parse(JSON.stringify(DATA)); // deep copy
+                        isCopied = true
```
Mark-Simulacrum:
```suggestion
                        isCopied = true;
```

site/static/compare.html
```diff
@@ -115,8 +115,16 @@ <h3>Brief explanation</h3>
         {
             let filter = document.querySelector("#filter").value;
             let test_names = unique([...Object.keys(data.a.data), ...Object.keys(data.b.data)]);
+            let isCopied = false
             for (let name of test_names) {
                 if (!name.includes(filter)) {
+                    if(!isCopied){
```
Mark-Simulacrum:
```suggestion
                    if (!isCopied) {
```

site/static/compare.html
```diff
@@ -115,8 +115,16 @@ <h3>Brief explanation</h3>
         {
             let filter = document.querySelector("#filter").value;
             let test_names = unique([...Object.keys(data.a.data), ...Object.keys(data.b.data)]);
+            let isCopied = false
```
Mark-Simulacrum:
```suggestion
            let isCopied = false;
```

site/static/compare.html
panstromek: Oops, sorry, I am used to Standard 😄

panstromek:
> Have you verified that this continues to work with all the toggles and such present on the page?
Yes, I tried to open my version against the online one side by side and tried ticking checkboxes and filtering. Everything seems to work as it should.

> Maybe we can do a deep copy in a cheaper way than the JSON parse/stringify, too?
Well, that object has around 5MB, so it will be slow anyway. In that case, it would make more sense to just refactor the code to not copy it at all, because in the end, that copy is kinda pointless - we copy the whole object just to delete most of the data from it anyway (because of filters). But yea, JSON parse/stringify is probably the slowest way to do this.

Mark-Simulacrum:
Thanks! Will deploy soon.

Mark-Simulacrum:
(and merge once ci passes)

PR #618, opened 2020-02-19
Add benchmark with many associated items

ecstatic-morse:
rust-lang/rust#68957 demonstrated an increase in compile times that was quadratic in the number of associated items. rust-lang/rust#68837 sped these cases up a lot, and rust-lang/rust#69072 hopes to speed them up even more by making associated item name lookup O(log n).

This benchmark has been reduced in size from the original example to run in ~2 seconds locally. I was hoping to find a more real-world example that demonstrated the pathological behavior–embedded HAL crates were mentioned–but gave up.


ecstatic-morse:
cc @Mark-Simulacrum

Mark-Simulacrum:
r=me but looks like this needs a rebase, sorry for missing it initially

PR #637, opened 2020-03-25
Upgrade crossbeam-channel to 0.4.2

rylev:
Closes #622


rylev:
@Mark-Simulacrum I'm not sure why but it seems that the Travis run still produces the error that #639 is meant to address, but the run still passed.

PR #640, opened 2020-03-25
Revert "Add benchmark for rust-lang/rust issue 58319"

rylev:
This reverts commit 44c2257ea50b59da0fcf4ff46adf60fd50eeb0d9.

This was the wrong benchmark and the [correct one](https://github.com/rust-lang/rust/issues/65131) cannot be made a benchmark currently because the code does not compile which rustc-perf expects.


Mark-Simulacrum:
I'm inclined to actually just leave this in, it seems like an interesting case regardless (and it'd be annoying to go back and delete it now that it's been in for a while anyway).

PR #645, opened 2020-04-09
Build the stdlib from rust-src sources.

eddyb:
See https://github.com/rust-lang/rust/issues/69060#issuecomment-604928032 for background.

~~I definitely *feels* slower than previous. We might want to build libstd only once and place it in the sysroot (and I believe that would mean we can skip downloading `rust-std` altogether).~~

**EDIT** I moved from `-Zbuild-std` to a custom `cargo build` of the stdlib sources, no slowdown left.

However, we only need this for a couple weeks, until https://github.com/rust-lang/cargo/pull/8073 (hopefully) reaches beta.
Then all of the hacks here will be unnecessary, since Rust itself will be built fully deterministically.

### DIFF
--- a/collector/src/bin/rustc-perf-collector/sysroot.rs
+++ b/collector/src/bin/rustc-perf-collector/sysroot.rs
@@ -1,11 +1,10 @@
 use anyhow::{anyhow, Context};
 use chrono::{DateTime, Utc};
 use collector::Sha;
-use std::ffi::OsStr;
 use std::fmt;
 use std::fs::{self, File};
 use std::io::{BufReader, Read};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use tar::Archive;
 use xz2::bufread::XzDecoder;
 
@@ -42,6 +41,7 @@ impl Sysroot {
         download.get_and_extract(ModuleVariant::Rustc)?;
         download.get_and_extract(ModuleVariant::Std)?;
         download.get_and_extract(ModuleVariant::Cargo)?;
+        download.get_and_extract(ModuleVariant::RustSrc)?;
 
         download.into_sysroot()
     }
@@ -66,14 +66,15 @@ struct SysrootDownload {
     triple: String,
 }
 
-const MODULE_URL: &str =
-    "https://rust-lang-ci2.s3.amazonaws.com/rustc-builds/@SHA@/@MODULE@-nightly-@TRIPLE@.tar.xz";
+const BASE_URL: &str = "https://rust-lang-ci2.s3.amazonaws.com/rustc-builds";
 
+// FIXME(eddyb) rename to just `Component`.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 enum ModuleVariant {
     Cargo,
     Rustc,
     Std,
+    RustSrc,
 }
 
 impl fmt::Display for ModuleVariant {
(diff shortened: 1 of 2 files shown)


collector/src/bin/rustc-perf-collector/sysroot.rs
```diff
@@ -82,47 +83,41 @@ impl fmt::Display for ModuleVariant {
...
-                path.canonicalize().with_context(|| {
-                    format!(
-                        "failed to canonicalize cargo path for {}: {:?}",
-                        self.rust_sha, path
-                    )
-                })?
-            },
+            rustc: sysroot_dir.join("bin/rustc"),
+            rustdoc: sysroot_dir.join("bin/rustdoc"),
+            cargo: sysroot_dir.join("bin/cargo"),
```
Mark-Simulacrum, line 120: So this is actually I believe a behavior change because previously we'd be able to detect that rustc/rustdoc/cargo don't exist and now we won't -- could we add some `assert!(path.exists(), "rustc exists");` calls perhaps? I agree this is clearer.

eddyb: Ahhhh that's what that was for, I didn't know canonicalization errored for non-existant paths.

collector/src/bin/rustc-perf-collector/sysroot.rs
```diff
@@ -42,10 +41,11 @@ impl Sysroot {
         download.get_and_extract(ModuleVariant::Rustc)?;
         download.get_and_extract(ModuleVariant::Std)?;
         download.get_and_extract(ModuleVariant::Cargo)?;
+        download.get_and_extract(ModuleVariant::RustSrc)?;
 
         download.into_sysroot()
+        }
     }
-}
```
eddyb, line 48: Argh, VSCode's git integration misled me (I have some NixOS hacks in here I didn't want to commit :P).

eddyb:
https://github.com/rust-lang/cargo/pull/8073 landed, so we shouldn't need the "fake version" trick anymore.
(well, once `rust-lang/rust` updates `src/tools/cargo` to include that)

eddyb:
This took 5h09m to finish on a 24-core/48-thread server. Somehow I don't think we can do it this way :).

eddyb:
Tried the same run but without `-Zbuild-std`, it finished in 1h20.
So this PR makes perf runs almost *four times* slower. That's unacceptable.

eddyb:
1h23 with the new approach. Also, for fun, I looked inside the `rlib`s built here and saw:
```rust

eddyb:
So the tests I did locally are with a recent noop change [for which we now have official results too](https://perf.rust-lang.org/compare.html?start=835428c35d785733e72bfbf32fc2f8fff3e50e63&end=df768c5c8fcb361c4dc94b4c776d6a78c12862e1&stat=instructions:u).

For "`syn-opt` clean":
| | Before | After | Δ% |
|-|-|-|-|
| `perf.rust-lang.org` | 38,584,396,789 | 37,490,787,593 | -2.8343% |
| This PR (locally) | 37,521,495,334 | 37,521,722,991 | 0.0006% |

So I *believe* this PR should remove the problem (but we should keep an eye on results for a while).

eddyb:
Oddly enough, [this had a beneficial perf effect](https://perf.rust-lang.org/compare.html?start=835428c35d785733e72bfbf32fc2f8fff3e50e63&end=d2230290f7220e740ec08f4d844bf5951e1b74b8&stat=instructions:u).

PR #649, opened 2020-05-12
detailed-query: tweak column titles some more

RalfJung:
I found it strange that the 'Time (s)' delta column says 'Execution time', so I made the column titles consistent. I also merged the two paragraphs that both talk about the '% total time' column.

What I didn't do because I did not know how to is reorder the column such that 'Time delta' is directly to the right of 'Time (s)', like we do with the other columns. IMO it would be better for 'Time (%)' to move either one to the left or one to the right.

### DIFF
--- a/site/static/detailed-query.html
+++ b/site/static/detailed-query.html
@@ -52,8 +52,7 @@
         <a href="dashboard.html">dashboard</a>, <a href="status.html">status</a>.</div>
     <div id="content">
         <h3 id="title"></h3>
-        <p>'Time (%)' is the percentage of the cpu-clock time spent on this query.
-        Totals row is as a comparison to the 'cpu-clock' stat (we do not use
+        <p>'Time (%)' is the percentage of the cpu-clock time spent on this query (we do not use
         wall-time as we want to account for parallelism).</p>
         <p>Executions do not include cached executions.</p>
         <table>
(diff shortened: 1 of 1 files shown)


site/static/detailed-query.html
```diff
@@ -52,17 +52,17 @@
         <a href="dashboard.html">dashboard</a>, <a href="status.html">status</a>.</div>
     <div id="content">
         <h3 id="title"></h3>
-        <p>Totals row '%total time' is as a comparison to the 'cpu-clock' stat (we do not use
+        <p>'Time (%)' is the percentage of the cpu-clock time spent on this query.
+        Totals row is as a comparison to the 'cpu-clock' stat (we do not use
         wall-time as we want to account for parallelism).</p>
-        <p>% total time is the percentage of the cpu-clock time spent on this query.</p>
```
RalfJung, line 57: This text sounds rather redundant now, but I just reordered and merged the previous two paragraphs about the sae column. To me it sounds like it is saying twice that time is relative to 'cpu-clock'? Or is "the cpu-clock time" and "the 'cpu-clock' stat" not the same thing? It is somewhat confusing.

Mark-Simulacrum: Hm, yes. Those are the same thing and especially now that you've merged the two sentences it definitely sounds repeated.

I had intended to have two because one references the value in the header and the other the values in the column, but they're all percentages of the CPU clock, so should be fine to merge them.

RalfJung: Done.

Mark-Simulacrum:
Thanks! Will merge and deploy later today.

PR #658, opened 2020-05-22
Deploy to ECS automatically

Mark-Simulacrum:
This isn't yet fully working.


Mark-Simulacrum:
Okay, I'm going to go ahead and merge this in. It should be much less invasive hopefully than rocksdb into people's workflows as we directly collect into the database now (which has other benefits, such as having hopefully easier access via just sqlite locally, enabling easier ad-hoc analysis). We also no longer only record the minimum into the dataset but rather all measurements taken; I'm hoping to work on displaying the deltas there or somehow showcasing whether some change is "in noise" or not. Some of that will likely involve trying to figure out what we can do given the limited amount of data points we have, though.

I've updated documentation in the various readme files, but to summarize the steps to migrate from older master and current master are pretty much the same - run the ingest-json script as before, doing something like `find rustc-timing/times/ -type f | xargs ./target/release/ingest perf-rlo.db finished-files/`, and then you should be able to just launch the site directly off of that database (site perf-rlo.db). It won't update on new measurements, but hitting the /perf/onpush endpoint will reload things. I hope to eventually remove that step too, but still figuring out the best way to do so :)

cc @nnethercote @eddyb on the above as you both have collection setups locally. I'll be taking a look at getting self-profiling graph PR merged in soon.

PR #660, opened 2020-05-23
Support --builds option in bench_commit subcommand.

eddyb:
Could probably share more flags but I'm not sure what the best way to do that would be.


Mark-Simulacrum:
I believe this'll be made obsolete by https://github.com/rust-lang/rustc-perf/pull/687 which unifies bench_commit and bench_local (and kinda drops support for bench_commit I believe, requiring local installation through RTIM or similar?)

PR #661, opened 2020-05-25
Add a way to graph self-profiling data.

eddyb:
I've tried to test this on the `rustc-timing` data (as opposed to the single-benchmark dataset I'm independently collecting), but it's too slow and times out (in ngrok, but it'd be unusably slow either way).

So I'm mostly opening this up looking for feedback. Maybe this is impossible without the recent changes on master (this is based on a week-old checkout).

Also I'm not sure how to best expose this in the UI, or if the way I made it configurable makes sense.
(Example URL: `?stat=self-profile:incr-full:top5` for top 5 queries for every `incr-full` run)

<hr/>

**EDIT**: by filtering for one crate and only `-check`, I was able to get the times down. Also, `end` is broken? Anyway, here are some screenshots (for `start=2020-03-25`, i.e. last 2 months):

#### Existing graphs: (unchanged by this PR, only shown as baselines)
* `stat=instructions:u`

  image

* `stat=wall-time` (noisier than `instructions:u`)

  image

#### New graph: (added in this PR)

* `stat=self-profile:incr-full:top10` (can be noisier than `wall-time` but mostly spikes up, so if you look "below" the graph, the baseline shape is visible)

  image

### DIFF
--- a/site/src/selector.rs
+++ b/site/src/selector.rs
@@ -122,7 +122,7 @@ pub trait Series<'a>: Sized {
         -> Result<Self, String>;
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Path {
     path: Vec<PathComponent>,
 }
--- a/site/src/server.rs
+++ b/site/src/server.rs
@@ -369,6 +369,156 @@ where
 }
 
 pub async fn handle_graph(body: graph::Request, data: &InputData) -> ServerResult<graph::Response> {
+    if body.stat.starts_with("self-profile:") {
+        let mut parts = body.stat.split(':').skip(1);
+        let cache_selector = parts.next().unwrap();
+        let filter = parts.next().ok_or(format!(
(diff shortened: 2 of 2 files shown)


rylev:
This is very interesting. Should we try to land this?

Mark-Simulacrum:
I have no objections to trying to land this. The database, in theory, is prepared to handle these queries as effectively as those for the pstat table -- it's the reason we load self profile information into the database essentially.

Adding this is likely a good exercise in plumbing through the frontend bits of adding a new metric to the frontend graphs.

Mark-Simulacrum:
It may also be helpful in the eventual desire for showing noise margins on the self-profile pages -- right now they're essentially useless when comparing unless a very large delta is present. Moving to instruction counts would likely help with that too, though.

rylev:
I'm going to close this for since no one is currently working on it. We can revive this in the future.

PR #672, opened 2020-06-11
Fix toml to drop README

Mark-Simulacrum:


Mark-Simulacrum:
This seems to still be affected by https://github.com/rust-lang/cargo/issues/8351 but regardless, merging, as at least it fixes CI somewhat.

PR #674, opened 2020-06-23
collector: passthrough env vars

davidtwco:
This PR modifies the collector to stop clearing the environment when invoking the fake rustc.

This is helpful for users of Nix development shells where environment variables are required to inform tools of the location of C dependencies, and for setting `RUSTFLAGS` to collect data from the same build with different flags.

However, this change can result in some environment variables affecting the data collection where that isn't intended.


Mark-Simulacrum:
I'm going to go ahead and merge this -- I think it's probably the right thing to do. I wonder if we should instead have some sort of "allowed env list" in the arguments passed to the executable... but that seems pretty annoying for users who do want some stable set of env variables.

PR #679, opened 2020-07-03
Avoid busting caches on path dependencies

Mark-Simulacrum:
Use `cp -p` to avoid losing mtimes on copy.


Mark-Simulacrum:
Some sort of error in the surfman git dep -- not sure what's up with that -- I'll just wait until later to retry, I guess, maybe github is having problems or something.

PR #694, opened 2020-07-19
Reduce runtime further

Mark-Simulacrum:
This vendors mozjs which takes a really long time (160s+ in check) to build, pre-running bindgen and copying into sources. The generated bindings may not be correct for all platforms, but we should hopefully not care as we're not actually intending to run servo, just compile it.

This shaves 20 minutes (53 -> 33 minutes) from the script-servo CI builder and essentially no time from the other CI builder; this is basically as expected.


Mark-Simulacrum:
This brought us down 10 minutes on the collector, from 2:40 to 2:30 per run, roughly, though we've just had the one benchmark since this deployed.

PR #697, opened 2020-07-20
Skip doc build-kind on 1.45.0 and lower versions

### LINKED ISSUE #696: Doc build failed on 1.45.0
The beta run succeeded, but 1.45.0 was unable to be benchmarked. I suspect this is because of https://github.com/rust-lang/cargo/pull/8449.

Mark-Simulacrum:
Fixes #696

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -97,6 +97,15 @@ where
     Ok(Option::deserialize(deserializer)?.unwrap_or(0.0))
 }
 
+pub fn version_supports_doc(version_str: &str) -> bool {
+    if let Some(version) = version_str.parse::<semver::Version>().ok() {
+        version >= semver::Version::new(1, 46, 0)
+    } else {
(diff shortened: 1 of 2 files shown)


collector/src/lib.rs
```diff
@@ -97,6 +97,15 @@ where
     Ok(Option::deserialize(deserializer)?.unwrap_or(0.0))
 }
 
+pub fn version_supports_doc(version_str: &str) -> bool {
+    if let Some(version) = version_str.parse::<semver::Version>().ok() {
+        version >= semver::Version::new(1, 46, 0)
+    } else {
+        assert!(version_str.starts_with("beta") || version_str.starts_with("master"));
```
jyn514, line 104: What about `"nightly"`?

Mark-Simulacrum: We don't intend to benchmark nightlies with bench_published (and this matches the incremental check above).

PR #720, opened 2020-07-26
Updates favicon

sl4m:
This simply updates the perf subdomain's favicon to be inline with the other subdomains on rust-lang.org.

Adds the following:

* a dark mode compatible SVG favicon
* a PNG favicon as fallback


Mark-Simulacrum:
Thanks!

PR #729, opened 2020-07-27
Bump Ubuntu to 20.04 for profiler tests

Mark-Simulacrum:
This gets us Valgrind 3.15 for free, which enables a few tests. See https://github.com/actions/virtual-environments/tree/master for documentation on what else changes in 20.04 (and the other builders available to us).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
 
   test_profiling:
     name: Test profiling
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
 
   test_profiling:
     name: Test profiling
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
Mark-Simulacrum, line 85: Don't ask me why GitHub chose to have a latest != the actual latest Ubuntu, doesn't make sense to me.

PR #732, opened 2020-07-28
Benchmark oldest commits first

Mark-Simulacrum:
This means that we have less cases of "straggler" holes being left behind. It can still happen, due to try parents being prioritized above anything else -- but mostly non-single commit gaps will be left.


Mark-Simulacrum:
cc @nnethercote -- this implements what we discussed yesterday

PR #733, opened 2020-07-29
Prepare build kinds in parallel

Mark-Simulacrum:
rustc is single threaded for a good portion of its compilation, and that means
that on many-core systems we're frequently only using one CPU core even during
preparation (where we don't care about noise). This builds all builds kinds in
parallel -- spawning N Cargos. We arrange for a jobserver to be passed to all
Cargos involved to make sure that we're not overloading the system with too many
processes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,13 +37,12 @@ jobs:
     strategy:
       matrix:
         BENCH_INCLUDE_EXCLUDE_OPTS: [
-          "--exclude script-servo,webrender-wrench,style-servo",
-          "--include webrender-wrench,style-servo",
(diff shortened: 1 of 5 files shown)


Mark-Simulacrum:
It is non-obvious whether this will be a win on the collection server; it's plausible that it won't be (if we have insufficient cores, basically). But it did show a (small) win locally, and it allows for us to merge our CI down to less builders.

Worst case -- if this is a regression -- or even if it is neutral, we can revert it. It should not have any effect on the final crate compilations (i.e., stats on the site).

Mark-Simulacrum:
This shaved 3.5 minutes off our builds, roughly, though we've only had the one since this merged. I consider that enough for the additional complexity to be worth it.

PR #735, opened 2020-08-02
Remove PR patch from script-servo

Mark-Simulacrum:
This patch adds a significant chunk of time to our longest benchmark, and doesn't add too much value in practice.

cc @nnethercote


Mark-Simulacrum:
This was a 6 minute win on the collection server, pretty much as expected (each script-servo build takes roughly 1 minute, and this saved 6 of those: 3 self profile, 3 w/o).

PR #738, opened 2020-08-04
Share target directories between benchmarks

Mark-Simulacrum:
Previously, each benchmark would rebuild dependencies from scratch, which was
somewhat inefficient: many dependencies in Rust are shared amongst crates (e.g.,
syn, serde). This shares them across all benchmarks.

The next step would be to go further, and share between build kinds (e.g., build
scripts only need to be built once). But we don't actually have a good way of
just building build scripts, so that's left out of scope for now. It is also unclear
how big a win this would be.

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -391,7 +391,7 @@ impl<'a> CargoProcess<'a> {
             if let Some((ref mut processor, run_kind, run_kind_str, patch)) = self.processor_etc {
                 let data = ProcessOutputData {
                     name: self.processor_name.clone(),
-                    cwd: self.target_directory,
+                    cwd: self.src_dir,
                     build_kind: self.build_kind,
                     run_kind,
                     run_kind_str,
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
I am working on gathering local numbers for how much of a win this is -- it seems like it might actually be a loss (perhaps from increased copying) on CI...

Mark-Simulacrum:
I am struggling to gather local numbers (still getting environment setup on a new computer), but I suspect that regardless the discrepancy in available parallelism and such may mean that I wouldn't really reflect what happens in prod anyway. I'm going to merge this. It is passing CI, and it should not affect the final crate's builds. In the worst case we can revert and delete generated artifacts if something does go wrong unexpectedly.

Mark-Simulacrum:
Okay this does indeed seem to be a regression:
 * 01:47:05 before this landed
 * 01:51:52 after this landed

Nothing obviously slower landed in rustc in that time either. I am unsure yet if there's some room for wins here or if we should just revert.

PR #739, opened 2020-08-04
Revert "Share target directories between benchmarks"

Mark-Simulacrum:
Reverts rust-lang/rustc-perf#738

It appears to be a performance regression (up 5 minutes on the collector).

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -198,8 +198,7 @@ impl Profiler {
 
 struct CargoProcess<'a> {
     compiler: Compiler<'a>,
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
I'm going to hold off on merging this until I finish a local collection with this on -- without, I see a full build taking 4014 seconds. If this is faster locally then I'll see if we can't reproduce that on the collector somehow before landing this.

Mark-Simulacrum:
4014.562306732s w/o shared cache
4160.392478718s w/ shared cache

Reverting.

PR #744, opened 2020-08-10
initial uPlot & JSON-based graph page prototype. (#742)

leeoniya:
this is a mostly-complete prototype for a faster graphs page.

unfortunately, it was taking too long to do proper integration into the existing codebase, so i'll have to leave that legwork up to you guys. i'm happy to answer follow-up questions regarding any missing bits you may still want.

couple notes:

- zoom reset works on double-click, but you can add a button and reset the scales programmatically, as well
- if you don't like the closest-series-focus effect, just set the `focus.alpha` to `1` instead of `0.3`
- while i did add handling for changing the tooltip border color for `is_interpolated`, uPlot does not support per-datapoint color customization (e.g. pink), but you have full access to `u.ctx`, and methods for getting those datapoint coordinates to draw annotations or whatever at those locations. if you really want those pink datapoints, then it's possible to override the point rendering completely and DIY that as well: https://leeoniya.github.io/uPlot/demos/draw-hooks.html
- auto-resize is expensive as it requires a full redraw, so it's not there by default. if you need it, you can always add it: https://leeoniya.github.io/uPlot/demos/resize.html


Mark-Simulacrum:
This looks like a good start.

I do think we want to show points on the graphs (circles or whatever is fine). I'm not quite sure how to use the draw hooks API -- there doesn't seem to be any documentation? I tried to pull off of https://github.com/leeoniya/uPlot/blob/master/demos/draw-hooks.html but I'm not having much success. AFAICT the diff here doesn't produce any difference at all...

```diff
diff --git a/site/static/index2.html b/site/static/index2.html
index 001514c2..e3d1f4e0 100644
--- a/site/static/index2.html
+++ b/site/static/index2.html
@@ -179,6 +179,38 @@
                     },
                 ],
                 plugins: [
+                    {
+                        opts: (u, opts) => {
+                            opts.series.forEach((s, i) => {
+                                if (i > 0) {
+                                    uPlot.assign(s, {
+                                        points: {
+                                            show: (self, i, i0, i1) => {
+                                                let ctx = self.ctx;
+                                                let { stoke, scale } = self.series[i];
+                                                for (let j = i0; i < i1; i++) {
+                                                    let val = self.data[i][j];
+                                                    let cx = Math.round(self.valToPos(u.data[0][j], 'x', true));
+                                                    let cy = Math.round(self.valToPos(val, scale, true));
+                                                    let x = cx;
+                                                    let y = cy;
+                                                    let outerRadius = 1000;
+
+                                                    ctx.beginPath();
+                                                    ctx.moveTo(cx, cy - outerRadius);
+                                                    ctx.arc(cx, cy, outerRadius, 0, Math.PI * 2);
+                                                    ctx.fill();
+                                                    ctx.closePath();
+
+                                                }
+                                                return false;
+                                            },
+                                        }
+                                    });
+                                }
+                            });
+                        }
+                    },
                     tooltipPlugin({
                         onclick(u, seriesIdx, dataIdx) {
                             let thisCommit = commits[dataIdx][1];

```

leeoniya:
showing points is by default conditioned on their density, so they *do* show up at higher zoom levels:

https://github.com/leeoniya/uPlot/blob/e50c20a0966fc27d9dd0a18eec53321991115658/src/opts.js#L454-L459

to show them unconditionally (which does eat into the perf), you can override `series.points.show`:

```js
seriesOpts.push({
    label: cacheState,
    width: devicePixelRatio,
    stroke: seriesColors[j],
    points: {
        show: true,
    },
});
```

this will reduce the need to draw them yourself, unless you still want to do this to show the pink points?

Mark-Simulacrum:
I do still want pink points.

leeoniya:
well, your inner `j` loop is messed up. why are you testing and incrementing `i`?

try this:

```js
{
    opts: (u, opts) => {
        opts.series.forEach((s, i) => {
            if (i > 0) {
                uPlot.assign(s, {
                    points: {
                        show: (self, i, i0, i1) => {
                            let ctx = self.ctx;
                            let { stoke, scale } = self.series[i];
                            for (let j = i0; j < i1; j++) {
                                let val = self.data[i][j];
                                let cx = Math.round(self.valToPos(u.data[0][j], 'x', true));
                                let cy = Math.round(self.valToPos(val, scale, true));
                                let x = cx;
                                let y = cy;
                                let outerRadius = 4;

                                ctx.beginPath();
                                ctx.fillStyle = isInterpolated(j) ? interpolatedColor : u.series[i].stroke;
                                ctx.moveTo(cx, cy - outerRadius);
                                ctx.arc(cx, cy, outerRadius, 0, Math.PI * 2);
                                ctx.fill();
                                ctx.closePath();
                            }
                            return false;
                        },
                    }
                });
            }
        });
    }
},
```

leeoniya:
my suggestion to maintain perf is to leave the point rendering density-dependent and series-colored, and simply identify runs of `is_interpolated` and draw shaded rect regions in those x intervals on the canvas. it will be much cheaper and still properly highlight those areas.

similar to how the medians plugin works here: https://leeoniya.github.io/uPlot/demos/draw-hooks.html

leeoniya:
this is how i'd do it:

```js
plugins: [
    {
        hooks: {
            drawAxes: [
                u => {
                    let { ctx } = u;
                    let { left, top, width, height } = u.bbox;

                    const interpolatedColorWithAlpha = "#fcb0f17a";

                    ctx.save();
                    ctx.strokeStyle = interpolatedColorWithAlpha;
                    ctx.beginPath();

                    u.data[0].forEach((v, j) => {
                        if (isInterpolated(j)) {
                            let cx = Math.round(u.valToPos(v, 'x', true));
                            ctx.moveTo(cx, top);
                            ctx.lineTo(cx, top + height);
                        }
                    });

                    ctx.closePath();
                    ctx.stroke();
                    ctx.restore();
                },
            ]
        },
    },
```

Mark-Simulacrum:
This looks great!

image


I've left the old index.html around but will merge this once CI passes (though it doesn't touch anything CI checks really so that shouldn't be a problem). If no one notes any problems we can get rid of the old copy soon!

leeoniya:
nice :)

nnethercote:
I agree that it would be nice to have a dot for each point.

leeoniya:
if you want to render points unconditionally (regardless of density) you can simply set `series.points.show: true`. this doesnt give you control over per-point styling but it's cleaner than doing DIY points in a hook.

Mark-Simulacrum:
Yeah, I'm still trying to figure out how I feel about it. I think I do want the dedicated points though...

leeoniya:
> I agree that it would be nice to have a dot for each point.

@nnethercote the dots do show up at higher zoom levels when their density is sufficiently sparse. this can be tuned by providing a custom `series.points.show`. the default implementation is here: https://github.com/leeoniya/uPlot/blob/5d1f92a511fe88c75947b19feee6328bf6b78e6a/src/opts.js#L470-L475. at high densities with constant overlap i guess i don't see what they add to the story that a larger stroke width does not accomplish with a much cheaper render.

image

PR #752, opened 2020-08-25
Avoid common failure modes

Mark-Simulacrum:
This changes two things that @jonas-schievink hit recently:
 * Changes the default to replacing, rather than appending to existing collected data (on CI, we still append)
 * Adds the default database path to the server, matching the collector.

### DIFF
--- a/collector/src/main.rs
+++ b/collector/src/main.rs
@@ -183,6 +183,13 @@ impl BenchmarkErrors {
     }
 }
 
+#[derive(Copy, Clone, Debug)]
+struct BenchOpts {
+    iterations: usize,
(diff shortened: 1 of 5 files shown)


Mark-Simulacrum:
One detail pointed out on Zulip is that if we by default overwrite, the chances of deleting just-collected data are pretty high. It's not entirely clear what a good solution is yet, though.

jyn514:
I would be happy with a way to opt-in to overwrite, instead of making it the default. Right now the only way to do that is to manually mess with the sqlite database.

rylev:

PR #759, opened 2020-09-14
Fix some fallout from #758

ecstatic-morse:
This brings back the colored tooltip border and swaps the canonical colors for `incr-unchanged` and `incr-patched: println` to match those on the "Dashboard" page.

### DIFF
--- a/site/static/index.html
+++ b/site/static/index.html
@@ -102,7 +102,10 @@
                 tooltip.style.top  = (tooltipTopOffset  + top + shiftX) + "px";
                 tooltip.style.left = (tooltipLeftOffset + lft + shiftY) + "px";
 
-                tooltip.style.borderColor = isInterpolated(dataIdx) ? interpolatedColor : "#000000";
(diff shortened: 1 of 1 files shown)


ecstatic-morse:
@Mark-Simulacrum I am trying to test these changes locally, but the current database is ~7GB and after downloading it all the graphs on `index.html` are blank even though the "Dashboard" shows up. Maybe it was corrupted somehow? Should I try generating data locally?

leeoniya:
@ecstatic-morse you can test the changes against the live db by simply changing `BASE_URL` in `/site/static/shared.js`.

see: https://github.com/rust-lang/rustc-perf/issues/742#issuecomment-670926209

Mark-Simulacrum:
Yeah, the database stopped getting updated because of its size -- I've been thinking about how to provide smaller databases that are still useful. Probably we'll just exclude the self profile table and that'll shrink the size considerably, but for now what @leeoniya suggests is probably the best bet.

PR #763, opened 2020-09-19
Add script to generate weekly report

ecstatic-morse:
This is barely even programming; It's just an API scraper. We probably want to integrate this functionality into the collector, but this was much easier for me.

### DIFF
--- a/triage/weekly_report.py
+++ b/triage/weekly_report.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+
+from datetime import date
+from math import log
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Hm, could we (optionally) accept an end commit and add a test for this to CI? Just running it on some known working range and checking against canonical output?

This looks like a great start though, excited to have some automation here.

Mark-Simulacrum:
I'm going to go ahead and merge this, it seems like we can add tests and such later :)

PR #769, opened 2020-09-23
Add match-stress-enum benchmark

ishitatsuyuki:
See https://github.com/rust-lang/rust/pull/76918 for the PR that makes this compile faster.


Mark-Simulacrum:
How long does this take currently and after that PR? (Locally is fine)

ishitatsuyuki:
> Old rustc takes around 2.7s, this build takes around 1.7s.

https://github.com/rust-lang/rust/pull/76918#issuecomment-695366829

Doesn't take too much time to run since this is about quadratic behavior rather than exponential.

PR #787, opened 2020-10-28
detailed-query page: add links to load the crox profile in speedscope and the FF profiler

lqd:
I've tested this locally as we discussed, on [an older PR](https://perf.rust-lang.org/compare.html?start=de32266a1780aa4ef748ce7f6200a1554fad0aca&end=4e4e6f15b872a60816b90eed0c8a72b153d4460c), so I hope it works deployed on perf.rlo.

(Technically, we could *also* use the FF profiler to compare two profiles, but it had errors when I tried it. I'll leave that for the future)

### DIFF
--- a/site/static/detailed-query.html
+++ b/site/static/detailed-query.html
@@ -128,14 +128,34 @@
                 let url = dl_url(commit, bench, run);
                 return `<a href="${url}">raw</a>`;
             };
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Looks great! I'm going to go ahead and merge this, we can always iterate on the fine details as needed.

PR #788, opened 2020-10-29
Wrap `analyzeme` 0.7.0 and 9.0.0 under a single API

lqd:
This PR tries to wrap both versions of `analyzeme` used by the site to process the self-profile data more easily.

As we discussed as a first WIP step, `cargo check`-quality: I was unable to test it, so I preferred opening a draft PR. I made assumptions on being able to infer the version from the names of files in the tarballs, so I hope it works.

I've added both dependencies: the same 0.7.1 revision used before the PR, and master, as the released 9.0.0 doesn't contain your recent `paged_buffer` changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,10 +21,23 @@ dependencies = [
 [[package]]
 name = "analyzeme"
 version = "0.7.1"
-source = "git+https://github.com/rust-lang/measureme#de3edc5cfb6961580dc04b720de963510c209fc4"
+source = "git+https://github.com/rust-lang/measureme?rev=de3edc5cfb6961580dc04b720de963510c209fc4#de3edc5cfb6961580dc04b720de963510c209fc4"
 dependencies = [
  "byteorder",
- "measureme",
+ "measureme 0.7.1",
+ "memchr",
+ "rustc-hash",
+ "serde",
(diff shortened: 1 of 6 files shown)


rylev:

tgnottingham:
@lqd, do you want to continue forward with this? If so, I can help review. If not, I might take inspiration from it. I had implemented #978 without seeing your PR first, but yours is a more complete solution.

lqd:
IIRC the latest status was: the APIs were unified and both worked at a low-level, but the high-level entry points, eg differentiating between the two different formats and stored archive files, were not fully done yet (because I didn't know how this worked and had no way of testing at the time). 

(There's currently also an incorrect assumption in the PR, about which version to use depending on the layout of the archive file, but that can be easily fixed)

@Mark-Simulacrum hoped to help for the last few bits required to land this, if time permitted, but we never did do so.

If either of you have time to help with this, that'd be cool. Feel free to review, take these commits, or tell me what to do to complete this. I really want to have the processed links back on perf.rlo :)

tgnottingham:
Well, #978 ended up getting merged, so the processed links work again. :) The old format isn't supported anymore, though. I think that's unlikely to be a problem in practice, but if it becomes one, we can address it.

PR #791, opened 2020-11-05
Add builtin derive benchmarks

rylev:
**tl;dr Using #[derive] for built in traits incurs significant overhead at scale**

I've created a benchmark which has 5,000 structs with one i32 field. I then `#[derive(Copy, Clone, Default, Eq, PartialEq, Debug)]` for all the structs. This builds using the current nightly compiler (f2bbdd0a3 2020-11-04) in ~5 seconds. When I manually implement these traits, the code compiles in ~2.5 seconds. 

This issue is being discussed [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance).


Mark-Simulacrum:
It may make sense to delete the Copy derive (at least in say half the cases); Copy as a derive makes the Clone impl much simpler and likely much cheaper too.

joshtriplett:
for the sake of making it easier to see optimizations or pessimizations that affect a series of derives on the same type, could you:

1) Add `Ord`, `PartialOrd`, and `Hash`, which rounds out the complete set of derivable traits.
2) Change half of these to write the derives separately (`#[derive(Debug)] #[derive(Default)] #[derive(PartialEq)]` ...)?

(This is in addition to @Mark-Simulacrum's suggestion.)

rylev:
@joshtriplett `rustsmt` collapses separate derives into one line. Obviously not everyone uses `rustfmt`, but maybe this makes it a bit less compelling to test that case?

joshtriplett:
@rylev My only concern was catching potential regressions, but it isn't critical, and as you said, rustfmt makes it unlikely anyone will hit that case.

PR #805, opened 2020-12-01
Don't delete the temporary directory when an error occurs

jyn514:
Deleting the directory makes it really hard to debug what caused the error.


Mark-Simulacrum:
I think CI is failing because packed-simd is broken -- see https://perf.rust-lang.org/status.html -- I haven't had time to investigate or fix that yet.

PR #809, opened 2020-12-08
767: triage; fetch pr titles

Horki:
- fixed typo ~prefereable~ into preferable
- msg: format ex.: "Fix SGX CI, take 3 [#79797](https://github.com/rust-lang/rust/issues/79797)"

### DIFF
--- a/triage/weekly_report.py
+++ b/triage/weekly_report.py
@@ -178,7 +178,8 @@ def gh_pr_title(pr):
         url = f'https://api.github.com/repos/rust-lang/rust/pulls/{pr}'
         req = urllib.request.Request(url)
         req.add_header('Content-Type', 'application/json')
-        req.add_header('Authorization', f'token {getenv("GITHUB_TOKEN")}')
+        if getenv("GITHUB_TOKEN") is not None:
+            req.add_header('Authorization', f'token {getenv("GITHUB_TOKEN")}')
 
         with urllib.request.urlopen(req) as f:
             data = json.loads(f.read())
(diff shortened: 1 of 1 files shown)


triage/weekly_report.py
```diff
@@ -172,6 +174,17 @@ def gh_link(pr):
     return f'https://github.com/rust-lang/rust/issues/{pr}'
 
 
+def gh_pr_title(pr):
+    if pr in pr_titles:
+        return pr_titles.get(pr)
+
+    url = f'https://api.github.com/repos/rust-lang/rust/pulls/{pr}'
```
Mark-Simulacrum, line 181: I think unfortunately this is going to be pretty error prone -- we land 7-10 PRs/day (without rollups, which wouldn't be noted here regardless). That means that over a 7 day period we'd have roughly ~50-55 PRs, and GitHub's rate limit without authentication is 60 requests per hour. Even running the script twice wouldn't be possible. Would you be up for adding authentication here? It should be mostly a matter of setting the "Authorization" header to something like "token $GITHUB_TOKEN", which is optionally present in the environment.

Horki: @Mark-Simulacrum, you are right, will fix it;
- besides limit 60 req/h; is msg format ok? "Fix SGX CI, take 3 #79797" `'{title} [#{pr}]({gh_link(pr)})'`

Mark-Simulacrum: Yes, that seems fine, thanks!

Horki:
@Mark-Simulacrum In case of 401, or other error, it returns an empty string; fetching `GITHUB_TOKEN` via os.getenv

Horki:
@Mark-Simulacrum please review, and add suggestions

Mark-Simulacrum:
Thanks! The CI failure is due to upstream problems with a rustdoc bug, and this looks good to me, so merging.

PR #810, opened 2020-12-15
Add `S-waiting-on-perf` once perf run is queued

camelid:
Also change a use of `@rustbot modify labels` to use `@rustbot label`.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -76,7 +76,7 @@ pub async fn handle_github(
         post_comment(
             &data.config,
             request.issue.number,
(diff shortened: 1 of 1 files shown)


site/src/github.rs
```diff
@@ -525,7 +525,9 @@ async fn enqueue_sha(
     };
     if queued {
         let msg = format!(
-            "Queued {} with parent {}, future [comparison URL]({}).",
+            "Queued {} with parent {}, future [comparison URL]({}).\
```
camelid, line 528: Actually there shouldn't be a `\ ` here.

PR #821, opened 2021-01-06
update doc references to bytecodealliance repos

rgrmrts:
Updating some stale references to Cranelift repositories in the README.


Mark-Simulacrum:
hm, these aren't really intended to be kept up to date - generally the benchmarks are snapshots in time of the crate, and I'd rather not keep them up to date in terms of links like this; I wouldn't expect folks to be reading them though so I'm a bit surprised to see this PR - can you say more about why you wanted this?

rgrmrts:
Makes sense, thanks!

I was making this change https://github.com/bytecodealliance/cranelift-jit-demo/pull/53 and wanted to update any references I found. Seems like this particular reference is unnecessary though. Closing!

PR #822, opened 2021-01-08
Update 2021-01-05 with results of investigations into rollup

rylev:


Mark-Simulacrum:
Hm, I thought we had identified the cause (well which PR) for both?

rylev:
@Mark-Simulacrum we did. I just forgot an "s" at the end of the issue title, but I updated the report with both.

Mark-Simulacrum:
ah perfect, ok, merging

PR #829, opened 2021-01-23
Update colors for slightly positive and negative

bjorn3:
The colors were a bit too faint I think.

Before:

Screenshot_2021-01-23 rustc performance data

After:

Screenshot_2021-01-23 rustc performance data(3)

### DIFF
--- a/site/static/perf.css
+++ b/site/static/perf.css
@@ -87,13 +87,13 @@ img {
   color: red;
 }
 .slightly-positive {
-  color: #ee9090;
+  color: #ff5050;
 }
 .negative {
   color: green;
 }
 .slightly-negative {
-  color: #90ee90;
+  color: #10ff10;
 }
 .neutral {
   color: black;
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Hm, those look a bit too bright and stand out more than the dark green...

bjorn3:
Maybe

image

```css
.slightly-positive {
    color: #ff5050 !important;
}
.slightly-negative {
    color: #30ff30 !important;
}
```

Mark-Simulacrum:
That still looks like it'll stand out more. Maybe we should just leave them black? I would rather add a toggle to hide results <0.1/0.2 magnitude or so.

bjorn3:
Without colors it is much harder to find the positive and negative changes in a quick glance.

tgnottingham:
How about something like this?

rustc_perf_colors

```css
.slightly-positive {
    color: #ea7272;
}
.slightly-negative {
    color: #74b374;
}
```

Mark-Simulacrum:
That looks much more reasonable to me.

bjorn3:
Yeah, that looks much better.

PR #838, opened 2021-02-06
Happy birthday to the last noise run

Julian-Wollersberger:
The old noise run is from 2020-02-06, which was exactly a year ago. 
The new noise run from 2021-02-06 measures the non-functional changes from rust-lang/rust#81792 "Bump nightly version to 1.52.0"

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -27,7 +27,7 @@ input[type="checkbox"] {
         <p>Warning: although measurements known to have high variation are marked with
             '?'/'??', this does not mean that unmarked measurements are guaranteed to have
             low variation. Verify the measurement against the
-            <a href="compare.html?start=333c32a5a4a51cae562c47e0669bc5aeaf741c45&end=1f8df2508f2772d83011f0f651de86181123e519&stat=instructions:u">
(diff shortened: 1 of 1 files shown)


bjorn3:
I don't think a version bump is the most suitable. It likely affects iteration order of multiple hashmaps, thereby introducing noise that doesn't nornally happen.

Mark-Simulacrum:
Yes, I agree. It would be better to find something affecting only e.g. README.md or otherwise not even touching .rs files that are part of the libraries or compiler. (Something like a cargo bump would also work).

Julian-Wollersberger:
> I don't think a version bump is the most suitable. It likely affects iteration order of multiple hashmaps, thereby introducing noise that doesn't nornally happen.

Oh, I didn't think of that. That's an interesting consideration, thanks for pointing it out. 
I'll find another change when I have time, probably next weekend.

Mark-Simulacrum:
Thanks. This one looks good to me.

Julian-Wollersberger:
Wow, you were fast ^^
So for reference, this is from rust-lang/rust#81666.

PR #840, opened 2021-02-08
perf.rust-lang.org database is no longer being published

### LINKED ISSUE #781: Site/README: data now more than 7GB
Just a note for updating the size of data fetched from the server on perf.rust-lang.org in README.

tmiasko:
Closes #781.


Mark-Simulacrum:
We stopped regularly publishing this actually, so it would probably be best to drop this text entirely.

PR #845, opened 2021-02-26
fix time diff

klensy:
Currently at https://perf.rust-lang.org/status.html
time shown as
```
No current collection in progress. Last one finished at 2/26/2021, 7:30:10 PM local time, -14m0-31s ago.
```
so fixing last portion to real time diff.

### DIFF
--- a/site/static/status.html
+++ b/site/static/status.html
@@ -118,7 +118,7 @@
                 let end = new Date(data.most_recent_end * 1000);
                 element.innerHTML = `No current collection in progress. Last one
                     finished at ${end.toLocaleString()} local time,
-                    ${format_duration(Math.trunc((new Date()/1000 - end)))} ago.`;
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Hm, have you tested this locally? This has bothered me for a while, but it feels like moving the division isn't the right fix - I'd expect flipping end and the current time to be closer to the right fix?

klensy:
You can check values: `end` in milliseconds, `new Date()` in milliseconds too, but `new Date()/1000 - end` will be negative, you can check that in js debugger.

klensy:
```js
format_duration(Math.trunc((new Date()/1000 - end)))
" -55m0-45s"
```
```js
format_duration(Math.trunc((new Date() - end)/1000))
"1h20m57s"
```

Mark-Simulacrum:
Hm, ok. Thanks. CI is a different bug, unrelated to this, so going to go ahead and merge.

PR #846, opened 2021-03-02
Add basic help page to document rust-timer commands

camelid:
Both I and `@jyn514` had no idea that `@rust-timer queue exclude`
existed because there's no documentation, or at least I wasn't able to
find any publicized documentation.

r? @Mark-Simulacrum
cc @jyn514

### DIFF
--- a/site/static/help.html
+++ b/site/static/help.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>rustc performance data</title>
+<link rel="stylesheet" type="text/css" href="perf.css">
+<link rel="alternate icon" type="image/png" href="/favicon-32x32.png">
+<link rel="icon" type="image/svg+xml" href="/favicon.svg">
+<style>
+    .help-content {
+        font-family: Helvetica, Arial, sans-serif;
+        line-height: 140%;
+        font-size: 16px;
+        max-width: 50em;
+    }
+    .help-content code {
+        background: #eee;
+        border-radius: 5px;
+        padding: 2px;
+    }
+</style>
+</head>
+<body class="container">
+    <div>&gt; <a href="index.html">graphs</a>, <a href="compare.html">compare</a>,
+        <a href="dashboard.html">dashboard</a>, <a href="bootstrap.html">bootstrap</a>,
+        <a href="status.html">status</a>, <a href="help.html">help</a>.</div>
+    <div style="text-align: center;">
+        <a href="https://github.com/rust-lang-nursery/rustc-perf">Contribute on GitHub</a>
+    </div>
+    <div class="help-content">
+        <h3><b><code>@rust-timer</code> commands</b></h3>
+        <p><code>@rust-timer</code> supports several commands, the most common (and simple) being <code>@rust-timer queue</code>. This command is usually invoked as <code>@bors try @rust-timer queue</code>, which starts a bors "try" run (not a merge). <code>@rust-timer</code> will wait for the try run to finish, and if it succeeds will then queue a perf run.</p>
+        <p><code>@rust-timer queue</code> has a few extra options that can be useful:</p>
+        <ul>
+            <li><code>include=&lt;INCLUDE&gt;</code> only runs benchmarks with <code>&lt;INCLUDE&gt;</code> in their title. For example, if you just want to run rustdoc benchmarks, do <code>@bors try @rust-timer queue include=-doc</code>; the <code>-doc</code> matches the <code>-doc</code> appended to rustdoc benchmarks. <code>&lt;INCLUDE&gt;</code> is a comma-separated list of substrings; only one of the items in the list of substrings must match.</li>
+            <li><code>exclude=&lt;EXCLUDE&gt;</code> is similar to <code>include=</code>, but instead skips any benchmarks that have one of the items in the list of substrings as a substring of its name.</li>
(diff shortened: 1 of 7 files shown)


camelid:
The page looks like this on my local machine:

image

site/static/help.html
```diff
@@ -0,0 +1,42 @@
...
+        <p><code>@rust-timer queue</code> has a few extra options that can be useful:</p>
+        <ul>
+            <li><code>include=&lt;INCLUDE&gt;</code> only runs benchmarks with <code>&lt;INCLUDE&gt;</code> in their title. For example, if you just want to run rustdoc benchmarks, do <code>@bors try @rust-timer queue include=-doc</code>; the <code>-doc</code> matches the <code>-doc</code> appended to rustdoc benchmarks. <code>&lt;INCLUDE&gt;</code> is a comma-separated list of substrings; only one of the items in the list of substrings must match.</li>
+            <li><code>exclude=&lt;EXCLUDE&gt;</code> is similar to <code>include=</code>, but instead skips any benchmarks that have one of the items in the list of substrings as a substring of its name.</li>
+            <li><code>runs=&lt;RUNS&gt;</code> configures how many times the benchmark is run. <code>&lt;RUNS&gt;</code> is an integer.</li>
+        </ul>
+        <p><code>@rust-timer</code> has more commands than just <code>@rust-timer queue</code>, but the <code>queue</code> command is the most used.</p>
```
jyn514, line 39: What other commands does it support?

site/static/help.html
```diff
@@ -0,0 +1,42 @@
...
+        <p><code>@rust-timer queue</code> has a few extra options that can be useful:</p>
+        <ul>
+            <li><code>include=&lt;INCLUDE&gt;</code> only runs benchmarks with <code>&lt;INCLUDE&gt;</code> in their title. For example, if you just want to run rustdoc benchmarks, do <code>@bors try @rust-timer queue include=-doc</code>; the <code>-doc</code> matches the <code>-doc</code> appended to rustdoc benchmarks. <code>&lt;INCLUDE&gt;</code> is a comma-separated list of substrings; only one of the items in the list of substrings must match.</li>
+            <li><code>exclude=&lt;EXCLUDE&gt;</code> is similar to <code>include=</code>, but instead skips any benchmarks that have one of the items in the list of substrings as a substring of its name.</li>
+            <li><code>runs=&lt;RUNS&gt;</code> configures how many times the benchmark is run. <code>&lt;RUNS&gt;</code> is an integer.</li>
```
jyn514, line 37: Do you know the default value?

camelid: I would assume `1`?

site/static/help.html
camelid: There's also `build`, `make-pr-for`, and `update-branch-for`, but I can't document them because I don't anything about them (except for a little bit about `make-pr-for`).

site/static/help.html
```diff
@@ -0,0 +1,42 @@
...
+<title>rustc performance data</title>
+<link rel="stylesheet" type="text/css" href="perf.css">
+<link rel="alternate icon" type="image/png" href="/favicon-32x32.png">
+<link rel="icon" type="image/svg+xml" href="/favicon.svg">
+<style>
+    .help-content {
+        font-family: Helvetica, Arial, sans-serif;
+        line-height: 140%;
+        font-size: 16px;
+        max-width: 50em;
```
camelid, line 14: This is to prevent the lines from getting too long, which can make the text very hard to read. In case you're wondering, I checked by resizing my window and it should look fine on mobile as well.

camelid:
CI failures look spurious.

site/static/help.html
Mark-Simulacrum: This is like cap-lints -- some benchmarks run once, some run 3 times by default (we generally take the minimum of the metrics), and by default there's not an override, so we use the configuration from the benchmark.

site/static/help.html
Mark-Simulacrum: I think documenting build is worthwhile, the other two are buggy and we shouldn't document them for now.

Mark-Simulacrum (commented):
This looks great to me, left a few comments which I think could be integrated, and then we can merge.

site/static/help.html
camelid: Could we save documenting `build` for a later PR? I know nothing about it so I can't document it.

site/static/help.html
camelid: So you're saying that some benchmarks run multiple times, e.g. with different `cap-lints` settings? And this option will let you override that?

Mark-Simulacrum: No, not with different cap-lints settings. I'm saying that this option *forces* a particular number of runs; without it set, the number of runs varies from benchmark to benchmark.

camelid: I think we're saying the same thing, but I must not have been clear enough. As I understand it, you're saying that by default different benchmarks have different numbers of runs (some might be 1, some 2, some 3, etc.), and if you use `runs=` you can make all the benchmarks have the same number of runs (e.g., `2`).

Are there any benchmarks that have their default run count set to `0`?

Mark-Simulacrum: No, all benchmarks run at least once by default. I'm not sure if runs=0 would work, but it would also be pretty useless I'd expect.

camelid: I added more info in 677a3091, does that look good to you?

site/static/help.html
Mark-Simulacrum: Ok! It's pretty much equivalent to queue, just takes an additional SHA argument (useful if e.g. bors flakes and forgets to post or so)

PR #847, opened 2021-03-04
fix timings order across browsers

klensy:
Currently order if timing differs in different browsers. For example, check this page in chrome and firefox https://perf.rust-lang.org/compare.html?start=15598a83db88ec7a32ea18a44dd6309f32edc07e&end=f288a8e816163b97425adb1baca86b573395e3ec , you can see that in chrome, bootstrap timings goes first, but in firefox - last.

Fixes by setting `is_bootstrap=false` for non bootstrap, and then sorting (as `is_bootstrap` became `true\false` always, instead being none sometimes).

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -246,7 +246,7 @@ input[type="checkbox"] {
             ...bootstrap_names
                 .map(name => name)
                 .map(name => ({name, is_bootstrap: true, fields: to_fields_bootstrap(name)})),
-            ...test_names.map(name => ({name, fields: to_fields(name)})),
+            ...test_names.map(name => ({name, is_bootstrap: false, fields: to_fields(name)})),
         ];
 
         function to_fields_bootstrap(name) {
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Interesting! I was not aware of this difference. Thanks.

Mark-Simulacrum:
Ah, bootstrap timings going last was actually the intent, so I think this swapped the ordering I was expecting -- I'll fix that up.

klensy:
Hmm, this actually broke ordering on not bootstrap https://perf.rust-lang.org/compare.html?start=2021-02-01&end=2021-03-01&stat=instructions%3Au

Mark-Simulacrum:
I am not obviously seeing what you mean, but happy to see a PR if there's some additional breakage - I pushed up the ordering switch.

klensy:
For boostrap - sorted by right column, but for not bootstrap - should be sorted by abs(avg) ?

Mark-Simulacrum:
No, we should be sorting by abs(max change) I think, something like that? i.e., the farthest_pct field is right I believe

PR #848, opened 2021-03-04
fix timings order across browsers: 2

klensy:
Same as https://github.com/rust-lang/rustc-perf/pull/847, but this time order seems work.

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -343,7 +343,7 @@ input[type="checkbox"] {
                     return 1;
                 }
                 else{
-                    return 0;
+                    return -1;
                 }
             }
         });


klensy:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

`
If compareFunction(a, b) returns less than 0, leave a and b unchanged.
`
`
If compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements. Note: the ECMAScript standard only started guaranteeing this behavior in 2019, thus, older browsers may not respect this.
`

Issue #837, opened 2021-02-04
bench_local IDs must be refs in rust-lang/rustc


adamgemmell:
To reproduce:
`./target/release/collector bench_local $(rustup +nightly which rustc) testrun`

Expected:
The run ID "testrun" should label the current run in the database

Actual:
It's also used as a ref when checking out the rustc repo and the whole collector run stops on this benchmark:
`15 benchmarks remaining
Cloning into 'rust'...
remote: Enumerating objects: 1406657, done.
remote: Total 1406657 (delta 0), reused 0 (delta 0), pack-reused 1406657
Receiving objects: 100% (1406657/1406657), 580.28 MiB | 2.47 MiB/s, done.
Resolving deltas: 100% (1133476/1133476), done.
fatal: ambiguous argument 'testrun': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
thread 'main' panicked at 'git reset --hard successful', collector/src/execute/rustc.rs:50:5`

Probably caused by these commits: https://github.com/rust-lang/rustc-perf/commit/965ad78b70514a29098ad57fb33601873716ca67
https://github.com/rust-lang/rustc-perf/commit/9eef993b9487784c79792573d33c40bcdf98528b (when the repo already exists locally)

I'd suggest either having bench_local runs default to checking out origin/master, or adding a new flag to collector to represent a ref in the rust-lang repo (bit more involved)

If someone can verify this is an issue then I can submit a fix for it.

Mark-Simulacrum:
Yeah, this should be fixed. Probably the best thing is to *attempt* to checkout the ref, and then if failing, just skip that benchmark - does that sound reasonable to you?

adamgemmell:
From the perspective of a local user it seems like it'd be better to attempt to checkout origin/master than skipping, though I can see this might mask unintended behaviour when run for perf.rust-lang.org

I'm looking at the collector/README section on comparing two different versions locally when thinking about this
```
./target/release/collector bench_local $RUST_ORIGINAL Original
./target/release/collector bench_local $RUST_MODIFIED Modified
```
as I'd argue this particular benchmark is quite valuable

Mark-Simulacrum:
So, origin/master may not compile is I guess my main worry. But I'd be ok with trying to check out and falling back to origin/master.

adamgemmell:
[...]

PR #861, opened 2021-04-01
Use default branch for rustc if the chosen ref doesn't exist

adamgemmell:
A small fix to default to `rust/origin/HEAD` if checking out a specific ref fails. Previously this would halt the whole collector run. This allows easier comparing of bench_local runs as you can freely choose the run ID.

If the chosen ref (whatever it is) fails to compile then the benchmark is recorded as an error and the run continues. No changes needed here in my opinion.

See https://github.com/rust-lang/rustc-perf/issues/837

### DIFF
--- a/collector/src/execute/rustc.rs
+++ b/collector/src/execute/rustc.rs
@@ -37,7 +37,7 @@ fn record(
     aid: database::ArtifactIdNumber,
 ) -> anyhow::Result<()> {
     let checkout = Path::new("rust");
-    let status = Command::new("git")
+    let mut status = Command::new("git")
         .current_dir("rust")
         .arg("reset")
         .arg("--hard")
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Thanks!

I am a bit worried about the case where somewhere along the line we introduce some kind of typo in the code calling this, and it ends up *always* falling back to the default upstream commit, rather than only on non-commit artifacts. We should be able to check that the ArtifactId passed in here is a commit, and if so, bail out if we fail to checkout. On ArtifactId::Artifact (which is an arbitrary string) we can implement the behavior as you have.

adamgemmell:
That change makes perfect sense - I'd assumed that was figured out by querying the Git repo anyway, but I now see it's contextual based on the CLI usage.

I rebased this branch too, but kept the commits separate for this review. Let me know if you want the branch squashing.

Mark-Simulacrum:
This looks great. I think a separate commit is fine.

Thanks!

PR #865, opened 2021-04-23
Add support for building rustc-perf and running the collector on Windows

### LINKED ISSUE #834: Run perf on Windows
It would be useful to also have the ability to run the perf suite on Windows. One open question would be whether this would run as often as the Linux tests for or if we would kick off Windows tests at a different cadence or on-demand. 

This might also help with #181

wesleywiser:
This PR implements the first two parts of #834: basic support for building the code on Windows and support for profiling using `-Zself-profile`. 

Currently, running all tests works, other than a number of benchmarks that fail to build on my machine due to missing native dependencies (such as `webrender`). 

The general strategy I took was to rewrite in Rust whenever convenient, otherwise falling back to a Windows equivalent of the unix-y tool. 

Part of #834


Mark-Simulacrum:
It looks like CI is failing non-spuriously here - it'll currently fail even on master due to syn incremental being broken in rustc itself, though, so you won't get it fully green :)

Before we land this we should likely add a windows builder to CI to make sure we don't accidentally break support; at least just basic building (e.g. cargo check), but maybe something more involved would work too. I don't think we can use `perf stat` on Windows, so perhaps a self-profile collection on a few benchmarks would also help iron out any latent bugs.

Otherwise I think this approach looks really good to me - some duplication and expansion for sure, but nothing too surprising or unacceptable I think :) Thanks!

collector/src/execute.rs
```diff
@@ -44,26 +58,48 @@ fn rename<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> anyhow::Result<()>
...
-    // This might be a bit slower but at least things build
-    let mut cmd = Command::new("bash");
-    cmd.current_dir(path)
-        .args(&["-c", "find . -name 'CMakeCache.txt' -delete"]);
-    command_output(&mut cmd).with_context(|| format!("deleting cmake caches in {:?}", path))?;
+    fn is_valid(path: &Path) -> bool {
+        let target_dir = Component::Normal(OsStr::new("target"));
+
+        // Don't touch files in `target/`, since they're likely generated by build scripts and might be from a dependency.
+        if path.components().any(|component| component == target_dir) {
```
rylev, line 74: This would break if any subfolder is named `target` no?

Mark-Simulacrum: I think that's only true if the subfolder actually needs touching; I'm not sure that will happen in practice. Realistically we can likely get away with touching the crate "roots" only, but those are a bit annoying to identify.

Mark-Simulacrum:
Hey @wesleywiser -- I noticed this is still marked as [WIP] but wondering if that's no longer actually the case?

Mark-Simulacrum:
```

2021-05-18T20:26:27.1254410Z [2021-05-18T20:26:27Z DEBUG collector::execute] applying patch add static arr item
2021-05-18T20:26:27.1255811Z [2021-05-18T20:26:27Z DEBUG collector::execute] applying add static arr item to "/tmp/.tmpV933YN"
2021-05-18T20:26:27.1330858Z error: patch failed: src/main.rs:65572
2021-05-18T20:26:27.1331599Z error: src/main.rs: patch does not apply
2021-05-18T20:26:27.1344711Z collector error: Failed to benchmark 'coercions', recorded: expected success, got exit status: 1
```

wesleywiser:
I believe this is ready for a final review. I've updated the title and description of the PR to match the current state and I've tested running a basic `profile_local` command on both Windows & Linux with success.

rylev (approved):
🎉

PR #866, opened 2021-04-24
add feature gate for const_fn splitting

RalfJung:
Cc https://github.com/rust-lang/rust/pull/84310
I have no idea if this will even build without that PR landing first, though... but that PR cannot land until the perf test suite is fixed...


Mark-Simulacrum:
Shouldn't be a problem if it breaks here, really.

PR #867, opened 2021-04-24
remove unused const_fn feature

RalfJung:
I verified that `packed-simd` still builds with a recent nightly.

I think the `const_fn` feature in `style-servo` is probably also unused (and should be removed since we plan to remove that feature flag), but I can't build-test that: it fails to build even on rustc-perf master.

### DIFF
--- a/collector/benchmarks/ctfe-stress-4/src/lib.rs
+++ b/collector/benchmarks/ctfe-stress-4/src/lib.rs
@@ -1,5 +1,5 @@
 #![allow(dead_code)]
-#![feature(const_fn, const_fn_trait_bound, const_fn_unsize, const_eval_limit)]
+#![feature(const_fn_trait_bound, const_fn_unsize, const_eval_limit)]
 #![const_eval_limit = "10000000"]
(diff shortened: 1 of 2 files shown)


RalfJung:
Turns out the `const_fn` in style-servo is dead code (that part of style-servo is not even built by the benchmarks), so probably no need to remove that.

RalfJung:
Looking at [the CI logs](https://github.com/rust-lang/rustc-perf/pull/867/checks?check_run_id=2428227074), I have no idea how to tell what actually went wrong...

RalfJung:
@Mark-Simulacrum I updated the PR to remove the `const_fn` feature also in the ctfe stress test. This feature does not do anything any more; removing it here is needed to land https://github.com/rust-lang/rust/pull/85109.

Mark-Simulacrum:
You'll also need to update this commit hash I think - https://github.com/rust-lang/rust/blob/ca82264ec7556a6011b9d3f1b2fd4c7cd0bc8ae2/src/ci/pgo.sh#L15

But merging this for now.

PR #874, opened 2021-05-25
Add triage 2021-05-25

rylev:


Mark-Simulacrum:
Note to self - needs check-in on action items from last week and TWIR.

lqd:
does/should the script also look at max-rss regressions btw ?

Mark-Simulacrum:
It does not; I try to check via the graph most weeks. max-rss is sufficiently noisy that the graph is the only currently good way of detecting max-rss changes, IMO.

rylev:
@Mark-Simulacrum one good thing we could do is have a checklist that all those who do triage should check. I'd be nice to know that we're always checking the same things.

Mark-Simulacrum:
https://github.com/rust-lang/rustc-perf/blob/master/triage/README.md is the place to update, I agree it'd be good.

PR #877, opened 2021-05-27
Remove incorrect and redundant options from git apply command line

tmiasko:
In d0712f6094b6f12de7674ed5559edf7b3c0aac5b `patch -Np1 -i` was replaced
with `git apply -Np1`, but `-N` flag has a different meaning in git:

> mark new files with `git add --intent-to-add`

Additionally the flag was added in v2.19.0, while git version used on
perf.rust-lang.org is apparently older and execution fails with error:

> unknown switch `N'

The default behaviour of git apply should be sufficient to match the
original one. Remove all flags from the command line.

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -1550,7 +1550,7 @@ impl Patch {
         log::debug!("applying {} to {:?}", self.name, dir);
 
         let mut cmd = Command::new("git");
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):

wesleywiser:
Thanks for fixing this @tmiasko! Looks like I'm running `git 2.30.2` locally so I didn't catch that.

PR #878, opened 2021-06-08
Rust triage generation api

rylev:
This adds an API for generating the triage report.


site/src/server.rs
```diff
@@ -768,6 +879,266 @@ impl DateData {
...
+    )
+}
+
+async fn gh_pr_title(pr: u32) -> String {
+    let url = format!("https://api.github.com/repos/rust-lang/rust/pulls/{}", pr);
+    let client = reqwest::Client::new();
+    let mut request = client
+        .get(&url)
+        .header("Content-Type", "application/json")
+        .header("User-Agent", "rustc-pef");
```
bjorn3, line 987:
```suggestion
        .header("User-Agent", "rustc-perf");
```

Mark-Simulacrum:
We don't currently worry too much about rate limiting across any of the Rust-run services other than crates.io, though it's definitely debt that we're putting in place. I wouldn't worry about it for the purposes of this PR.

site/src/comparison.rs
```diff
@@ -0,0 +1,583 @@
...
+    let lo = lo.map(|lo| benchmarks.remove(lo));
+    let hi = benchmarks
+        .iter()
+        .enumerate()
+        .max_by(|&(_, b1), &(_, b2)| cmp(b1, b2))
+        .filter(|(_, c)| c.is_significant() && c.is_increase())
+        .map(|(i, _)| i);
+    let hi = hi.map(|hi| benchmarks.remove(hi));
+
+    benchmarks.clear();
```
Mark-Simulacrum, line 134: Why are we clearing out benchmarks here? It's going to get dropped immediately afterwards anyway and is just a vector as far as I can tell?

Mark-Simulacrum (commented):
Few small comments, but overall looks good

site/src/comparison.rs
rylev: Ah this is a leftover from a previous change. I will remove

site/src/comparison.rs
```diff
@@ -0,0 +1,583 @@
...
+type BoxedError = Box<dyn Error + Send + Sync>;
+
+pub async fn handle_triage(
+    body: api::triage::Request,
+    data: &InputData,
+) -> Result<api::triage::Response, BoxedError> {
+    let start = body.start;
+    let end = body.end;
+    // Compare against self to get next
+    let master_commits = rustc_artifacts::master_commits().await?;
```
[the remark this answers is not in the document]
rylev: Looked into this more. Indeed this would save us having to do the first `compare` to get the next commit, but we still need to fetch the `master_commits` in order to get the PR number. I tried doing a db lookup for each PR number but this didn't seem to work. 

Either way, perhaps this should be a follow up to this PR since there doesn't seem to be a trivial way to improve this.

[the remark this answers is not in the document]
Mark-Simulacrum: Ah, that makes sense. Yeah, we can leave fine tuning to later; we do store the PR number I think but it'll be fine to do that as a follow-up.

Mark-Simulacrum:
Alright, let's merge this in. Thanks!

PR #879, opened 2021-06-09
Update deps (including tokio to 1.x)

rylev:
~*NOTE*: this is ready for review, but not ready to merge. We first need to update [Mark-Simulacrum/rustc-artifacts](https://github.com/Mark-Simulacrum/rustc-artifacts) to use the latest version of `reqwest`. Without that change this change does not work.~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


Mark-Simulacrum:
Thanks for doing this! I think the PR description may no longer be valid as you've inlined the logic from that crate? If so, I am OK merging this -- I'll probably try to make note of when I do so in case it causes any differences in site or collector performance (that wouldn't be expected but we are updating a lot of stuff).

rylev:
@Mark-Simulacrum correct. This is now ready to merge. 😊 Let me know how you want to monitor things, and I can help.

PR #881, opened 2021-06-10
Categorize perf runs when posting to GitHub

rylev:
When the triage bot posts the results of running a performance run, it now provides a summary of the performance run. It should look something like this:

*****
Finished benchmarking try commit (ab345): [comparison url]().
 
**Summary**:  This change led to significant regressions 😿 in compiler performance.
- Moderate regression in [instruction counts](https://perf.rust-lang.org/compare.html?start=5b638c1d3751b7ab31cac9739add516bdf39e10a&end=35fff69d043b1c0f5c29894e7f4b0da8b039c131&stat=instructions:u) (up to 4.3% on `full` builds of `wg-grammar-check`)

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -638,7 +638,7 @@ pub async fn post_finished(data: &InputData) {
                 format!(
                     "Finished benchmarking try commit ({}): [comparison url]({}).
 
-                    Summary: {}
+                    **Summary**: {}
 
 Benchmarking this pull request likely means that it is \
 perf-sensitive, so we're automatically marking it as not fit \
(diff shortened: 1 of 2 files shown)


site/src/github.rs
```diff
@@ -649,10 +652,36 @@ regressions or improvements in the roll up.
...
+        data,
+    )
+    .await
+    {
+        Ok(Some(c)) => c,
+        _ => return String::from("ERROR categorizing benchmark run!"),
+    };
+    let summary = match crate::comparison::ComparisonSummary::summarize_comparison(&comparison) {
+        Some(s) => s,
+        None => return String::from("This benchmark run did not return any significant changes"),
```
Mark-Simulacrum: Can you adjust this and the summaries in general to ask users to file an issue on rustc-perf if they disagree?

Mark-Simulacrum (approved):
Happy to merge with one comment resolved (feel free to do so yourself).

site/src/github.rs
```diff
@@ -631,12 +631,15 @@ pub async fn post_finished(data: &InputData) {
...
                 &data.config,
                 commit.pr,
                 format!(
                     "Finished benchmarking try commit ({}): [comparison url]({}).
 
+**Summary**: {}
+
 Benchmarking this pull request likely means that it is \
 perf-sensitive, so we're automatically marking it as not fit \
 for rolling up. Please note that if the perf results are \
```
LeSeulArtichaut, line 645: Just an idea: if the perf results don't return any significant perf changes, is this `rollup=never` necessary?

bjorn3: A perf run is likely done when a change to a perf sensitive part of the compiler is done. It may not result in any significant perf changes when the perf run is done, but due to interaction with other PR's it may still have a perf change when actually landing.

bjorn3: This categorization is also way too insensitive IMO. A perf run with improvements up to 0.6% and a couple of regressions up to 0.8% was marked as not having any significant changes. Many changes >0.2% across multiple benchmarks are unlikely to be noise. https://github.com/rust-lang/rust/pull/86404#issuecomment-866714363

image

Mark-Simulacrum: I do think we can fine-tune sensitivity, but until we're even more confident I agree that we shouldn't drop the rollup=never. Our queue doesn't suffer much from it today anyway, to my knowledge.

PR #882, opened 2021-06-11
escape query name to prevent rendering it as html

klensy:
This fixes issue with `<unknown>` query name will rendered as tag and became invisible.

For example see https://perf.rust-lang.org/detailed-query.html?sort_idx=1&commit=4bc7477fe1c867c78f756dbad76db4ea55c7eebf&base_commit=c5fbcd35a8217a17f6b63a22217ace06cf8f5f02&benchmark=coercions-debug&run_name=incr-unchanged (first string in table)

### DIFF
--- a/site/static/shared.js
+++ b/site/static/shared.js
@@ -252,3 +252,13 @@ function make_request(path, body) {
         console.log("error fetching ", path, ": ", err);
     });
 }
+
+// https://stackoverflow.com/questions/6234773
+function escapeHtml(unsafe) {
+    return unsafe
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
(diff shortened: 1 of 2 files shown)


site/static/shared.js
```diff
@@ -252,3 +252,14 @@ function make_request(path, body) {
         console.log("error fetching ", path, ": ", err);
     });
 }
+
+// https://stackoverflow.com/questions/6234773
+// handmade escape, as we don't use any libs (why?)
```
rylev, line 257: It's good to ask why we don't use any libs, but I don't think it's necessary to pose this in a comment.

rylev (changes requested):
Looks good to me, just one nit about removing an unnecessary comment.

Mark-Simulacrum (approved):

PR #885, opened 2021-06-18
Support collecting hardware performance counters on Windows

wesleywiser:
Collect the `InstructionRetired` and `TotalCycles` hardware performance counters on Windows using a combination of `xperf` and `tracelog`. After collection, we process the resulting trace file to calculate the totals for the performance counters and store those in the appropriate parts of the database for the run. The site is able to display the results without modification and we also have self-profiling data available to site users. 

Note: the `cpu-clock` `perf` event is not an actual hardware event and no corresponding event is available out of the box on Windows. I believe we can calculate an equivalent value from the trace data but that is left as a future TODO item. 

Part of #834

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -156,6 +156,39 @@ If you've collected new data, you can run `curl -X POST
 localhost:2346/perf/onpush` to update the site's view of the data, or just
 restart the server.
 
+### Benchmarking on Windows
+
+To benchmark on Windows, you will need to run the collector in a elevated context
+so that it can access the hardware performance counters. Note: some virtualized
+environments do not permit access to these counters for guest VMs.
+
+You will also need to provide the paths to the xperf and tracelog tools (or have them
+available on your PATH). Some common paths to these tools look like:
+
+```
+$env:XPERF="C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\xperf.exe"
+$env:TRACELOG="C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\tracelog.exe"
+```
+
+Finally, while most of the options you can pass to the collector are supported, the majority of
+the profilers used in the `profile_local` command are not. In Windows, the only currently supported
+profiler is the `self-profiler`.
+
+As a complete example, let's run just the `regex` benchmark in the `Debug` build with
+self-profiling results available:
+
+```
+$env:XPERF="C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\xperf.exe"
+$env:TRACELOG="C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\tracelog.exe"
+.\target\release\collector.exe bench_local $env:RUST_ORIGINAL Original --builds Debug --include regex --self-profile
+.\target\release\collector.exe bench_local $env:RUST_MODIFIED Modified --builds Debug --include regex --self-profile
+.\target\release\site.exe .\results.db
+```
+
+The open a web browser to `http://localhost:2346/compare.html?start=Original&end=Modified&stat=instructions%3Au`.
+
+Note: This example uses Powershell syntax.
+
 ### Technical details of the benchmark server
 
 We download the artifacts (rustc, rust-std, cargo) produced by CI and properly
--- a/collector/README.md
+++ b/collector/README.md
@@ -165,7 +165,7 @@ environments do not permit access to these counters for guest VMs.
 You will also need to provide the paths to the xperf and tracelog tools (or have them
 available on your PATH). Some common paths to these tools look like:
 
-```
+```pwsh
 $env:XPERF="C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\xperf.exe"
 $env:TRACELOG="C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\tracelog.exe"
 ```
@@ -177,7 +177,7 @@ profiler is the `self-profiler`.
 As a complete example, let's run just the `regex` benchmark in the `Debug` build with
 self-profiling results available:
 
-```
+```pwsh
 $env:XPERF="C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\xperf.exe"
 $env:TRACELOG="C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\tracelog.exe"
 .\target\release\collector.exe bench_local $env:RUST_ORIGINAL Original --builds Debug --include regex --self-profile
--- a/collector/src/etw_parser.rs
+++ b/collector/src/etw_parser.rs
@@ -287,10 +287,11 @@ fn parse_events(r: &mut dyn BufRead, headers: Vec<EventHeader>) -> anyhow::Resul
     Ok(events)
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, PartialEq)]
 pub struct Counters {
     pub instructions_retired: u64,
     pub total_cycles: u64,
+    pub cpu_clock: f64,
 }
 
 impl std::ops::Add for Counters {
@@ -300,6 +301,7 @@ impl std::ops::Add for Counters {
         Self {
             instructions_retired: self.instructions_retired + rhs.instructions_retired,
             total_cycles: self.total_cycles + rhs.total_cycles,
+            cpu_clock: self.cpu_clock + rhs.cpu_clock,
         }
     }
 }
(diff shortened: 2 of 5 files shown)


collector/src/execute.rs
```diff
@@ -561,9 +571,17 @@ impl<'a> MeasureProcessor<'a> {
...
     ) -> Self {
-        // Check we have `perf` available.
-        let has_perf = Command::new("perf").output().is_ok();
-        assert!(has_perf);
+        // Check we have `perf` or (`xperf.exe` and `tracelog.exe`)  available.
+        if cfg!(unix) {
+            let has_perf = Command::new("perf").output().is_ok();
+            assert!(has_perf);
+        } else {
+            let has_xperf = Command::new(env::var("XPERF").unwrap_or("xpref.exe".to_string())).output().is_ok();
```
Mark-Simulacrum, line 579: XPERF vs. xpref (swapping pref and perf) seems a bit odd here -- by mistake?

collector/README.md
```diff
@@ -156,6 +156,38 @@ If you've collected new data, you can run `curl -X POST
...
+You will also need to provide the paths to the xperf and tracelog tools (or have them
+available on your PATH). Some common paths to these tools look like:
+
+```
+$env:XPERF="C:\Program Files (x86)\Windows Kits\10\Windows Performance Toolkit\xperf.exe"
+$env:TRACELOG="C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\tracelog.exe"
+```
+
+Finally, while most of the options you can pass to the collector are supported, the majority of
+the profilers are not. In Windows, the only currently supported profiler is the `self-profiler`.
```
Mark-Simulacrum, line 174: Based on the code/implementation it seems like xperf-stat might be supported too?

Mark-Simulacrum:
Haven't reviewed too closely but at a high level this seems reasonable to me, will review closely later.

collector/src/execute.rs
wesleywiser: Good catch! Fixed.

collector/README.md
wesleywiser: I was more referring to the list of profilers supported by the `perf_local` command which doesn't support `perf-stat` either (but does support a `perf` profiler). I think that probably makes sense to support an `xperf` profiler there but that's not currently implemented.

wesleywiser:
Thanks @Mark-Simulacrum!

collector/README.md
Mark-Simulacrum: Ah, makes sense. Could you clarify this here? I think it's generally a confusing piece of our infrastructure today but seems good to not add to that!

wesleywiser: Sure! Let me know if you think I should add more.

collector/README.md
```diff
@@ -156,6 +156,39 @@ If you've collected new data, you can run `curl -X POST
...
+```
+
+Finally, while most of the options you can pass to the collector are supported, the majority of
+the profilers used in the `profile_local` command are not. In Windows, the only currently supported
+profiler is the `self-profiler`.
+
+As a complete example, let's run just the `regex` benchmark in the `Debug` build with
+self-profiling results available:
+
+```
```
rylev, line 180: I think if you add pwsh here you'll get some syntax highlighting

collector/src/etw_parser.rs
```diff
@@ -0,0 +1,677 @@
...
+    Ok(headers)
+}
+
+fn validate_os_header_line(r: &mut dyn BufRead) -> anyhow::Result<()> {
+    let mut line = String::new();
+    r.read_line(&mut line)?;
+
+    anyhow::ensure!(line.starts_with("OS Version"), "OS version line not found");
+
+    let components: Vec<_> = line.split(',').collect();
```
rylev, line 101: Nit: is allocating a vec here necessary?

collector/src/etw_parser.rs
```diff
@@ -0,0 +1,677 @@
...
+    anyhow::ensure!(events_lost == 0, "events were lost");
+
+    let buffers_lost: u32 = get_value("Buffers lost:").parse()?;
+    anyhow::ensure!(buffers_lost == 0, "buffers were lost");
+
+    Ok(())
+}
+
+#[derive(Debug, Eq, PartialEq)]
+struct Pmc {
```
rylev, line 124: Can we note somewhere that `Pmc` stands for "performance monitoring counter"?

collector/src/etw_parser.rs
```diff
@@ -0,0 +1,677 @@
...
+    timestamp: u64,
+    thread_id: u64,
+    instructions_retired: u64,
+    total_cycles: u64,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+struct CSwitch {
```
rylev, line 132: I assume this is a context switch? It's fine to use the short form if that's what's normally used, I'd just note as a doc.

collector/src/etw_parser.rs
```diff
@@ -0,0 +1,677 @@
...
+        }
+    }
+}
+
+impl Default for Counters {
+    fn default() -> Self {
+        Self {
+            instructions_retired: 0,
+            total_cycles: 0,
+            // FIXME(wesleywiser): We should be properly calculating this value by taking the total time
```
rylev, line 323: Should we create issues in the repo for these FIXMEs?

collector/src/execute.rs
```diff
@@ -182,6 +185,7 @@ impl Profiler {
             // is rejected because it can't be used with the `profiler`
             // subcommand. (It's used with `bench_local` instead.)
             "perf-stat" => Err(anyhow!("'perf-stat' cannot be used as the profiler")),
+            "xperf-stat" => Err(anyhow!("'xperf-stat' cannot be used as the profiler")),
```
rylev, line 188: Should we explain why these profilers can't be used in the error message?

rylev (approved):
♥ Great!

collector/src/etw_parser.rs
wesleywiser: Probably not but I didn't want to resplit the string twice. I'm not sure which is more expensive...

collector/src/execute.rs
wesleywiser: I'm not entirely sure why this isn't allowed. I think technically it should work but the results will just be the statistics captured during the run which is probably not very useful for profiling (vs benchmarking).

collector/src/etw_parser.rs
wesleywiser: That seems reasonable to me! I'll create issues when this merges and link to the FIXME lines.

wesleywiser:
Thanks for the review @rylev!

collector/src/rustc-fake.rs
```diff
@@ -77,62 +78,62 @@ fn main() {
...
+                }
+            }
+
+            "xperf-stat" | "xperf-stat-self-profile" => {
+                // Read the path to tracelog.exe from either an environment variable, falling back to assuming it's on the PATH.
+                let tracelog = std::env::var("TRACELOG").unwrap_or("tracelog.exe".to_string());
+                let mut cmd = Command::new(tracelog);
+                assert!(cmd.output().is_ok(), "tracelog.exe could not be started");
+
+                cmd.args(&["-start", "counters", "-f", "counters.etl", "-eflag", "CSWITCH+PROC_THREAD+LOADER", "-PMC", "InstructionRetired,TotalCycles:CSWITCH"]);
```
michaelwoerister, line 91: Would you mind adding documentation somewhere on how this all works? I.e. what the workflow is, which programs are called in which order, what the flags mean? A link to a good description would be fine too.

Mark-Simulacrum:
Happy to merge this whenever, looks like there's a few unresolved comments left though -- just let me know.

collector/src/rustc-fake.rs
michaelwoerister: From playing around with this it looks like the "counters" parameter a globally visible "LoggerName" used for coordinating between the `tracelog` and the `xperf` invocations, right? I think it might be a good idea to generate a more meaningful name here that includes the fact that it belongs to "rustc-perf" and maybe the name of benchmark being run. Something like `rustc-perf-{benchmark}-logger`.

I don't know if it would also make sense to add some randomly generated ID to it too. At least I had some problems where I had to run `xperf -stop counters` manually because otherwise `tracelog` would complain that `counters` already exists. I think my system got into that state because there was some crash after `tracelog` allocated the logger with name "counters" and the `xperf` command stopping the logger was never executed. Making that more robust one way or other seems like a good idea.

wesleywiser: Added some comments here explaining what the workflow for collecting the counters is. 

I also added code to issue a `xperf -stop` before we do our current run. I think that's better than generating a different name each time as I've read there is a limit to how many concurrent collections you can have running and, if we know the collection name, we can just try stopping it.

wesleywiser:
This is ready to merge! I'll leave it open for now in case @Mark-Simulacrum wants to take another look but if not, I'll merge later today.

PR #886, opened 2021-06-23
Change GitHub comment output depending on perf categorization

rylev:
This changes:
* When we run `@bors rollup=never` to only be when we detect performance changes
* The labels we create after a perf run. We now add `perf-regression` on regressions and mixed results and we remove that label otherwise.

This also makes the codebase only compile on Rust 1.53 due to use of or-patterns.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -1,4 +1,5 @@
 use crate::api::{github, ServerResult};
+use crate::comparison::{ComparisonSummary, Direction};
 use crate::load::{Config, InputData, TryCommit};
 use anyhow::Context as _;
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Let's not have the rollup=never conditionality -- I don't think we're quite ready for that yet. Otherwise r=me

lqd:
> 
>     * The labels we create after a perf run. We now add `perf-regression` on regressions and mixed results and we remove that label otherwise.
> 

It seems the timer bot [doesn't have the credentials](https://github.com/rust-lang/rust/pull/86525#issuecomment-868593482) to set this label yet

Issue #887, opened 2021-06-29
I disagree with rustc-perf's assessment of rust-lang/rust#85499


jyn514:
rust-lang/rust#85499 is a very large improvement on some cases with moderately large regressions on others. If triagebot isn't sure which, it shouldn't add a label. Right now it seems to give precedence to regressions.

jyn514:
In particular, this led @jackh726 to spend quite a lot of time looking into the regression for not much benefit in the end: https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Analyzing.20a.20perf.20regression.

Are there docs somewhere for how to debug regressions? It seems like a lot of manual work and asking around right now.

rylev:
Thanks for reporting this! This issues raises several questions. I'll do my best to list the questions as well as my opinion on what the answers are:

* **Was this issue appropriately labeled?** Regressions accompanied by improvements are still regressions. I think labeling this as a performance regression is the right thing to do. 
* **Should the triage bot be even more conservative with labeling?** Performance regressions require _attention_ (note: I'm not saying they require action) and labeling is the best way to ensure that attention is paid to regressions. Again, I think the labeling here was appropriate.
* **Was the right procedure for @jackh726 to spend time investigating the regression?** Probably not. This is where I see the real breakdown. The intention of the label was not to change any procedure (by encouraging more investigation into performance issues than was already taking place) but to ensure that we can track performance issues so that they're not simply forgotten about. This was a mistake on my part for not better communicating the purpose of the new label. 
* **Are there docs somewhere for how to debug regressions?** Very few (and they are mostly [related to the triage process](https://github.com/rust-lang/rustc-perf/tree/master/triage)). This is something we need to address, but I see it as a separate issue from the labeling.

jyn514:
> This was a mistake on my part for not better communicating the purpose of the new label.

I saw on Zulip you're going to make an announcement in the compiler meeting - that's great, but I think it would be more helpful to put it directly in the comment that rustc-perf posts. Not everyone who works on performance is on the compiler team and they should be made aware too

rylev:
@jyn514 Do you have any suggestions on what the comment should say? Again, I think the label was correct - it _was_ a mixed result. We need to be clear that sometimes mixed results or regressions are ok.

We could for instance only show the following when the the run is categorized as regression or mixed:

[...]

PR #890, opened 2021-07-01
More clearly indicate procedure after perf run completes

rylev:
Attempts to address #887

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -637,6 +637,32 @@ pub async fn post_finished(data: &InputData) {
                 Some(Direction::Regression | Direction::Mixed) => "+perf-regression",
                 Some(Direction::Improvement) | None => "-perf-regression",
             };
+            let msg = direction
+                .map(|d| {
+                    format!(
+                        "While you can manually mark this PR as fit \
+            for rollup, we strongly recommend not doing so since this PR led to changes in \
+            compiler perf.{}",
+                        match d {
+                            Direction::Regression | Direction::Mixed =>
+                                "\n\n**Next Steps**: If you can justify the \
+                regressions found in this perf run, please indicate this with \
+                `@rustbot label: +perf-regression-justified` along with \
+                sufficient written justification. If you cannot justify the regressions \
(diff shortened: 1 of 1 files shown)


site/src/github.rs
```diff
@@ -637,6 +637,32 @@ pub async fn post_finished(data: &InputData) {
...
+                neutral or positive results, the label will be automatically removed.",
+                            _ => "",
+                        }
+                    )
+                })
+                .unwrap_or(
+                    "Please note that since \
+            performance does not seem to be significantly changed, you may *consider* undoing the \
+            rollup directive by specifying `@bors rollup-`. Only do this if you strongly believe \
+            the PR will continue to not lead to performance changes after merging."
```
Mark-Simulacrum, line 662: I think we should avoid giving this advice just yet. In practice, we've definitely seen some perf runs with a certain amount of noise (0.2-0.1% mostly on a scattering of benchmarks) that we categorize as "unchanged" here. I would rather not hint at removing rollup=never; lately the bors queue is not particularly long and I think we should lean on the side of caution.

site/src/github.rs
```diff
@@ -637,6 +637,32 @@ pub async fn post_finished(data: &InputData) {
...
+            compiler perf.{}",
+                        match d {
+                            Direction::Regression | Direction::Mixed =>
+                                "\n\n**Next Steps**: If you can justify the \
+                regressions found in this perf run, please indicate this with \
+                `@rustbot label: +perf-regression-justified` along with \
+                sufficient written justification. If you cannot justify the regressions \
+                please fix the regressions and do another perf run. If the next run shows \
+                neutral or positive results, the label will be automatically removed.",
+                            _ => "",
```
Mark-Simulacrum, line 654: Minor point: I prefer exhaustive matching in case we change things up (though probably not *really* likely we add more directions...).

Mark-Simulacrum (commented):
Looks great overall, left a few minor comments. Also filed https://github.com/rust-lang/rustc-perf/issues/891 for the future.

PR #895, opened 2021-07-07
Add some basic docs

rylev:
Starting to add some docs to make onboarding into the codebase a bit easier.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -15,24 +15,33 @@ pub use self_profile::{QueryData, SelfProfile};
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Deserialize)]
 pub struct DeltaTime(#[serde(with = "round_float")] pub f64);
 
+/// The bound of a range changes in codebase
+///
+/// This can either be the upper or lower bound
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Bound {
-    // sha, unverified
+    /// An unverified git commit (in sha form)
     Commit(String),
+    /// A date in time
     Date(NaiveDate),
+    /// No bound
     None,
 }
 
 impl Bound {
+    /// Tests whether self bounds the commit to the left
     pub fn left_match(&self, commit: &Commit) -> bool {
-        let last_month = chrono::Utc::now().date().naive_utc() - chrono::Duration::days(30);
         match self {
             Bound::Commit(sha) => commit.sha == **sha,
             Bound::Date(date) => commit.date.0.naive_utc().date() >= *date,
-            Bound::None => last_month <= commit.date.0.naive_utc().date(),
+            Bound::None => {
+                let last_month = chrono::Utc::now().date().naive_utc() - chrono::Duration::days(30);
+                last_month <= commit.date.0.naive_utc().date()
+            }
         }
     }
 
+    /// Tests whether self bounds the commit to the right
     pub fn right_match(&self, commit: &Commit) -> bool {
         match self {
             Bound::Commit(sha) => commit.sha == **sha,
@@ -148,7 +157,7 @@ pub fn run_command(cmd: &mut Command) -> anyhow::Result<()> {
 pub fn robocopy(
     from: &std::path::Path,
     to: &std::path::Path,
-    extra_args: &[&dyn AsRef<std::ffi::OsStr>]
+    extra_args: &[&dyn AsRef<std::ffi::OsStr>],
 ) -> anyhow::Result<()> {
     let mut cmd = Command::new("robocopy");
     cmd.arg(from).arg(to).arg("/s").arg("/e");
@@ -219,7 +228,7 @@ pub fn command_output(cmd: &mut Command) -> anyhow::Result<process::Output> {
             output.status,
             String::from_utf8_lossy(&output.stderr),
             String::from_utf8_lossy(&output.stdout)
-        )); 
+        ));
     }
 
     Ok(output)
@@ -246,7 +255,8 @@ pub struct MasterCommit {
 /// Note that this does not contain try commits today, so it should not be used
 /// to validate hashes or expand them generally speaking. This may also change
 /// in the future.
-pub async fn master_commits() -> Result<Vec<MasterCommit>, Box<dyn std::error::Error + Sync + Send>> {
+pub async fn master_commits() -> Result<Vec<MasterCommit>, Box<dyn std::error::Error + Sync + Send>>
+{
     let response = reqwest::get("https://triage.rust-lang.org/bors-commit-list").await?;
     Ok(response.json().await?)
 }
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -378,9 +378,12 @@ pub enum Label {
     Query(QueryLabel),
 }
 
+/// An identifier for a built version of the compiler
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum ArtifactId {
+    /// A built version of the compiler at an exact commit
     Commit(Commit),
+    /// A symbolic tag for a built compiler like "1.51.0"
     Artifact(String),
 }
 
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -189,13 +189,18 @@ impl Ord for Commit {
     }
 }
 
+/// The compilation profile (i.e., how the crate was built)
 #[derive(
     Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
 )]
 pub enum Profile {
+    /// A checked build (i.e., no codegen)
     Check,
+    /// A debug build (i.e., low optimizations)
     Debug,
+    /// A doc build
     Doc,
+    /// An optimized "release" build
     Opt,
 }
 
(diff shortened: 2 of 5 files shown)


site/src/load.rs
```diff
@@ -64,23 +64,32 @@ impl TryCommit {
...
 }
 
+/// Site configuration
 #[derive(Debug, Deserialize)]
 pub struct Config {
     pub keys: Keys,
 }
 
+/// Site context object that contains global data
 pub struct InputData {
```
rylev, line 85: We might want to consider renaming this. InputData is descriptive enough to make me think it's something specific but not descriptive enough to know what that might be. 

Considering it's sort of a collection of cached items we just want to keep around, something purposefully vague like `SiteCtx` might be better.

site/src/load.rs
```diff
@@ -64,23 +64,32 @@ impl TryCommit {
     }
 }
 
+/// Keys for accessing various services
+///
+/// At the moment only used for accessing GitHub
 #[derive(Debug, Default, Deserialize)]
 pub struct Keys {
+    /// GitHub API token from the `GITHUB_API_TOKEN` env variable
     pub github: Option<String>,
```
rylev, line 73: These fields might have better names like `github_api_token` and `github_webhook_secret`.

database/src/lib.rs
```diff
@@ -434,11 +434,11 @@ pub struct LabelId(pub u8, pub u32);
 #[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub struct ArtifactIdNumber(pub u32);
 
+/// Cached results of various queries.
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct Index {
     commits: Indexed<Commit>,
```
rylev, line 440: I'd like to add comments to these fields, but I'm still not exactly sure what they are. For instance, commits seems to be from the `artifact` table with `type` master or try. I'm not exactly sure what that should be.

collector/src/lib.rs
```diff
@@ -15,24 +15,33 @@ pub use self_profile::{QueryData, SelfProfile};
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Deserialize)]
 pub struct DeltaTime(#[serde(with = "round_float")] pub f64);
 
+/// The bound of a range changes in codebase
```
rylev, line 18: `Bound` doesn't seem to always be a bound in the true sense. If it's a `Commit` we're usually looking for exact matches. I'm not sure what `Bound` is really supposed to be.

collector/src/lib.rs
```diff
@@ -15,24 +15,33 @@ pub use self_profile::{QueryData, SelfProfile};
...
+    /// No bound
     None,
 }
 
 impl Bound {
+    /// Tests whether self bounds the commit to the left
     pub fn left_match(&self, commit: &Commit) -> bool {
-        let last_month = chrono::Utc::now().date().naive_utc() - chrono::Duration::days(30);
         match self {
             Bound::Commit(sha) => commit.sha == **sha,
```
rylev, line 35: Here, for example, we don't treat the `Bound` as a bound but rather an exact match.

collector/src/lib.rs
```diff
@@ -246,7 +255,8 @@ pub struct MasterCommit {
 /// Note that this does not contain try commits today, so it should not be used
 /// to validate hashes or expand them generally speaking. This may also change
 /// in the future.
-pub async fn master_commits() -> Result<Vec<MasterCommit>, Box<dyn std::error::Error + Sync + Send>> {
+pub async fn master_commits() -> Result<Vec<MasterCommit>, Box<dyn std::error::Error + Sync + Send>>
+{
```
rylev, line 259: My editor runs cargo fmt on save, and it looks like this file wasn't fully formatted. Shall we add checking for formatting to the CI?

site/src/selector.rs
```diff
@@ -33,22 +33,26 @@ use std::fmt;
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
-pub fn data_for(data: &Index, is_left: bool, query: Bound) -> Option<ArtifactId> {
+/// Gets an artifact id for the given bound
+///
+/// Searches in the indexed commits either from the right or left
+/// If not found in the indexed commits, searches through the artifacts cache
+pub fn data_for(data: &Index, is_left: bool, bound: Bound) -> Option<ArtifactId> {
```
rylev, line 40: This function was hard for me to grok at first. I mean it's simple to understand, but every time I came across it, I'd have to reread the entire body because the name didn't help me at all to remember what it does.

site/src/comparison.rs
```diff
@@ -257,18 +259,30 @@ pub async fn compare_given_commits(
     }))
 }
 
-/// Data associated with a specific date
+/// Data associated with a specific artifact
 #[derive(Debug, Clone, Serialize)]
 pub struct DateData {
```
rylev, line 264: This should probably be renamed `ArtifactData` as this is for a given artifact which may or may not have a date associated with it.

site/src/comparison.rs
```diff
@@ -257,18 +259,30 @@ pub async fn compare_given_commits(
...
+    /// The artifact in question
+    pub commit: String,
```
rylev, line 266: This should probably be `artifact: ArtifactId` since `commit` is wrong (it might be an artifact "tag" and not a commit sha) and there's no need for it to be a string at this point.

site/src/comparison.rs
```diff
@@ -257,18 +259,30 @@ pub async fn compare_given_commits(
...
+    /// Bootstrap data in the form "$crate" -> nanoseconds
     pub bootstrap: HashMap<String, u64>,
 }
 
 impl DateData {
+    /// For the given `ArtifactId`, consume the first datapoint in each of the given `SeriesResponse`
+    ///
+    /// It is assumed that the provided ArtifactId is the same as artifact id returned as the next data
+    /// point from all of the series `SeriesResponse`s. If this is not true, this function will panic.
     async fn consume_one<'a, T>(
```
rylev, line 286: I finally understand why this is called `consume_one` but I don't really like the name....

site/src/load.rs
Mark-Simulacrum: Happy to see this renamed -- I didn't ever come up with a nice name for it. I think SiteCtx is not a bad idea.

collector/src/lib.rs
Mark-Simulacrum: It is supposed to be "exact" in the sense of identifying a particular commit in our storage, at least at the time of access.

Commits hashes are easy, obviously, though the string in Commit(String) can also be the artifact ID (e.g., 1.53.0 or whatever the user provides on the command line when recording data locally).

collector/src/lib.rs
Mark-Simulacrum: Yeah, I think the thing that might be missing in the intuition is that for the date bounds, we don't want an exact match because there may not be commits on that day, but you still want to have the query for e.g. the last month or year of data to work, not error out saying "no commits on 2021-01-01" or something like that.

With commits there's not really such a problem; it doesn't make sense to have a partial match there as much.

database/src/lib.rs
```diff
@@ -227,15 +232,23 @@ impl fmt::Display for Profile {
     }
 }
 
+/// The incremental cache state
+///
+/// These are usually reported to users in a "flipped" way. For example,
+/// `Cache::Empty` means we're doing a "full" build. We present this to users as "full".
```
Mark-Simulacrum, line 238: FWIW it would be fine by me to rename the enum variants here -- they were named when we were trying different things but it's no big deal to rename if needed.

database/src/lib.rs
Mark-Simulacrum: Can you say more about your confusion? These fields are *intended* to be essentially duplicates of the various tables (artifacts, error_series, pstat_series, etc.) so that we can avoid a network round-trip in the case of Postgres when fetching the ID used in the "main" table.

To some extent I think the presence of these is mostly an artifact of not wanting to write too much SQL -- it's much easier to have ad-hoc searching in Rust code than threading through SQL queries. In practice IIRC it was also a little faster, but whether that's still true or if I'm remembering right I'm less sure.

site/src/comparison.rs
Mark-Simulacrum: Seems reasonable. It was named prior to the notion of artifacts -- and "CommitData" was already used for something else I think at that time...

site/src/comparison.rs
```diff
@@ -257,18 +260,30 @@ pub async fn compare_given_commits(
...
+    /// The date of the artifact if known
     pub date: Option<Date>,
+    /// The pr of the artifact if known
     pub pr: Option<u32>,
-    pub commit: String,
+    /// Benchmark data in the form "$crate-$profile" -> Vec<("$cache", nanoseconds)>
+    ///
+    /// * $profile refers to the flavor of compilation like debug, doc, opt(timized), etc.
+    /// * $cache refers to how much of the compilation must be done and how much is cached
+    /// (e.g., "incr-unchanged" == compiling with full incremental cache and no code having changed)
```
Mark-Simulacrum, line 276: It's plausible we could have non-String data here too. IIRC it's like this because making sure all the JS code is up to date when changing the Rust serialization was a bit of a pain - and Serde can be a bit picky in terms of serializing enums and struct nicely when they're nested inside vecs/hashmaps etc

site/src/selector.rs
Mark-Simulacrum: As always, happy to make any changes -- I've never had good experiences naming things in rustc-perf.

Mark-Simulacrum:
Left comments on most of your comments - largely I think everything looks good in terms of existing docs, don't think I had any concerns.

database/src/lib.rs
rylev: My comment is a bit old as I know understand more about the structure. Perhaps we should do a short write up on the database structure. We could then refer to that here.

rylev:
I'm going to merge this since there's no real concerns. I'll submit a small PR renaming things. I definitely feel much better about the code now. As far as documentation goes, I think the only other major thing I'd like to do is to document the database structure. Other changes can just be made along the way.

PR #899, opened 2021-07-09
Start documenting schema

rylev:
Start documenting the database schema

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -2,6 +2,48 @@
 
 Below is an explanation of the current database schema. This schema is duplicated across the (currently) two database backends we support: sqlite and postgres.
 
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


database/schema.md
```diff
@@ -0,0 +1,113 @@
...
+----------  ----------
+helloworld  0   
+```
+
+### artifact
+
+A description of a rustc compiler artifact being benchmarked. 
+
+This description includes:
+* name: usually a commit sha or a tag like "1.51.0"
```
Mark-Simulacrum, line 25: I think it'd be good to note here this is not a "git tag" and is basically free-form text.

rylev:
Even though the doc isn't complete, it's definitely in a good enough state to be useful, so I'm going to merge this.

PR #902, opened 2021-07-12
Noise and variance

rylev:
This is the start of an attempt to better quantify noise and variance. We do this by collecting the 100 previous commits and doing some small statistical analysis to determine if a particular benchmark has high noise or variance. We define noise and variance in the following ways:
* variance: the benchmark in question tends to produce "significant" changes in performance more often than we expect. Concretely, if a benchmark produces changes in performance larger than 2% more than 5% of the time, the benchmark is labeled as highly variable. This may be legitimate if a particular part of the compiler is being worked on often, but we don't expect significant changes to happen _that_ often. 
* noise: if a benchmark's "non-significant" changes (i.e., performance changes less than 2%) still show high variability, we consider the benchmark noisy. Concretely, of non-significant changes, if the average change is still above 0.1%, the data is noisy.

The calculations now drive the "??" labeling on benchmarks in the comparison page. Eventually, we'll want to better expose what type of variance and noisy we're finding for a particular benchmark so that we're not encouraging readers to just disregard any results labeled as noisy.

### DIFF
--- a/database/Cargo.toml
+++ b/database/Cargo.toml
@@ -4,8 +4,6 @@ version = "0.1.0"
 authors = ["Mark Rousskov <mark.simulacrum@gmail.com>"]
 edition = "2018"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 hashbrown = { version = "0.11", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -70,25 +70,6 @@ pub struct CommitResponse {
     pub commit: Option<String>,
 }
 
-pub mod data {
-    use crate::comparison::ArtifactData;
-    use collector::Bound;
-    use serde::{Deserialize, Serialize};
-
-    #[derive(Debug, Clone, Serialize, Deserialize)]
-    pub struct Request {
-        pub start: Bound,
-        pub end: Bound,
-
-        /// Which statistic to return data for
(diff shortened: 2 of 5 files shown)


Mark-Simulacrum:
I don't think I'll get a chance to review this in detail code wise - don't block on me.

I do think we should talk about how much we can push down the margin of false negatives though -- 2% is a pretty large change; sometimes performance work results in 0.5% changes and those (at least historically) are something we've judged as "statistically" significant (and this seems to prove itself out on graphs etc).

Maybe I'm misinterpreting?

rylev:
@Mark-Simulacrum You're not misinterpreting. I tried with a 0.5% change ratio and it makes a large portion (37%) of the benchmarks show up as highly variable meaning that they have "significant changes" more than 5% of the time. This could also just mean that a lot of our benchmarks are noisy and their range of noise goes beyond the 0.5% range often. I'm not yet sure how we should handle this

bjorn3:
I generally tend to use a higher threshold for single benchmarks, but a smaller threshold when a similar change happens across a lot of benchmarks. Eg if a single benchmark regresses 0.4% that is likely not significant. If many regress 0.2% it is likely significant.

Mark-Simulacrum:
> I generally tend to use a higher threshold for single benchmarks, but a smaller threshold when a similar change happens across a lot of benchmarks. Eg if a single benchmark regresses 0.4% that is likely not significant. If many regress 0.2% it is likely significant.

Yes, I think this is definitely a part of it. In any case, it need not block progress here - just something we should keep in mind.

PR #906, opened 2021-07-14
Fix copy/pasta mistake in Comparison::next and Comparison::prev

rylev:
This broke triage report generation.

Reminds me we should work on tests...


rylev:
Merging this as this is breaking the tests

PR #927, opened 2021-07-20
Add summary

rylev:
Adds summary above benchmark tables

image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -203,6 +203,11 @@
         .benchmark-name {
             text-align: center;
         }
+
+        .summary {
+            display: flex;
+            align-items: center;
+        }
     </style>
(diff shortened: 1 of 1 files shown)


lqd:
it would be good to vertically align these icons

rylev:
@lqd thoughts?
image

lqd:
I would _personally_ increase the right spacing between the value and the icons, and try to horizontally center the value + icon container to the columns below. But there's no need to spend time on this specifically, it looks fine as-is and already provides value to users.

rylev:
I agree regarding the spacing between the value and icons. As for aligning with the columns below, I'd like to avoid that since it visually implies that the values are related (which they're not). Perhaps having more a visual separator between the table and the summary could make this look cleaner. 

Going to merge any way. At some point, we can do another visual cleanup.

PR #931, opened 2021-07-21
Show diff in bootstrap timings

### LINKED ISSUE #929: Total perf diff for bootstrap timings no longer shown
image

rylev:
Fixes #929 

image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -184,7 +184,11 @@
             height: 10px;
         }
 
-        #benches td {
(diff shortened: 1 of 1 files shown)


bjorn3:
I was actually thinking about a percentage, but an absolute value is fine too I guess if you prefer that.

rylev:
@bjorn3 Thoughts?
image

bjorn3:
That would work.

Issue #938, opened 2021-07-26
A glossary of domain terms


rylev:
This is an attempt to give concrete terms to domain concepts that sometimes do not have set terminology within the codebase. If we can agree on these terms, we should move this glossary to the docs, and make sure the codebase consistently refers to these concepts using these terms. 

@wesleywiser and I devised this list in order to better be able to talk about how to model things within the codebase. 

# Proposal

## Basic terms

* **benchmark**: the source of a crate which will be used to benchmark rustc. For example, ["hello world"](https://github.com/rust-lang/rustc-perf/tree/master/collector/benchmarks/helloworld).
* **profile**: a [cargo profile](https://doc.rust-lang.org/cargo/reference/profiles.html). Note: the database uses "opt" whereas cargo uses "release". 
* **scenario**: The scenario under which a user is compiling their code. Currently, this is the incremental cache state and a change in the source since last compilation (e.g., adding a `println!` statement).  
* **metric**: a name of quantifiable statistic being measured (e.g., instruction count)
* **artifact**: a specific version of rustc (usually a commit sha or some sort of human readable "tag" like 1.51.0)

## Derived terms 
* **test case**: a combination of some benchmark, profile, and scenario.
* **test**: the act of running an artifact under a test case
* **test result**: the collection of all metrics as a result of a given artifact tested against a given test case. Each test result is composed of many iterations.
* **test iteration**: a single iteration that makes up a test case result. Note: we currently normal run 2 test iterations for each test. 
* **statistic**: a single value for a metric in a test case result
* **run**: a collection of test results for all currently available test cases run on a given artifact. Note: this term also means the process of collecting these test results.

Mark-Simulacrum:
I think it's good to note somehow the self profile "attribute" (kinda a boolean), as well as document vocabulary for that data, it doesn't quite fit this existing terminology I think.

rylev:
More terms:
* **stress test benchmark**: a benchmark that is specifically designed to stress a certain part of the compiler (e.g., [deeply-nested-async](https://github.com/rust-lang/rustc-perf/tree/master/collector/benchmarks/deeply-nested-async) is meant to stress parts of the compiler used in async code)
* **real world benchmark**: a benchmark based on a real world crate. These are typically copied as-is from crates.io. (e.g., [serde](https://github.com/rust-lang/rustc-perf/tree/master/collector/benchmarks/serde) is a popular crate and the benchmark has not been altered from a release of serde on crates.io). 
[...]

Issue #937, opened 2021-07-26
Rename database schema items for clarity and consistency


wesleywiser:
I've been looking at the current database schema to try to understand how it works. I've looked at the recently added [schema documentation](https://github.com/rust-lang/rustc-perf/blob/master/database/schema.md) and some of the names were difficult to decipher at a glance. For example, the column names `aid` and `cid` and the table names `pstat_series` and `pstats`. (What does the `p` stand for?)

I think it would help the reader understand the database schema faster if some of these things were renamed. I propose the following:

- `aid` columns -> `artifact_id`
- `cid` columns -> `collection_id`
- `crate` columns -> `benchmark_id`
- `series` columns -> `series_id`
- `benchmark.stablized` -> `benchmark.runs_on_stable`
- `pstats` table -> `statistics`
- `pstat_series` table -> `statistics_series`

This creates consistency in the schema in that all columns which reference other tables' primary keys are identified via the `_id` suffix and it makes it easier to understand what a column is at a glance by expanding some of the single letter abbreviations. 

This is relatively easy to do and the migrations to perform these renamings will run quickly since this only affects table metadata and not the content of the tables themselves.

Mark-Simulacrum:
> What does the p stand for?

process (i.e., the statistics are on a particular invocation of rustc)

I have no objections to renaming columns; migrations should essentially be automatic I think.

rylev:
A few nits:
* `statistics_series` might not be the most appropriate. This table houses a description of a test case and a particular metric (i.e., the elements that describe a statistics series). The actual series data (i.e., the values for that test case and metric pair over time) is derived from this description and the data in the `statistics` (nee `pstats`) table. The name `statistics_series` isn't bad, but my only concern would be that it might give the impression that the series data itself lives in that table. However, I can't really think of a better alternative.  
* `benchmark_id` will point to the benchmark name. Some prefer that columns with `id` point to autoincrementing, primary key integer columns not string data. I don't feel too strongly though.

PR #942, opened 2021-07-27
Consistent terminology

rylev:
This attempts to address #938 by moving the glossary into a docs folder and renaming variables and structs in the code to reflect the names in the glossary. 

You can find a rendered version of the glossary [here](https://github.com/rylev/rustc-perf/blob/consistent-terminology/docs/glossary.md).

This is only renaming of internal variable names so nothing that gets serialized should be touched. 

This will have to be a bit messy until we can rename database columns and tables as detailed in #937.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,0 +1,37 @@
+# Glossary
+
+The following is a glossary of domain specific terminology. Although benchmarks are a seemingly simple domain, they have a surprising amount of complexity. It is therefore useful to ensure that the vocabulary used to describe the domain is consistent and precise to avoid confusion. 
+
+## Basic terms
+
+* **benchmark**: the source of a crate which will be used to benchmark rustc. For example, ["hello world"](https://github.com/rust-lang/rustc-perf/tree/master/collector/benchmarks/helloworld).
+* **profile**: a [cargo profile](https://doc.rust-lang.org/cargo/reference/profiles.html). Note: the database uses "opt" whereas cargo uses "release". 
+* **scenario**: The scenario under which a user is compiling their code. Currently, this is the incremental cache state and a change in the source since last compilation (e.g., full incremental cache and a `println!` statement is added).  
+* **metric**: a name of quantifiable metric being measured (e.g., instruction count)
+* **artifact**: a specific version of rustc (usually a commit sha or some sort of human readable "tag" like 1.51.0)
+
+## Testing 
+
+* **test case**: a combination of a benchmark, a profile, and a scenario.
+* **test**: the act of running an artifact under a test case. Each test result is composed of many iterations.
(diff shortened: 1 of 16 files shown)


site/src/server.rs
```diff
@@ -581,27 +581,29 @@ pub async fn handle_graph(
         })
     }
 
-    let mut by_krate = HashMap::new();
-    let mut by_krate_max = HashMap::new();
+    let mut by_test_case = HashMap::new();
```
wesleywiser, line 584: Should this be `by_benchmark`?

wesleywiser: Oh, I see. This is a `HashMap<Benchmark, HashMap<Profile, Vec<Scenario>>>`. Nevermind.

docs/glossary.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Analysis
+
+* **significant test result delta**: a test result delta that is above some threshold that we have determined to be an actual change in performance and not noise. 
+* **highly sensitive benchmark**: a benchmark that tends to produce large (over some threshold) significant test result deltas. For example, some of the stress tests are highly sensitive; when they do produce significant test result deltas, the percentage change is typically quite large. 
+* **highly variable benchmark**: a benchmark that frequently produces (over some threshold) significant test result deltas. This differs from sensitive benchmarks in that the deltas aren't necessarily large, they just happen more frequently than one would expect.  Note: highly variable benchmarks can be classified as noisy. They are different from classically noisy benchmarks in that they often have produce deltas that are above the significance threshold. We cannot therefore be 100% if they are variable because they are noisy or because parts of the compiler being exercised by these benchmarks are just being changed very often.
```
wesleywiser, line 39:
```suggestion
* **highly variable benchmark**: a benchmark that frequently produces (over some threshold) significant test result deltas. This differs from sensitive benchmarks in that the deltas aren't necessarily large, they just happen more frequently than one would expect.  Note: highly variable benchmarks can be classified as noisy. They are different from classically noisy benchmarks in that they often produce deltas that are above the significance threshold. We cannot therefore be 100% if they are variable because they are noisy or because parts of the compiler being exercised by these benchmarks are just being changed very often. 
```

wesleywiser (commented):
I need to run a smoke test locally with these changes but overall they look good to me.

rylev:
I've checked over this again, and I'm pretty confident this only changes names and should not affect any functionality. I've tested as much as I can locally. I'm going to merge, and keep an eye on the service once the changes go live.

tgnottingham:
Did you want to change occurrences of `BuildKind`, `build_kind`, etc. to `Profile`, `profile`, etc.?

I know this PR was only for internal names, but at some point, do you want to change the `--runs` and `--builds` flags to `--scenarios` and `--profiles`?

rylev:
Yes, I wanted to start with internal facing names only, but once we feel comfortable with the terminology, we should change external facing names as well.

PR #946, opened 2021-07-29
Add more figures to the summary

### LINKED ISSUE #188: comparisons should include "summary" numbers
I would like it if comparisons included the "summary", "summary-check", and "summary-opt" numbers. It's very hard right now to get a kind of "overall" picture. Once we make more progress on established benchmark *suites* (i.e., for the scenarios in https://github.com/rust-lang/rust/issues/48547), then I would want those included (I'm assuming those would replace the "summary" numbers anyway).

### LINKED ISSUE #935: Summary line on comparison page does not update when filtering
I think giving a total might be useful even when filtering, but we should display the currently visible count too. This would let one easily check if (for example) ~all regressions are due to incremental

### LINKED ISSUE #953: summary counters off
https://perf.rust-lang.org/compare.html?start=4e282795d7d1d28a4c6c1c6521045ae2b59f3519&end=a0a6babf191847038860207075667effc9a67dec shows:

image

which seems to be a single regression, but the filtered view is not showing any regressions. Maybe we're counting the summary from the bootstrap data? On https://perf.rust-lang.org/compare.html?start=a0a6babf191847038860207075667effc9a67dec&end=f381e77d3590bc36f09b0d48cffb504f92febf5e though we don't see a perfect match even including the bootstrap data...

The unfiltered view seems to have several regressions (though most small), so maybe that's causing some problem?

rylev:
This adds more figures to the summary. 

**Note:** I've purposefully haven't worked on finishing the visual presentation. 

Screenshot 2021-07-29 at 15 05 15

The figures include:
* all: all test results including insignificant and filtered out results
* filtered: whatever has not been filtered out by a filter 
* scenario kinds: the various scenario kinds there are (full, incr-full, incr-unchanged, incr-patched) regardless if they have been filtered.

Also a border (red or green) is applied around any summary element that has more improvements or regressions than unchanged results. This is an attempt to draw attention to summary elements that are "significant". 

Questions:
* which summary elements are most useful? 
* Is it useful to highlight summary elements that show a "significant" number of improvements or regressions? If so, should it be how it's currently done (if there are more improvement or regression results than unchanged) or by some percentage of the total number of results? If by percentage, which?
* Are we missing any good summary figures?
* Should we try to summarize the summary (i.e., attempt to interpret the results for the viewer so that they do not have to)?

Fixes #935, fixes #188, fixes #953

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -210,7 +210,17 @@
 
         .summary {
             display: flex;
+            justify-content: center;
             align-items: center;
+            width: 20%;
+        }
+
+        #summary .improvement {
+            border: 1px solid green;
+        }
+
+        #summary .regression {
+            border: 1px solid red;
         }
     </style>
     <script src="https://cdn.jsdelivr.net/npm/vue@2/dist/vue.js"></script>
@@ -322,7 +332,8 @@
                         </li>
                         <li>
                             <label>
-                                <input type="checkbox" id="build-incremental-unchanged" v-model="filter.cache.incrUnchanged" />
+                                <input type="checkbox" id="build-incremental-unchanged"
+                                    v-model="filter.cache.incrUnchanged" />
                                 <span class="cache-label">incr-unchanged</span>
                             </label>
                             <div class="tooltip">?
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Might also fix https://github.com/rust-lang/rustc-perf/issues/953, not sure -- I had forgotten about this when opening that issue.

rylev:
This fixes #953 

I've removed the explicit summaries for the different scenario kinds. We can add those back if we want, but for now that information can be retrieved by explicitly filtering. 

image

PR #950, opened 2021-07-30
Remove unused self profile query time

rylev:
@Mark-Simulacrum this used to be hidden behind query for `f64` series, but since that was simplified in #944, this code is n longer used. I've checked the site, and everything seems to be working without it. Can you confirm that this indeed is not needed?


Mark-Simulacrum (approved):
The code here is pretty obvious and I believe you are correct that nothing is using it; we can re-introduce it later if necessary.

IIRC, this was added as an "alternative" to https://github.com/rust-lang/rustc-perf/pull/661 -- well, an earlier version of that. If we want to support graphing, something like this code is likely desirable, but I haven't put too much thought into it.

Mark-Simulacrum:
(I'll leave it up to you to merge).

PR #967, opened 2021-08-13
Test Databases in CI

### LINKED ISSUE #966: Test server initialization in CI
There's a lot of initialization code that doesn't go through CI. What's worse, because we often run sqlite locally and postgres in production we can miss things that work in sqlite and not in postgres - see #957. 

We should at least run the sever with an empty sqlite and postgres database to make sure that the initialize happens properly.

rylev:
Fixes #966

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,3 +131,33 @@ jobs:
       - name: Check benchmarks
         run: sh -x -c "ci/check-profiling.sh"
 
+  database-check-postgres:
+    name: Database Check (postgres)
+    runs-on: ubuntu-latest
+    services:
+      postgres:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -131,3 +131,52 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Install stable
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Connect to PostgreSQL
+        run: |
+          cargo build --bin site --release
```
tgnottingham, line 161: Could get away with a debug build, like in `test_benchmarks`, etc.

PR #969, opened 2021-08-15
Use pointer hash rather than str hash when interning strings

tgnottingham:
Guarantee that `ArenaStr`s are interned so that hashing can be
pointer-based and interning can be more efficient.

Rename `ArenaStr` to `InternedArenaStr`.

### DIFF
--- a/intern/src/lib.rs
+++ b/intern/src/lib.rs
@@ -15,7 +15,7 @@ pub trait InternString {
 #[macro_export]
 macro_rules! intern {
     (pub struct $for_ty:ident) => {
-        #[derive(Serialize, Debug, Copy, Clone)]
+        #[derive(Serialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
         pub struct $for_ty($crate::ArenaStr);
 
         impl $for_ty {
@@ -23,14 +23,6 @@ macro_rules! intern {
                 self.0.as_str()
             }
         }
-
-        impl std::cmp::PartialEq for $for_ty {
-            fn eq(&self, other: &Self) -> bool {
-                self.0.hash_ptr() == other.0.hash_ptr()
-            }
-        }
-
-        impl std::cmp::Eq for $for_ty {}
 
         impl std::cmp::PartialOrd for $for_ty {
             fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
@@ -41,12 +33,6 @@ macro_rules! intern {
         impl std::cmp::Ord for $for_ty {
             fn cmp(&self, other: &Self) -> std::cmp::Ordering {
                 self.0.as_str().cmp(other.0.as_str())
-            }
-        }
-
-        impl std::hash::Hash for $for_ty {
-            fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-                state.write_usize(self.0.hash_ptr());
             }
         }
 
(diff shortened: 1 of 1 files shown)


tgnottingham:
The difference this makes is small, admittedly, but since we're already interning for efficiency, may as well make it a little more so.

This mainly moves the bulk of `preloaded` and `intern` into `InternedArenaStr` static methods to guarantee that any instances are interned, and updates `Hash`, `PartialEq`, and `Eq` implementations appropriately. The `Hash` implementation for `InternedArenaStr` is the one that was changed from being `str`-based to pointer-based.

tgnottingham:
CI issues seem to be unrelated to the change, as the same thing is happening on another PR. A bunch of `No such file` errors.

```
[2021-08-15T23:31:57Z INFO  cargo::core::compiler::fingerprint] fingerprint error for await-call-tree v0.1.0 (/tmp/.tmpDpl4hN)/Doc { deps: false }/TargetInner { ..: lib_target("await-call-tree", ["lib"], "/tmp/.tmpDpl4hN/src/lib.rs", Edition2018) }
[2021-08-15T23:31:57Z INFO  cargo::core::compiler::fingerprint]     err: failed to read `/tmp/.tmpDpl4hN/target/debug/.fingerprint/await-call-tree-f7705a294dd743d7/doc-lib-await-call-tree`
    
    Caused by:
        No such file or directory (os error 2)
```

Mark-Simulacrum:
Those fingerprint errors are normal; the real problem is in piston-image somehow. My guess is a recent change to fingerprints/depinfo in rustc or cargo is causing this (it's also failing on master builds - https://perf.rust-lang.org/status.html)

intern/src/lib.rs
```diff
@@ -9,29 +9,21 @@ use std::ptr::NonNull;
 use std::sync::Arc;
 
 pub trait InternString {
-    unsafe fn to_interned(s: ArenaStr) -> Self;
+    unsafe fn to_interned(s: InternedArenaStr) -> Self;
```
rylev, line 12: I know this wasn't documented before, but can we add documentation on the safety requirements for this `unsafe` function?

intern/src/lib.rs
```diff
@@ -130,74 +116,32 @@ macro_rules! intern {
...
 lazy_static::lazy_static! {
-    static ref INTERNED: (ArcSwap<HashSet<ArenaStr>>, Mutex<(HashSet<ArenaStr>, Bump)>)
+    static ref INTERNED: (ArcSwap<HashSet<InternedArenaStr>>, Mutex<(HashSet<InternedArenaStr>, Bump)>)
         = (ArcSwap::new(Arc::new(HashSet::new())), Mutex::new((HashSet::new(), Bump::new())));
 }
 
 pub fn preloaded<T: InternString>(value: &str) -> Option<T> {
-    let set = INTERNED.0.load();
-    unsafe { Some(T::to_interned(*set.get(value)?)) }
+    InternedArenaStr::preloaded(value).map(|s| unsafe { T::to_interned(s) })
```
rylev, line 124: Similar to above, it would be great to document the use of `unsafe` and why this particular usage is safe. We should make sure we document all usages of unsafe thoroughly.

rylev (changes requested):
I personally feel a little uncomfortable with the usage of unsafe in this crate especially considering the performance implications aren't really clear to me. However, that predates this PR so we shouldn't block on that. In the meantime, I'd much prefer if we could thoroughly document the usage of unsafe.

rylev:
@Mark-Simulacrum @tgnottingham Should we file an issue in rust-lang/rust for the fingerprinting issue?

Mark-Simulacrum:
Yeah, that seems good -- worth additional investigation.

tgnottingham:
I just noticed that this change is incorrect due to `InternedArenaStr` implementing `Borrow<str>` but not implementing the `Hash`, etc. traits equivalently to `str`. This breaks interning in the sense that it results in multiple copies of the same string being allocated.

I think it was also incorrect before this change, because `PartialEq` wasn't equivalent with `str`'s, plus `PartialEq` and `Hash` were inconsistent (former was pointer-based, latter was `str`-based). Interestingly, this didn't seem to break interning. I assume this is just due to implementation details, and that the inconsistency should be fixed, but please correct me if I'm misunderstanding.

I'll close this and address the above issue in another PR if necessary. I'll also look at documenting the `unsafe` code.

Mark-Simulacrum:
Yeah I think I agree with @rylev that we should see about dropping the interning -- it was added back when we were still loading *all* of our data on each start up, and I'd guess is significantly much less necessary/effective today. I think some quick experiments to check whether it's necessary would be good.

tgnottingham:
Opened #974 for reevaluating interning.

PR #971, opened 2021-08-16
Only show "Incremental loading delta" when comparing two commits

tgnottingham:


tgnottingham:
Before:

before

After:

after

Still shows when comparing two commits:

after_compare

rylev (approved):
👍

PR #977, opened 2021-08-18
Adjust confidence levels based on experience from past triage report

rylev:
This adjusts the "confidence" levels (i.e., how confident we are a performance change is worth looking at) in the following ways:
* Comparisons with any large or very large changes are deemed definitely relevant.
* Comparisons containing more than 1 medium change are also definitely relevant.
* Comparisons containing exactly 1 medium change are probably relevant
* Comparisons containing enough small and very small changes such that `small * 2 + very_small > 10` changes are considered probably relevant.
* Everything else is maybe relevant.

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -260,34 +260,28 @@ impl ComparisonSummary {
     }
 
     pub fn confidence(&self) -> ComparisonConfidence {
+        let mut num_very_small_changes = 0;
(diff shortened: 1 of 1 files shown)


tgnottingham (commented):
This looks good to me. The `small * 2 + very_small > 10` addition looks like it could be pretty sensitive, but that's not necessarily a bad thing.

PR #978, opened 2021-08-18
Update flamegraph and crox support

tgnottingham:
Add support for generating flamegraph and crox files from newer
self-profile format, and remove support for old format. Fixes broken
flamegraph and crox links on detailed query page for self-profile data
in newer format. Breaks links for self-profile data in older format,
for what it's worth.

### DIFF
--- a/site/src/request_handlers/self_profile.rs
+++ b/site/src/request_handlers/self_profile.rs
@@ -25,14 +25,33 @@ pub async fn handle_self_profile_processed_download(
         body.benchmark,
         body.run_name
     );
+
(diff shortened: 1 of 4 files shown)


tgnottingham:
I just noticed the closed #788 by @lqd, which was going to support both formats. Instead of merging this PR, I could take some time to review #788 and update it if necessary to help get it merged.

Mark-Simulacrum:
Yeah I think it's been long enough now that the wrapping isn't really necessary and could also be added later if we really wanted to, ultimately, this seems good to merge for now.

PR #982, opened 2021-08-22
Categorize percents less than -1% as negative, not those less than 1%

tgnottingham:
Fixes certain delta percent changes between -1% and 1% being regarded as
improvements (and colored green), rather than neutral.


tgnottingham:
Before, 0.9% regression and 0.9% improvement would be colored green:
before

After:
after

rylev (approved):

PR #996, opened 2021-09-08
Change significance to be determined by IQR fencing

rylev:
This changes how we define (and subsequently implement in code) a "significant" test result to a more formal and less arbitrary mechanism (described below). Additionally, the documentation is updated to reflect this change.  

**Before**
Until we've used a simple threshold of either 0.2% change for non-"dodgy" test cases (i.e., test cases which we've determined to *not* have some sort of historical noise) and 0.8% for "dodgy" test cases.

**After**
Significance is defined as being an outlier when compared with historical data. We use interquartile range fencing to determine whether a given result is an outlier. 

IQR fencing uses this formula:

```
interquartile_range = Q3 - Q1
result > Q3 + (interquartile_range * 1.5)
```

### DIFF
--- a/docs/comparison-analysis.md
+++ b/docs/comparison-analysis.md
@@ -18,11 +18,11 @@ At the core of comparison analysis are the collection of test results for the tw
 100 * ((statisticForArtifactB - statisticForArtifactA) / statisticForArtifactA)
 ```
 
-## High-level analysis description 
+## High-level analysis description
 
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum (approved):

rylev:
I'd like to hold off on merging this for now since it makes a lot more comparisons show up as "definitely relevant". We should address that first.

PR #999, opened 2021-09-10
Add a severity column

### LINKED ISSUE #997: Introduce new "severity" column in comparison
Especially with the introduction of a mechanism for deciding on a non-hardcoded significance threshold in #996, it is nice to know by what factor a particular test result is over the significance threshold. I propose to add a "severity" column which shows by how much the test result is over the significance threshold. 

For example, for a test result of 1.5% and a significance threshold of 0.5%, the "severity" would be 3.0.

rylev:
Fixes #997 

We now show by how much a given test result is over it's significance threshold. Unfortunately it's not super easy to know what this just from looking at it. We may want to consider adding headers or a legend. 

image

### DIFF
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -187,6 +187,7 @@ pub mod comparison {
         pub profile: String,
         pub scenario: String,
         pub is_significant: bool,
+        pub significance_threshold: f64,
         pub is_dodgy: bool,
         pub historical_statistics: Option<Vec<f64>>,
         pub statistics: (f64, f64),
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Headers seems like a good idea, and noting somewhere what range the values are in. I'd guess 1x is sort of "as little change as could be", if we're showing it on the filtered view?

Mark-Simulacrum:
Going to go ahead and merge this -- we can continue to play with it, of course, but it seems like a solid improvement for now.

(Selfishly, I want these for perf triage...)

PR #1003, opened 2021-09-14
Refactor significance calculations

Mark-Simulacrum:
This moves significance factor calculation to the backend and does some refactoring around that, but doesn't actually change the calculation significantly.

The minor difference is that the significance threshold is calculated as `median + ...` rather than `q3 + ...`, which seems more appropriate. Basing the significance threshold at the q3 marker introduces a sort of "partial iqr" and that feels fishy.

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -114,7 +114,7 @@ pub async fn handle_compare(
             scenario: comparison.scenario.to_string(),
             is_dodgy: comparison.is_dodgy(),
             is_significant: comparison.is_significant(),
-            significance_threshold: comparison.signifcance_threshold() * 100.0,
+            significance_factor: comparison.significance_factor(),
             historical_statistics: comparison.variance.map(|v| v.data),
             statistics: comparison.results,
         })
@@ -658,7 +658,30 @@ impl BenchmarkVariance {
         deltas
     }
 
-    fn upper_fence(&self) -> f64 {
+    // This is an absolute value indicating the noise barrier for changes on
+    // this benchmark.
+    //
+    // A number line could be divded like this:
+    //
+    // ------o-------o----------
+    //       ^   ^   ^
+    //       |   |   |
+    //       |   |   |
+    //       |   |   ---- +significance_threshold
+    //       |   |
+    //       |   - not significant, includes zero
+    //       |
+    //       ---- -significance_threshold()
+    fn significance_threshold(&self) -> f64 {
+        let (q1, median, q3) = self.quartiles();
+
+        // Changes that are IQR_MULTIPLIER away from the median are considered
+        // significant (i.e. outliers).
+        median + (q3 - q1) * Self::IQR_MULTIPLIER
+    }
+
+    // (q1, q2=median, q3)
+    fn quartiles(&self) -> (f64, f64, f64) {
         let pcs = self.percent_changes();
         fn median(data: &[f64]) -> f64 {
             if data.len() % 2 == 0 {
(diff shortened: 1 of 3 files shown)


site/src/comparison.rs
```diff
@@ -818,14 +843,22 @@ impl TestResultComparison {
...
             self.variance
                 .as_ref()
-                .map(|s| s.upper_fence())
+                .map(|v| v.significance_threshold())
                 .unwrap_or(Self::SIGNIFICANT_RELATIVE_CHANGE_THRESHOLD)
         }
     }
 
+    /// This is a numeric magnitude of a particular change.
+    fn significance_factor(&self) -> Option<f64> {
```
rylev, line 852: What's the reason for this being an optional?

site/src/comparison.rs
```diff
@@ -658,7 +658,30 @@ impl BenchmarkVariance {
...
+    //       |   |   |
+    //       |   |   ---- +significance_threshold
+    //       |   |
+    //       |   - not significant, includes zero
+    //       |
+    //       ---- -significance_threshold()
+    fn significance_threshold(&self) -> f64 {
+        let (q1, median, q3) = self.quartiles();
+
+        // Changes that are IQR_MULTIPLIER away from the median are considered
```
rylev, line 678: Can you talk more about why you changed calculation from `q3 + IQR` to `median + IQR`? I believe [outlier fences](https://www.statology.org/upper-and-lower-fences/) are normally calculated from the quartiles and not from the median.

site/src/comparison.rs
Mark-Simulacrum: Hm, I think this was leftover from an earlier version I had locally. The reasoning was that we may not always be able to compute a change's magnitude in a reasonable way -- we approximate significance with the 0.002 threshold constants, but there's no such thing really for "how significant" if there's no prior data.

I actually think this may be a little buggy in the sense that I think we should show something like "-" in the UI if we don't have enough historical data to judge the factor, not just scale based on the 0.002 threshold, which would be misleading.

site/src/comparison.rs
Mark-Simulacrum: Hm. So my reasoning here was that it *felt* like using the median is somehow "better" -- adding IQR to the median seems more reasonable than adding it to Q3, which sort of already includes part of the IQR. This was also partially based on my (incorrect) recollection was that my statistics classes did have it based on the median, though since the pages you've linked to do all have q3, I was probably just wrong here :)

In practice our median and q3 are usually pretty close, so this probably has a negligible difference.

rylev: Hmm I'm not sure if it feels better to me. Perhaps you could try to explain more why it does?

In any case if it's unlikely to make a huge difference, perhaps we should just stick with the text book definition?

Mark-Simulacrum: I'm happy to revert this particular change.

I think it mostly felt like we ended up with something like

`median + IQR * (some coefficient representing how "right skewed" we are) + IQR * IQR_MULTIPLIER`

where the middle term is the Q3-median delta. That felt a little weird for our purposes, where we don't actually care about the right-skew that much, in practice we will always have high right skew since all data is going to be bound to [0, infinity) so there's practically no way to be left skewed.

rylev: That makes sense. I don't feel too strongly about this which I guess is why my gut says to fallback to the book mechanism, but I'll leave it to you. 

However, can you update the docs where we explain this? https://github.com/rust-lang/rustc-perf/blob/master/docs/comparison-analysis.md#what-makes-a-test-result-significant

Mark-Simulacrum: I'll just revert this change I think, that seems best. I'll update the constant 1.5 in the docs as well.

Mark-Simulacrum: https://github.com/rust-lang/rustc-perf/pull/1008

PR #1008, opened 2021-09-15
Update documentation for changes to significance

Mark-Simulacrum:
Also reverts to using the Q3 + IQR*c algorithm, rather than median.


rylev (approved):
🎉

PR #1010, opened 2021-09-16
Fix new warnings generated on nightly

Mark-Simulacrum:
First commit fixes a bug, and could affect benchmarks -- particularly if the string-wise path sort didn't match the numeric sort. Not sure if that happened in practice.

Second commit just drops an unused struct.

cc https://github.com/rust-lang/rust/issues/88900 (actionable and real cases, fwiw)

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -1166,9 +1166,8 @@ impl Benchmark {
             }
         }
 
-        patches.sort();
-
-        let patches = patches.into_iter().map(|p| Patch::new(p)).collect();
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
We'll likely want to be a little careful around deploying this -- any builds over the edge (before this PR / after this PR) may have changes as a result of the first commit here. But we do want that patch so I think we should just do this and keep an eye out for any weird perf results (most of which will go away with a re-run if they're try builds).

cc @pnkfelix @rylev -- may arise in triage next week, so pinging here. Not sure who is up next week. Likely to be a change limited to just 1-2 patched incr scenarios, with (probably) high numeric indices.

Issue #980, opened 2021-08-22
Perf summaries completely ignore rustdoc


jyn514:
This perf run had over 100% regressions for rustdoc, but rust-timer said there were no significant changes.

_Originally posted by @rust-timer in https://github.com/rust-lang/rust/issues/88219#issuecomment-903230805_

jyn514:
Relevant code seems to be https://github.com/rust-lang/rustc-perf/blob/5c8e64ac2a98f48573021f2a5102d535c680d7bb/site/src/github.rs#L704 / https://github.com/rust-lang/rustc-perf/blob/5c8e64ac2a98f48573021f2a5102d535c680d7bb/site/src/comparison.rs#L270.

@rylev do you know when you'll get a chance to look into this? Posting "no relevant changes" when there's a regression is worse than not posting anything at all IMO.

jyn514:
Or alternatively, how can I run that code locally? `cargo run --bin site` seems to run the site, not the bot.

jyn514:
Ugh, they're being deliberately filtered out for some reason.
https://github.com/rust-lang/rustc-perf/blob/5c8e64ac2a98f48573021f2a5102d535c680d7bb/site/src/comparison.rs#L571-L573
We discussed removing this in https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/triage.202021.2009.2007/near/252361701 so I'm going to go ahead and make a PR that doesn't filter them out.

PR #1011, opened 2021-09-17
Don't ignore rustdoc when generating comparison summaries

jyn514:
Fixes https://github.com/rust-lang/rustc-perf/issues/980.


rylev:
This looks good to me, but I'd like @Mark-Simulacrum's perspective on why doc test cases were filtered out in the first place (as that decision predates my involvement in the project).

Mark-Simulacrum:
Yeah, this is fine with me as well. I think when we just added the python triage script doc builds were still pretty recently added, and I recall numerous comments from @jyn514 on regressions that they should basically be ignored. I think those reasons no longer hold, though, and based on the perf runs I've seen doc builds don't behave particularly unusually. So I think we can proceed here.

PR #1016, opened 2021-09-18
mono-item profiler

Mark-Simulacrum:
This adds another profiler which runs the benchmark with -Zprint-mono-items and then slightly parses the output to split into per-CGU lists of mono items. These are the post-merging CGUs.

Posting this to get it merged into master; I'm using this locally to mediocre success but I feel like it is probably at least a little useful to have. Eventually I'd like to work some more on it and get this displayed as part of our results on each benchmark, so we can e.g. give warnings when CGU allocation changes. Potentially there's some useful work to show the "forest" of CGUs that got merged into the final set as well, but I am not sure that is genuinely helpful (vs. pretty).

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -13,6 +13,7 @@ use std::fmt;
 use std::fs::{self, File};
 use std::hash;
 use std::io::Read;
+use std::io::Write;
 use std::mem::ManuallyDrop;
 use std::path::{Path, PathBuf};
 use std::process::{self, Command};
@@ -178,6 +179,7 @@ pub enum Profiler {
     Massif,
     Eprintln,
     LlvmLines,
+    MonoItems,
 }
 
 impl Profiler {
@@ -198,6 +200,7 @@ impl Profiler {
             "massif" => Ok(Profiler::Massif),
             "eprintln" => Ok(Profiler::Eprintln),
             "llvm-lines" => Ok(Profiler::LlvmLines),
+            "mono-items" => Ok(Profiler::MonoItems),
             _ => Err(anyhow!("'{}' is not a known profiler", name)),
         }
     }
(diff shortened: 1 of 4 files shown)


bjorn3:
> Potentially there's some useful work to show the "forest" of CGUs that got merged into the final set as well, but I am not sure that is genuinely helpful (vs. pretty).

I can see that be useful when trying to figure out why the cgu merging has changed.

tgnottingham (commented):
May want to add an entry for `mono-items` to `collector/README.md`.

collector/src/execute.rs
```diff
@@ -1132,6 +1138,51 @@ impl<'a> Processor for ProfileProcessor<'a> {
...
+            // mono item results are redirected (via rustc-fake) to a file
+            // called `mono-items`. We copy it from the temp dir to the output
+            // dir, giving it a new name in the process.
+            Profiler::MonoItems => {
+                let tmp_file = filepath(data.cwd.as_ref(), "mono-items");
+                let out_dir = self.output_dir.join(&out_file("mono-items"));
+                let _ = fs::create_dir_all(&out_dir);
+                let result_file = filepath(&out_dir, "raw");
+
+                fs::copy(&tmp_file, &result_file)?;
```
rylev, line 1150: Are these result files severed from the site or is this currently only useable locally?

rylev (approved):
Looks good to me as a start! Thanks 🎉

collector/src/execute.rs
Mark-Simulacrum: These are all just usable locally, the perf collector server doesn't hit this function at all, I believe.

Eventually I'd like to surface this specific profile in the site - potentially integrated with the recently added codegen schedule graphs - but that's future work.

PR #1019, opened 2021-09-18
Support directly rendering codegen scheduling

Mark-Simulacrum:
This cherry picks a few of the commits from https://github.com/rust-lang/rustc-perf/pull/1005 to build up the ability to use self profile with sqlite (currently using an export of the prod db locally).

The schedule is rendered onto canvas, essentially a subset of the view that crox (and then chrome profiler tooling) provides, but easier to get to and with significantly less complexity. We'll likely want to continue iterating on how to get to these views (currently integrated into the self profile page) and what exactly is displayed -- this isn't really going to mesh well with any kind of query nesting, but seems like a good start still.

Graph drawing is at least partly inspired by Cargo's -Ztimings view.

Example:

rendered page

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -128,7 +128,7 @@ series      aid         cid         value
 
 **TODO**
 
-### pull_request_builds
+### pull_request_build
 
 **TODO**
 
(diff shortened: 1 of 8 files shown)


tgnottingham:
This is really awesome! Feeling inspired to improve codegen scheduling.

PR #1023, opened 2021-09-21
Add ability to filter out 'very small' changes

### LINKED ISSUE #1006: Add filtering by "magnitude" to the comparison page
Sometimes significant results aren't interesting because their magnitude is so small. We should provide a way to filter by the magnitude of the change and provide a sensible default. This is likely to be different depending on the stat, but for now anything above 0.2% (which has traditionally been the significance threshold) might be a good start.

rylev:
Fixes #1006.

This introduces a new filter (on by default) which filters out any changes that are deemed to have a "very small" magnitude. You can find the definition for magnitude [here](https://github.com/rust-lang/rustc-perf/blob/master/docs/comparison-analysis.md#magnitude).

### DIFF
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -201,6 +201,7 @@ pub mod comparison {
         pub is_significant: bool,
         pub significance_factor: Option<f64>,
         pub is_dodgy: bool,
+        pub magnitude: String,
         pub historical_statistics: Option<Vec<f64>>,
         pub statistics: (f64, f64),
     }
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Not sure how we should fix that, though -- I lean towards not showing very small changes in perf summaries, under the assumption that they are likely to be irrelevant in practice (same reason we're hiding them in comparison pages).

rylev:
So we only show the largest individual changes for either improvements or regressions in the triage and PR summaries. Because of this, those changes are only going to be "very small" if there are a decent number of very small changes. So it is true that it's possible that the comparison page doesn't show the change but the summary does, but only in specific situations.

There's only one way to fully fix this IMO: Don't use very small changes _at all_ in determining perf categorizations. Thoughts on this?

Mark-Simulacrum:
Yeah, I think that's the way to go.

rylev:
@Mark-Simulacrum I've gone ahead and filtered out very small changes from the categorization determination.

PR #1025, opened 2021-09-21
Support auto-installing toolchain with rustup-toolchain-install-master

Mark-Simulacrum:


collector/src/main.rs
```diff
@@ -397,15 +397,41 @@ fn get_local_toolchain(
...
+            .args(&["which", "rustc", "--toolchain", &toolchain])
+            .output()
+            .context("failed to run `rustup which rustc`")?;
+
+        // Looks like a commit hash? Try to install it...
+        if output.status.code() == Some(101) && toolchain.len() == 40 {
+            // No such toolchain exists, so let's try to install it with
+            // rustup-toolchain-install-master.
+
+            if !Command::new("rustup-toolchain-install-master")
```
rylev, line 411: A common failure mode here would like be not having rustup-toolchain-install-master installed. Should we specifically handle that case?

Mark-Simulacrum: Yeah, we actually have sysroot building code built-in already and we should use that, I had forgotten about it when writing this.

PR #1027, opened 2021-09-22
Show data for queries present only in one profile in comparison

### LINKED ISSUE #1022: detailed-query page missing some lines from table
https://perf.rust-lang.org/detailed-query.html?sort_idx=12&commit=635ba4d6ff55c13d3f40dd9b82d773d172377606&base_commit=60e70cc909f84e84ed3bca3fe3459ddceefdd08c&benchmark=await-call-tree-debug&run_name=incr-full

`Executions delta: -35319.000` total
And only two rows with non zero values
```
2795 | -7064.000  (-71.7%)
2880 | -7064.000  (-71.0%)
```
So sum of rows didn't equal total.

Similar example here: https://perf.rust-lang.org/detailed-query.html?commit=635ba4d6ff55c13d3f40dd9b82d773d172377606&base_commit=60e70cc909f84e84ed3bca3fe3459ddceefdd08c&benchmark=await-call-tree-debug&run_name=full

total: -14127 and zero non zero lines
P.S. executions delta column should be integer type.

tgnottingham:
Add showing rows for self-profile queries present only in base profile,
and add showing deltas for queries present only in main profile.

Fixes #1022.

### DIFF
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -310,7 +310,7 @@ pub mod self_profile {
     #[derive(Serialize, Debug, Clone)]
     pub struct SelfProfileDelta {
         pub totals: QueryDataDelta,
-        pub query_data: Vec<Option<QueryDataDelta>>,
+        pub query_data: Vec<QueryDataDelta>,
     }
 
     #[derive(Serialize, Clone, Debug)]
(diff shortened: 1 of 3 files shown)


tgnottingham:
Where base profile doesn't have `lit_to_const` query:

comparison_positive

Where only base profile has `lit_to_const` query:

comparison_negative

tgnottingham:
Added showing executions delta as integer as well:

integer_executions_delta

Mark-Simulacrum:
I think you should have permission to merge, so please do so when you think this is good -- I gave a quick skim and the changes look good to me :)

Generally don't feel you *need* to wait on review, particularly for UI improvements like this.

tgnottingham:
Alright, thanks!

PR #1036, opened 2021-09-28
Change the threshold for what constitutes a very large change from 18x the sig threshold to 12x

rylev:
These numbers are arbitrary so this is a bit more of an art than a science. However, this does move the threshold to be double the previous threshold which is much more mathematically satisfying.


Mark-Simulacrum (approved):
Seems OK. 12x the threshold feels pretty "large" to me, but then, so does 6x, so :shrug:

rylev:
Yes - everything from 6-12x the significance threshold is a large change and everything 12x and up is a very large change.

PR #1042, opened 2021-10-02
Remove support for the old graph page and refactor graph handler

tgnottingham:


tgnottingham:
Visually verified that the graphs produced for the last several months of data are the same before and after this change. The dump of the production database came in handy here.

Intending to follow this up with something to improve the graph handler's performance.

rylev (approved):
🎉This looks great. Thanks for taking the time to refactor.

tgnottingham:
Thanks! Merged.

PR #1043, opened 2021-10-02
Run cachegrind output through rustfilt

erikdesjardins:
I have `valgrind-3.16.1`, which should be able to demangle Rust symbols (according to the changelog, support was added in 3.13.0), but it doesn't do a very good job...

e.g. `cargo +nightly run --release --bin collector -- diff_local cachegrind --include helloworld --builds Opt --runs IncrUnchanged +3b3ce374d203445eb1d0dce50f4211f4aceb7db6 +20997f6ad81721542e9ef97bb2f58190903a34d8` outputs:

```
-10,848  ???:rustc_query_system::query::plumbing::incremental_verify_ich
 -6,532  ./string/../sysdeps/x86_64/strcmp.S:strcmp
  5,320  ???:_ZN18rustc_query_system5query8plumbing22incremental_verify_ich17h7b6e470314c3c592E
 -3,960  ???:llvm::PMDataManager::initializeAnalysisImpl(llvm::Pass*)
  3,945  ???:llvm::PassRegistry::registerPass(llvm::PassInfo const&, bool)
  2,792  ???:_ZN18rustc_query_system5query8plumbing22incremental_verify_ich17hd69396997e117d33E
  2,736  ???:_ZN18rustc_query_system5query8plumbing22incremental_verify_ich17h6c9f4a8f7833e3a5E
...
```
(with rustfilt the incremental_verify_ich clones get demangled and they cancel out)

### DIFF
--- a/collector/src/main.rs
+++ b/collector/src/main.rs
@@ -510,10 +510,22 @@ fn generate_cachegrind_diffs(
                 let id_diff = format!("{}-{}", id1, id2);
                 let cgout1 = out_dir.join(filename("cgout", id1));
                 let cgout2 = out_dir.join(filename("cgout", id2));
+                let cgfilt1 = out_dir.join(filename("cgfilt", id1));
+                let cgfilt2 = out_dir.join(filename("cgfilt", id2));
                 let cgdiff = out_dir.join(filename("cgdiff", &id_diff));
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Yeah, I've seen the same as well. Not sure what is up with cachegrind, but seems reasonable to just patch over it.

PR #1047, opened 2021-10-03
Use new significance algorithm by default

Mark-Simulacrum:
This currently just completely removes the old algorithm (at least where it was if-gated). I think we could allow switching back, but I've not really seen a use case for that (you can always disable filtering by significance in the UI).

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -48,7 +48,7 @@ pub async fn handle_triage(
             "instructions:u".to_owned(),
             ctxt,
             &master_commits,
-            body.calcNewSig.unwrap_or(false),
+            body.calcNewSig.unwrap_or(true),
         )
         .await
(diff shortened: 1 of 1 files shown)


rylev (changes requested):
I think we should keep the old algorithm around for a little bit. If we happen to see interesting or surprising results, it will be good to check against the old algorithm

Mark-Simulacrum:
Alright, replaced with just swapping the default, I believe.

rylev (changes requested):
Can you also change the default [here](https://github.com/rust-lang/rustc-perf/blob/265aa9447e37a935960aaa22b6f6f2fdd685ca9e/site/src/comparison.rs#L356)?

Mark-Simulacrum:
Yes, yes. Sorry for being so bad at identifying these places :)

PR #1052, opened 2021-10-04
Make single query for graph data

tgnottingham:
Make one query for graph data instead of sixteen. With local SQLite DB,
this reduces the time to handle a request for 8 months of data from 8
seconds to 3. Hoping that the results are even better for Postgres, as
currently, such a request causes a timeout on perf.rust-lang.org.


tgnottingham:
Verified that graphs look the same before and after change.

tgnottingham:
Rebased.

PR #1055, opened 2021-10-05
Use term 'profile' instead of 'build' as that's what we're using elsewhere

rylev:
This brings some of the vocabulary closer in line with what we use elsewhere including in the glossary.


Mark-Simulacrum (approved):
Looks reasonable with CI passing.

PR #1056, opened 2021-10-05
Do not set calcNewSig to false in request

Mark-Simulacrum:
Missed another spot where this needs adjusting...

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -819,9 +819,13 @@
             let values = Object.assign({}, {
                 start: "",
                 end: "",
+                calcNewSig: true,
                 stat: "instructions:u",
             }, state);
-            values["calcNewSig"] = values.calcNewSig === 'true';
+            // Make sure we pass a boolean.
+            if (typeof values.calcNewSig === "string") {
+                values.calcNewSig = values.calcNewSig === "true";
+            }
             makeRequest("/get", values).then(function (data) {
                 app.data = data;
             });


site/static/compare.html
```diff
@@ -819,9 +819,13 @@ <h2>Comparing <span id="stat-header">{{stat}}</span> between <span id="before">{
             let values = Object.assign({}, {
                 start: "",
                 end: "",
+                calcNewSig: true,
                 stat: "instructions:u",
             }, state);
-            values["calcNewSig"] = values.calcNewSig === 'true';
+            // Make sure we pass a boolean.
+            if (typeof values.calcNewSig === "string") {
```
rylev, line 826: Can you explain why we need to do this?

Mark-Simulacrum: I haven't verified, but I believe serde will fail to deserialize `calcNewSig:"true"` as a Option<bool>, so we need to force the conversion to boolean in JS for the case where it's set from the query parameters

rylev: Yea that seems true, but we were already setting `calcNewSig` to a `bool`. This seems to just conditionally set if it were previously a `string`. I'm just confused why we need to add this conditional.

Mark-Simulacrum: Without this patch, values.calcNewSig is undefined (since it's not mentioned on the query string), and that is `!== "true"`, so we get false and don't enable the new alg.

If it *is* mentioned, we need to "deserialize" it with the `=== "true"` but if we do that unconditionally we'll end up setting it to false (since we've put a boolean in the map). I guess an alternative could be to put `"true"` in the default object here.

rylev (approved):

PR #1061, opened 2021-10-06
Fix unit tests and run them in CI

### LINKED ISSUE #1060: Run unit tests in CI
We are currently not running unit tests in CI which is made evident by the fact that there is currently one failing unit test, but CI continues to pass.

rylev:
Fixes #1060


.github/workflows/ci.yml
```diff
@@ -21,6 +21,9 @@ jobs:
       - name: Formatting check
         run: cargo fmt --all -- --check
 
+      - name: Run unit tests
+        run: cargo test --all
```
Mark-Simulacrum, line 25: Can we move this to a separate builder (e.g., perhaps the windows builder)? I try to keep this one as fast as possible since it's the critical path to deployment.

We generally still block merging PRs on the Windows builder passing, and generally outage is not terrible for perf, so I think that this sort of optimization still makes sense.

PR #1069, opened 2021-10-11
Don't group testCases by benchmark/profile in comparison page

### LINKED ISSUE #1068: Sort by `% change` or signifiance factor in results page?
Currently, the results page sorts changes alphabetically by benchmark name:

image

However, I wonder if it would be better to sort by `% change` or significance factor, especially because `rust-timer`'s summary comments always point out the largest change, which is not necessary at the top of the results page.

rylev:
Fixes #1068 

This changes the comparison page in two major ways:
* Test results are no longer grouped by benchmark/profile. Each test result exists independent of each other. 
* The raw results (which are less important than the percent change and significance factor) are moved to the end of each row.

I don't love how this looks, but I think the overall behavior is an improvement. Any thoughts on how we can make this look better?

image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -423,7 +423,8 @@
             <table id="benches" class="compare">
                 <tbody>
                     <tr>
-                        <th>Benchmark Profile Scenario</th>
+                        <th>Benchmark & Profile</th>
+                        <th>Scenario</th>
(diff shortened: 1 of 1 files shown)


rylev:
Howw about this? 
image

PR #1087, opened 2021-11-02
Fetch self profile data from s3 on demand and parse artifact sizes

rylev:
co-authored with @michaelwoerister 

This is the start of functionality for showing artifact sizes in the detailed-query.html page. Instead of story the data in the database, this approach fetches and parsed the data on demand. If this approaches works, we may want to consider doing the same for the other self-profile data.

Unfortunately, it's not possible to test this code locally so it's likely horribly broken. Any tips or thoughts on how best to test this code and push it forward would be very helpful.

### DIFF
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -296,6 +296,7 @@ pub mod self_profile {
     pub struct SelfProfile {
         pub totals: QueryData,
         pub query_data: Vec<QueryData>,
+        pub artifact_sizes: Option<Vec<ArtifactSize>>,
     }
 
     #[derive(Serialize, Deserialize, Clone, Debug)]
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Why can't you test locally? You might need to hardcode the self profile data urls for S3, but it should be possible to run most of the code as-is.

FWIW, as we continue investing in self profile rendering, I suspect we'll want a way to run the site against local data - but I would rather not block this on that larger effort.

Mark-Simulacrum:
I am a little worried about the latency here -- fetching and "parsing" the self-profile artifacts for S3 is currently not in a particularly hot path (needs a couple clicks), and is pretty slow -- easily in the ~hundreds of milliseconds range. We've (I've) not spent much time optimizing that though, so maybe there's easy wins to be had.

We can also consider bumping the size of the perf.rlo instance (currently we're at 0.5 vCPUs, IIRC), but it seems a little unfortunate to have to do that.

rylev:
> Why can't you test locally?

I need a `cid` from the `raw_self_profile` table in order to properly generate the url. `raw_self_profile` is empty in the local database and so it's not possible to find the right `cid` for the url. 

> I am a little worried about the latency here

Fetching a 6mb file from S3 and parsing it shouldn't be too slow. If it is, I'm fairly confident we'll be able to do this quickly, but there's only one way to find out 😉

Mark-Simulacrum:
Ah, right. Hm. I exported two recent aid's manually from the production db as a csv of the columns for raw_self_profile table (https://gist.github.com/Mark-Simulacrum/668c13233e1232535ccd24007cb06e97) -- maybe you can use that to test out?

PR #1089, opened 2021-11-04
fix(compare) spacing before tooltip

### LINKED ISSUE #1070: question mark overlaps with text
This overlaps in firefox 93
pic2

but not in chrome 94?
pic3

For example: https://perf.rust-lang.org/compare.html?start=15491d7b6be1065217e09bd735e4ecffbe9838ba&end=bc8ad24020a160e1acd7ac9f7671947dcc01264c&stat=max-rss

Resolution in both cases 1920x1080

https://github.com/rust-lang/rustc-perf/blob/e92928f818d0587de2080f6c30b2fe486140a605/site/static/compare.html#L429-L439

Perhaps this text without dedicated containers renders badly as table cells, idk.

homersimpsons:
Fixes #1070

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -427,7 +427,8 @@
                         <th>Scenario</th>
                         <th>% Change</th>
                         <th>
-                            Significance Factor<span class="tooltip">?
+                            Significance Factor
+                            <span class="tooltip">?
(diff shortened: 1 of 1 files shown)


homersimpsons:
For the record, after the change: 
image

PR #1091, opened 2021-11-04
feature(compare) Add option to hide raw data, and improve formatting

### LINKED ISSUE #998: Remove the absolute test result numbers from the comparison page
The comparison page provides two columns with the absolute numbers (in milliseconds?) of the two artifacts being compared. I never look at these numbers since it's only important what the difference between them is. Perhaps these can be removed and more useful data (e.g., the severity number proposed in #997) could be used instead. 

image

homersimpsons:
fixes #998

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -389,6 +389,22 @@
                 </div>
             </div>
         </fieldset>
+        <fieldset id="display">
+            <legend id="display-toggle" class="section-heading">Display<span id="display-toggle-indicator"></span>
+            </legend>
+            <div id="display-content" style="display: none;">
+                <div class="section">
+                    <div class="section-heading"><span>Display raw data</span>
+                        <span class="tooltip">?
+                            <span class="tooltiptext">
+                                Whether to display or not raw data columns.
+                            </span>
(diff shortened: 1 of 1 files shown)


homersimpsons:
image
image

Mark-Simulacrum:
It looks like the column headers are pretty poorly aligned - that might be something to tackle in a different PR, since it appears pre-existing, but seems worth fixing. Probably needs to be played with to make it look good.

homersimpsons:
Yes, the headers are left-aligned while the content are centered in the cell. I'm no expert in UI so I do not know which content should be aligned to which side ....

Mark-Simulacrum:
I would match the alignment of the values to start.

Mark-Simulacrum:
@homersimpsons I think it'd be helpful to move the toggle for displaying raw values into the Filter box -- that takes up less vertical space, and I don't think it's inherently a different "kind" of toggle -- at least not enough to warrant a separate box.

homersimpsons:
I can do that, the idea was that this checkbox do not change the displayed results but only the way they are displayed

PR #1093, opened 2021-11-08
fix(compare) Reduce content top offset: merge display into filter

homersimpsons:
Addresses https://github.com/rust-lang/rustc-perf/pull/1091#issuecomment-961331799

r? @Mark-Simulacrum 

```javascript
document.getElementById('content').offsetTop; // 404
```

I also opened #1094 with an alternative solution

PR #1094, opened 2021-11-08
fix(compare) Reduce display section: collapse by default

homersimpsons:
Addresses https://github.com/rust-lang/rustc-perf/pull/1091#issuecomment-961331799

r? @Mark-Simulacrum

```javascript
document.getElementById('content').offsetTop; // 420
```

This is an alternative to #1093


homersimpsons:
image

PR #1100, opened 2021-11-15
Pass parsed scenario name to downstream profile data source

rylev:
We were passing raw scenario names like "incr-patched: println" to s3 instead of parsing and passing it as a normalized id like "incr-patched-println".


Mark-Simulacrum:
We should also consider unifying the different pieces of code that do this in the future, so we avoid bugs like this.

PR #1122, opened 2021-12-15
Name bin target differently from lib

Mark-Simulacrum:
This avoids file collisions, which lead to spurious rebuilds and errors from
Cargo (recently added by rust-lang/cargo#10172).

I haven't been able to confirm this locally yet, possibly because the wrong
Cargo gets used (not sure), but this should hopefully fix the benchmark on perf.r-l.o.

### DIFF
--- a/collector/benchmarks/token-stream-stress/Cargo.toml
+++ b/collector/benchmarks/token-stream-stress/Cargo.toml
@@ -5,11 +5,15 @@ edition = "2018"
 publish = false
 
 [lib]
+# Documenting the library means that we try to rebuild it after src/main.rs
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
This avoids clobbering the output (good, I guess) but still runs into the same problem. Discussing on https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/document.20lib.20before.20bin.

Mark-Simulacrum:
https://github.com/rust-lang/cargo/pull/10201 should let us move forward here once a Cargo bump including it lands on rust-lang/rust.

Issue #1126, opened 2021-12-20
Per-version hashing information


Mark-Simulacrum:
Since the merge of https://github.com/rust-lang/rust/pull/89836, we are seeing high variance on the regression-31157-opt benchmark:

![image](https://user-images.githubusercontent.com/5047365/146846204-edcdba07-018b-4587-83c3-a9ce7406e0e2.png)

This is likely due to hashing being effectively randomized, affecting some hash table layout (I've not narrowed down where/what exactly).

I suspect this may be creating noise on other benchmarks, as well. Our significance algorithm should learn to cope with this in another ~20 commits (roughly 2-3 days) as our backwards window is for 50 commits, but it likely makes sense to try to avoid this problem at the root, and in general it'll make regression triage harder by increasing the baseline changes commit to commit even with small code deltas.

I am not sure yet whether a revert of rust-lang/rust#89836 makes sense, or whether we should use RUSTC_FORCE_INCR_COMP_ARTIFACT_HEADER on rustc-perf to reduce noise levels... in the meantime I'll set the env variable for perf and that'll hopefully eliminate noise in the meantime.

cc @jyn514

Mark-Simulacrum:
This is essentially fixed by https://github.com/rust-lang/rustc-perf/pull/1127.

![image](https://user-images.githubusercontent.com/5047365/147984855-80079475-dac1-4ee4-916c-5b2dae6a23b9.png)
![image](https://user-images.githubusercontent.com/5047365/147984882-1c3ae5fb-8273-47b3-b303-ea71a946f2c4.png)

PR #1127, opened 2021-12-20
Remove rustc version from stable crate IDs

Mark-Simulacrum:
Since rust-lang/rust#89836, rustc stable crate IDs include a hash of the
rustc version (including the git commit it's built from), which means
that hashmaps or other structures have different behavior when comparing
different rustc builds. This is bad for rustc-perf, as it means that
comparing two commits has a source of noise that makes it harder to know
what the actual change between two artifacts is.

cc https://github.com/rust-lang/rustc-perf/issues/1126

### DIFF
--- a/collector/src/main.rs
+++ b/collector/src/main.rs
@@ -321,7 +321,7 @@ fn bench(
 }
 
 fn check_measureme_installed() -> Result<(), String> {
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
CI is failing but that's because of https://github.com/rust-lang/rust/issues/92163, not this PR in particular.

PR #1146, opened 2022-01-13
Include the profile in a result's summary; more documentation for rust-timer commands

### LINKED ISSUE #1048: "full builds of benchmark" message should include profile
https://github.com/rust-lang/rust/pull/88679#issuecomment-932841079 says "Very large regression in instruction counts (up to 6.6% on full builds of regression-31157)". But the actual regressed build is "regression-31157-doc".

### LINKED ISSUE #1099: Please link the full rust-timer commands from the help page
https://perf.rust-lang.org/help.html says
> @rust-timer has more commands than just @rust-timer queue, but the queue command is the most used. 

It took me more time than I care to admit to find <https://github.com/rust-lang/rustc-perf/blob/b3b52c5e8f91170c0dbe541117236f1c09e31100/site/src/request_handlers/github.rs#L12> to know what those other ones are.  A link -- just to the code, if there aren't docs -- would be handy from the site.

lqd:
Fixes #1048 by changing the summary line of a benchmark result by adding its profile as well.

This matches the compare page's "Benchmark & Profile" column, so I think it's clearer in that sense (but maybe it can be put elsewhere in the summary sentence). But this change will also impact the automated weekly report: I therefore wonder if this is how you'd want it to be fixed ?

For example, [this summary](https://github.com/rust-lang/rust/pull/88679#issuecomment-932841079) would now mention that the regression was in a `doc` benchmark: from "Very large regression in instruction counts (up to 6.6% on `full` builds of `regression-31157`)" to "Very large regression in instruction counts (up to 6.6% on `full` builds of `regression-31157 doc`)".

And [this one](https://github.com/rust-lang/rust/pull/92534#issuecomment-1008392640) would go from "Large regression in instruction counts (up to 1.6% on `incr-unchanged` builds of `derive`)" to "Large regression in instruction counts (up to 1.6% on `incr-unchanged` builds of `derive check`)"

Resolves #1099 by adding a tiny bit more documentation about rust-timer commands on the Help page, and the link Scott suggests. 

<details>
<summary>Here's how it looks rendered</summary>

image

</details>

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -1000,8 +1000,8 @@ impl TestResultComparison {
         .unwrap();
         writeln!(
             summary,
-            " (up to {:.1}% on `{}` builds of `{}`)",
-            percent, self.scenario, self.benchmark
+            " (up to {:.1}% on `{}` builds of `{} {}`)",
+            percent, self.scenario, self.benchmark, self.profile
         )
         .unwrap();
     }
--- a/site/static/help.html
+++ b/site/static/help.html
@@ -54,8 +54,20 @@
                 <code>&lt;RUNS&gt;</code> times.
             </li>
         </ul>
-        <p><code>@rust-timer</code> has more commands than just <code>@rust-timer queue</code>, but the
-            <code>queue</code> command is the most used.
+        <p><code>@rust-timer</code> has
+            <a href="https://github.com/rust-lang/rustc-perf/blob/dcae222b6a7fbb4c9e21d95bc93313ebe5107c7c/site/src/request_handlers/github.rs#L12-L21">more commands</a>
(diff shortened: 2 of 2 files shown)


rylev (approved):
Looks good to me, but I'd like to see @Mark-Simulacrum's opinion especially on the changes to the help page.

Mark-Simulacrum:
Seems good to me. I think we may wish to consider linking the scenario and profile to documentation (e.g., https://github.com/rust-lang/rustc-perf/blob/master/docs/glossary.md) and the benchmark name to the relevant line in https://github.com/rust-lang/rustc-perf/blob/master/collector/benchmarks/README.md -- but that's a larger project.

FWIW, in the future, it seems nice to split this PR into two (for each part of the changes).

PR #1156, opened 2022-01-21
Improve the summary message both in GitHub and in the triage report

rylev:
This improves the summaries generated for both the perf bot on GitHub as well as the weekly triage report. 

Note: I've not tested this locally as I don't have a local database in a useable form anymore. 

### Perf Bot

**before**

**Summary**: This change led to large relevant regressions 😿 in compiler performance.
- Large regression in instruction counts (up to 2.2% on `incr-patched: reverse` builds of `regex check`)

**after**

**Summary**: This benchmark run shows 10 relevant regressions 😿 to instruction counts.
- Average relevant regression: 1.7%
- Largest relevant regression in instruction count: 2.2% on `incr-patched: reverse` builds of `regex check`

### Triage Report

**before**

- Very large regression in [instruction counts](https://perf.rust-lang.org/compare.html?start=22e491ac7ed454d34669151a8b6464cb643c9b41&end=f312a5e610d47601e9a3da828002f5e1ffeb272a&stat=instructions:u) (up to 8.2% on `full` builds of `keccak check`)

**after** 

- Largest relevant regression in [instruction counts](https://perf.rust-lang.org/compare.html?start=22e491ac7ed454d34669151a8b6464cb643c9b41&end=f312a5e610d47601e9a3da828002f5e1ffeb272a&stat=instructions:u): 8.2% on `full` builds of `keccak check`

## Other Changes

Additionally, this PR changes the message slightly when there are no "relevant" changes (and thus nothing needs to be done) but there are still some (small) statistically significant changes. The following is added in the perf bot message.

"N results were found to be statistically significant but too small to be relevant."

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -163,19 +163,36 @@ async fn populate_report(
     }
 }
 
+/// A summary of a given comparison
+///
+/// This summary only includes changes that are significant and relevant (as determined by a changes magnitude).
 pub struct ComparisonSummary {
     /// Significant comparisons of magnitude small and above
     /// and ordered by magnitude from largest to smallest
     comparisons: Vec<TestResultComparison>,
+    /// The cached number of comparisons that are improvements
+    num_improvements: usize,
+    /// The cached number of comparisons that are regressions
+    num_regressions: usize,
 }
 
 impl ComparisonSummary {
     pub fn summarize_comparison(comparison: &Comparison) -> Option<ComparisonSummary> {
+        let mut num_improvements = 0;
+        let mut num_regressions = 0;
+
         let mut comparisons = comparison
             .statistics
             .iter()
             .filter(|c| c.is_significant())
             .filter(|c| c.magnitude().is_small_or_above())
+            .inspect(|c| {
+                if c.is_improvement() {
+                    num_improvements += 1;
+                } else {
+                    num_regressions += 1
+                }
+            })
             .cloned()
             .collect::<Vec<_>>();
         // Skip empty commits, sometimes happens if there's a compiler bug or so.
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum (approved):

site/src/comparison.rs
```diff
@@ -277,6 +292,35 @@ impl ComparisonSummary {
...
+        for r in changes {
+            sum += r.relative_change();
+            count += 1;
+        }
+
+        (sum / count as f64) * 100.0
+    }
+
+    fn improvements(&self) -> impl Iterator<Item = &TestResultComparison> {
+        self.comparisons.iter().filter(|c| c.is_regression())
```
Mark-Simulacrum, line 317: Er, typo? is_improvement()?

Mark-Simulacrum:
Feel free to merge with nit fixed

nnethercote:
Oh! I've been thinking about this stuff, and have been meaning to make some changes myself.

- I would like the message to say how many relevant improvements/unchangeds/regressions there were.
- I would like finer classification of "mixed results" cases. Sometimes they are genuinely mixed, e.g. half improvements and half regressions. But sometimes they are like 90% improvements and 10% regressions, or vice versa. Adding a "mostly positive" or "mostly negative" classification might help here? Though perhaps the previous point would reduce the need for this.

What do you think?

rylev:
> * I would like the message to say how many relevant improvements/unchangeds/regressions there were.

I believe this is what I already have (it shows the number of improvements and regressions) except it doesn't say the number of "unchanged". Do you want the number of unchanged to be the number of test cases where there was no significant change - this is essentially just all other test cases that aren't improvements or regressions.

> * I would like finer classification of "mixed results" cases. Sometimes they are genuinely mixed, e.g. half improvements and half regressions. But sometimes they are like 90% improvements and 10% regressions, or vice versa. Adding a "mostly positive" or "mostly negative" classification might help here? Though perhaps the previous point would reduce the need for this.

We could do this, but it might be useful to first see how these changes look in practice before figuring out the line between "genuinely mixed" and "mixed but leaning one way predominantly". What do you think?

camelid:
Maybe "important" would be a better word than "relevant"? The distinction between "significant" and "relevant" seems a bit fuzzy, but "significant" and "important" seem better separated.

nnethercote:
> Do you want the number of unchanged to be the number of test cases where there was no significant change - this is essentially just all other test cases that aren't improvements or regressions.

I want this in the summary, basically:

Screen Shot 2022-01-26 at 10 02 23 am

PR #1164, opened 2022-01-31
Fix try grandparent commit from ordering incorrectly

### LINKED ISSUE #1163: Missing start commit for 90891
https://github.com/rust-lang/rust/pull/90891#issuecomment-1025606783

Mark-Simulacrum:
This is a larger PR than I was hoping for, but this area of the code is actually somewhat tested (and this PR adds another test). In some ways I think the new code is also cleaner, which seems like a nice win.

Happy to receive suggestions on how to further clean it up.

Resolves #1163

### DIFF
--- a/site/src/load.rs
+++ b/site/src/load.rs
@@ -28,11 +28,37 @@ pub enum MissingReason {
     },
     Try {
         pr: u32,
+        parent_sha: String,
         include: Option<String>,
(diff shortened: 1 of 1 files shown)


nnethercote (approved):
I don't know this code at all but it seems reasonable, and as you say the test gives confidence.

PR #167, opened 2017-11-01
Update index.html to include a "fork me on GitHub" ribbon

furpuddle:
issue #141 - added "Fork me on GitHub" ribbon

PR #1168, opened 2022-02-08
Clip 'Fork me on GitHub' buttons

dtolnay:
The site is otherwise unusable on some mobile browsers because the clickable area of the 'Fork me on GitHub' button from https://github.com/rust-lang/rustc-perf/pull/167 completely covers the `instructions:u` dropdown.

<p align="center"></p>

Clickable area before:

Screenshot from 2022-02-07 18-01-39

Clickable area after:

Screenshot from 2022-02-07 18-01-49

Switching to https://github.com/simonwhitaker/github-fork-ribbon-css would be a good alternative fix.

### DIFF
--- a/site/static/bootstrap.html
+++ b/site/static/bootstrap.html
@@ -76,7 +76,7 @@
     <div id="totalChart"></div>
     <div id="as-of"></div>
     <a href="https://github.com/rust-lang-nursery/rustc-perf">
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
This seems ok to me - some of the pages just have this on the bottom and some omit it entirely I think, which might be a better approach. This seems like a good start though.

Mark-Simulacrum (approved):

PR #1175, opened 2022-02-17
Avoid sort failure due to NaN comparison

Mark-Simulacrum:
Noticed while looking at some bootstrap results that were incorrectly sorted -- this should ensure we always have a good sort order.

(Fixing the NaN presence is a separate problem -- I believe caused by division by zero, but not sure yet).


nnethercote (approved):
This confused me briefly until I realized I was reading JS code rather than Rust code :)

PR #1181, opened 2022-02-24
Benchmark categorization

Kobzol:
This PR adds a `category` attribute to each benchmark, which should serve to quickly filter benchmarks that fall into different categories (mostly real-world vs stress-test benchmarks). The category is entered in `perf-config.json` (default is `secondary`), then it's put into the DB and threaded from the DB to the compare page. The compare page now shows two tables (for primary and secondary benchmarks) and it also has a filter so that you can show e.g. only the primary ones (the filter is mainly useful for the summary, so that you can see e.g. only the percentual changes for real-world crates).

benchmarks

Couple of comments:
1) I modified the DB to store the benchmark category, instead of hardcoding it in the dashboard or the `site` crate. I thus had to work around the `supports-stable` column in the DB, which looks quite complicated (its handling is complicated). And the attribute is not even used or read anywhere, the crates that support stable are hardcoded (https://github.com/rust-lang/rustc-perf/blob/9a9d1a807a1a5edfdde7000283e646f48b8d4fad/site/src/request_handlers/dashboard.rs#L88). I suggest to either remove this attribute from DB or start using it (that's for another PR though).
2) Doing large changes to the compare page is quite cumbersome. It could get some refactoring, and switching to Vue 3 would also be nice (for example in this PR I had to put the whole benchmark table into the component, instead of just the rows, because Vue 2 doesn't support components that render as multiple DIV elements). This change was IMO on the verge of what can still be changed until the page's code gets messed up.
3) The categorization of crates into primary/secondary was done according to my taste, it should be discussed.
4) Primary/secondary naming seems a bit opaque, maybe real-world/artificial or real-world/stress-test would be better.
5) I'm not sure how to pass the category state from SQLite to Postgre in the import binary, as that code doesn't even touch the `benchmarks` table. It will probably require some larger changes (or some hack :) ).


collector/src/execute.rs
```diff
@@ -110,6 +110,19 @@ fn default_runs() -> usize {
     3
 }
 
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum BenchmarkCategory {
```
nnethercote, line 115: Nit: just call it `Category`/`category` throughout, it's shorter and no less meaningful given that there aren't other kinds of category :)

collector/src/execute.rs
```diff
@@ -1239,6 +1256,13 @@ impl Benchmark {
         self.config.supports_stable
     }
 
+    pub fn category_name(&self) -> &str {
```
nnethercote, line 1259: Make this a method on `Category` itself? (Possibly even `impl Display`) Then below where you currently have `"primary"` you could instead do `(Category::Primary).to_string()`

collector/benchmarks/native-tls-0.1.5/perf-config.json
```diff
@@ -1,3 +1,4 @@
 {
-    "touch_file": "src/lib.rs"
+    "touch_file": "src/lib.rs",
+    "category": "primary"
```
nnethercote, line 3: `native-tls-0.1.5` isn't actually a benchmark. For some reason it's in here because `tokio-webpush-simple` needs it, I think. So I'm not sure if it should even have a `perf-config.json` file at all? And perhaps we could move it elsewhere, to avoid this kind of confusion. (Not as part of this PR, though.)

collector/benchmarks/rust-mozjs/perf-config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "category": "primary"
```
nnethercote, line 2: Similarly, `rust-mozjs` isn't a benchmark, it's needed for `style-servo`.

nnethercote:
> Primary/secondary naming seems a bit opaque, maybe real-world/artificial or real-world/stress-test would be better.

When thinking about this change I originally had a real-world/artifical distinction, but after some thought I now have a strong preference for primary/secondary because it's more flexible. E.g. there might be a real-world benchmark (i.e. from crates.io) that stresses the compiler in an interesting but obscure way, and we could put it in the secondary category. Likewise, `helloworld` isn't exactly real-world, but I do want to have it in the primary category because it's a good measure of the shortest possible compile time, which is interesting. It also avoids arguments about the exact meaning of "real-world" :)

Having said that, most of the crates in the primary category will be "real-world", and most of the crates in the secondary category will be artificial in some way, and we can certainly mention that in the [glossary](https://github.com/rust-lang/rustc-perf/blob/master/docs/glossary.md). This PR should also update the [benchmark description list](https://github.com/rust-lang/rustc-perf/blob/master/collector/benchmarks/README.md) should be updated. Currently it has three categories ("Real programs that are important", "Real programs that stress the compiler", "Artificial stress tests"), it should just be split into primary/secondary.

nnethercote:
Here's the draft primary/secondary split that I came up. Jakub's primary choices are marked with `*`, there's a lot of overlap but a few differences: `inflate`, `keccak`, `tokio-webpush-simple`, `wg-grammar`.

Primary
-------
*cargo
*clap-rs
*cranelift-codegen
*diesel
*encoding
*futures
*helloworld
*html5ever
*hyper-2
inflate
keccak
*piston-image
*regex
*ripgrep
*serde
*stm32f4
*style-servo
*syn
tokio-webpush-simple
*ucd
*unicode_normalization
*webrender
*webrender-wrench

Secondary
---------
await-call-tree
coercions
ctfe-stress-4
deeply-nested
deeply-nested-async
deeply-nested-closures
deep-vector
derive
externs
issue-46449
issue-58319
issue-88862
many-assoc-items
match-stress-enum
match-stress-exhaustive_patterns
projection-caching
regression-31157
token-stream-stress
tuple-stress
unify-linearly
unused-warnings
wf-projection-stress-65510
*wg-grammar

nnethercote:
Finally, I'll say that the screen capture above looks great, very much what I was imagining and hoping for, so thank you for doing this! I have commented on some parts of the code, but I am no expert on the database stuff or the webdev stuff, so I will leave that so @Mark-Simulacrum and @rylev.

Kobzol:
I rebased over `master` and addressed the comments. I agree with your revised categorization, the table with the asterisks was incredibly helpful, thanks!

collector/benchmarks/README.md
```diff
@@ -7,20 +7,33 @@ The suite changes over time. Sometimes the code for a benchmark is updated, in
...
 These are real programs that are important in some way, and worth tracking.
+They mostly consist of real-world crates.
 
 - **cargo**: The Rust package manager. An important program in the Rust
   ecosystem.
 - **clap-rs**: A command line argument parser. A crate used by many Rust
   programs.
 - **cranelift-codegen**: The largest crate from a code generator. Used by
   Firefox.
+- **diesel**: A type save SQL query builder. Utilizes the type system to
```
nnethercote, line 23:
```suggestion
- **diesel**: A type safe SQL query builder. Utilizes the type system to
```
?

collector/src/main.rs
```diff
@@ -252,6 +255,7 @@ fn bench(
         measure_and_record(
             &BenchmarkName("rustc".to_string()),
             /* supports_stable */ false,
+            "primary",
```
nnethercote, line 258: Can this be `Category::Primary.to_string()`?

nnethercote:
One presentation nitpick: the "Primary benchmarks" and "Secondary benchmarks" could be a big bigger, bolder, and with a bit more space around them. Maybe even centred above the table.

Kobzol:
image

rylev:
@Kobzol is this PR ready for review - I marked as such without thinking because it seems like it had already been reviewed by @nnethercote, but I should have asked first.

Overall, things look good (depending on your answer above, I will go through the PR and comment on some nits I have). I'm unsure though if we should be keeping this info in the database. Changing the database is quite difficult and risky, and I'm not sure what we get from it. Thoughts?

Kobzol:
It is, I just didn't want to unmark it as draft because there was still one TODO (import from SQLite into Postgre), so it wouldn't be good if it was merged with this TODO.

Regarding the database, I discussed this with @Mark-Simulacrum and our agreement was that the database solution is more general, because then the `site` doesn't need access to the benchmarks directory and can fully work with the DB. That being said, currently the `supports-stable` attribute is stored into the DB, but then it's still being hardcoded in the source code.

Hardcoding the values seems a bit inflexible. A solution that would be rather flexible but didn't require info in DB would be if the `site` crate would load the benchmarks (and information about stable support and category) from the `benchmarks` directory upon startup. But of course then it would depend on the directory being up-to-date and accessible.

nnethercote:
I did a partial review but [as mentioned above](https://github.com/rust-lang/rustc-perf/pull/1181#issuecomment-1050340021) I'm no expert on the database and website side of things, so others should review those changes.

collector/src/execute.rs
```diff
@@ -130,6 +151,9 @@ struct BenchmarkConfig {
     /// directory that `Cargo.toml` is in.
     #[serde(default)]
     touch_file: Option<String>,
+
+    #[serde(default)]
```
Mark-Simulacrum, line 155: I think we should require a category for all benchmarks, rather than defaulting to secondary.

collector/src/main.rs
```diff
@@ -188,6 +189,7 @@ fn bench(
     let mut measure_and_record =
         |benchmark_name: &BenchmarkName,
          benchmark_supports_stable: bool,
+         benchmark_category: &str,
```
Mark-Simulacrum, line 192: This is being passed as &str, but it seems like it should be passed as the enum -- that probably means moving the enum definition to database/src/lib.rs?

database/src/bin/import-sqlite.rs
```diff
@@ -41,7 +41,7 @@ async fn main() {
         for &(benchmark, profile, scenario, metric) in sqlite_idx.all_statistic_descriptions() {
             if benchmarks.insert(benchmark) {
                 postgres_conn
-                    .record_benchmark(benchmark.as_str(), None)
+                    .record_benchmark(benchmark.as_str(), None, "") // TODO
```
Mark-Simulacrum, line 44: I think this probably wants to be a meaningful string -- secondary.

Mark-Simulacrum:
I think this largely makes sense; I am a little worried that the secondary benchmarks might be hidden below the fold if there's a lot of primary benchmarks. Maybe that's ok, but I wonder if we should try to do something to explicitly show them -- particularly when there's regressions there.

One option might be to put the summary (+ neutral -) stats, with avg, on each of the 'primary' or 'secondary' headers, and then fold them if they get too long? Not sure how I feel about that though.

collector/src/main.rs
Kobzol: Well, it was my intention not to put this into the database crate. In the database it will just be a string, so that's the interface that I wanted to use for it. This way we can change the representation in the collector and it won't affect the database.

database/src/bin/import-sqlite.rs
Kobzol: Well mainly this should be copied from SQLite to Postgre, right? Is this import being used anywhere, is it important?

collector/src/execute.rs
Kobzol: Ok, sounds reasonable.

collector/src/main.rs
Mark-Simulacrum: I would prefer to intentionally update the database -- yes, we're storing it as a string under the hood, but all Rust code should largely deal with the enum and convert to it at boundary layers with disk (i.e., benchmarks directory) or db (postgres/sqlite).

Kobzol: The string from the DB is passed verbatim to the frontend though, it's not being handled in Rust at all at the database/site crate. But I agree that it would be nicer, I can change it.

Kobzol:
Regarding the implementation - do we have an agreement whether we want to store the category in DB? Or should it be loaded from the benchmark directory (is the directory always available for the `site` crate?).

Mark-Simulacrum:
Yes, I think we should store it in the DB -- it's much easier to migrate away from that if we see a need than vice-versa.

Kobzol:
Moved Category to DB and did something with the SQLite import (based on `git blame` it should be used to import local data to the perf server). It seems like changing the database is quite tricky and brittle, some ORM sure could make this easier.

nnethercote:
After some thought, I think `keccak` should be a secondary benchmark. @Kobzol, can you update accordingly? Thanks.

Kobzol:
Switched `keccak` to secondary category, rebased on master and squashed and added support for categories into the `download` command.

Kobzol:
Finally CI is green! This is ready for another round of review.

collector/src/main.rs
```diff
@@ -1170,12 +1180,27 @@ fn main_result() -> anyhow::Result<i32> {
...
 }
 
+fn add_perf_config(directory: &PathBuf, category: Category) {
+    let data = serde_json::json!({
+        "category": category.to_string()
+    });
+
+    let mut file = File::create(directory.join("perf-config.json"))
+        .expect("Could not create perf-config.json file");
+    serde_json::to_writer_pretty(&mut file, &data).expect("Could not write perf-config.json file");
```
Mark-Simulacrum, line 1200: nit: generally File wants to be wrapped in a BufWriter or similar, to avoid many tiny write syscalls.

Mark-Simulacrum:
Very small nit, and then let's move ahead with this!

collector/src/main.rs
Kobzol: I was wondering whether you or someone else would find this fishy :laughing: I'll wrap it.

Kobzol:
Added `BufWriter` and also [this](https://github.com/rust-lang/rustc-perf/pull/1181/files#diff-3f6b7d26c2c059b7950aef4625d5e193dec9345c5f12fe6f8fdede2ebc49405dR142), because after moving `Category` to DB and getting it in a strongly-typed form from the DB, we need to convert it to a string, so that JavaScript doesn't have to deal with Serde serialization of enums.

Mark-Simulacrum:
Hm, ok. I would have expected serde to just serialize the enum to a string value which should be pretty easy for JS to deal with, but maybe I'm missing something?

Kobzol:
If it were serde JSON, then yes. However, msgpack is used here, which currently serializes the enum as `"category": {"1": null}`, which is not really usable in Javascript.

Mark-Simulacrum (approved):

PR #1190, opened 2022-03-04
Fix profiling CI

Kobzol:
Without `apt update` there was a missing package.


Kobzol:
CC @nnethercote this should fix CI.

nnethercote (approved):
Thanks!

PR #1194, opened 2022-03-09
Compare site: cache master-branch Rust commits to improve performance

miwig:
The compare site currently gets the [Bors commit list](https://triage.rust-lang.org/bors-commit-list) over the network for every single request. On my machine, this means a time-to-first-byte of 1.4 seconds (400ms of which is spent parsing JSON) even for comparing just a single benchmark! Caching the parsed result cuts that down to a couple ms. Of course I expect it to matter less in the prod system, but there could still be some nice performance gains here, making browsing the site much more pleasant. 

Some questions where a review would be appreciated:

- Is the general idea ok like this? What changes still need to be made?
- What's the maximum acceptable stale response? I've chosen one minute here.
- Maybe stale responses/premature refreshes can be avoided altogether? I've set the cache to refresh when [perf/onpush](https://perf.rust-lang.org/perf/onpush) is called. Does this happen automatically somewhere? Ideally we would just load the commit data once when it changes.

### DIFF
--- a/site/src/load.rs
+++ b/site/src/load.rs
@@ -210,7 +210,7 @@ impl SiteCtxt {
         let conn = self.conn().await;
         let (queued_pr_commits, in_progress_artifacts) =
             futures::join!(conn.queued_commits(), conn.in_progress_artifacts());
-        let master_commits = &self.master_commits.load().commits;
+        let master_commits = &self.get_master_commits().commits;
 
         let index = self.index.load();
         let all_commits = index
--- a/site/src/load.rs
+++ b/site/src/load.rs
@@ -208,15 +208,9 @@ impl SiteCtxt {
     /// Returns the not yet tested commits
     pub async fn missing_commits(&self) -> Vec<(Commit, MissingReason)> {
         let conn = self.conn().await;
-        let (master_commits, queued_pr_commits, in_progress_artifacts) = futures::join!(
-            collector::master_commits(),
-            conn.queued_commits(),
-            conn.in_progress_artifacts()
-        );
-        let master_commits = master_commits
-            .map_err(|e| anyhow::anyhow!("{:?}", e))
-            .context("getting master commit list")
-            .unwrap();
+        let (queued_pr_commits, in_progress_artifacts) =
+            futures::join!(conn.queued_commits(), conn.in_progress_artifacts());
+        let master_commits = &self.master_commits.load().commits;
 
         let index = self.index.load();
         let all_commits = index
@@ -226,14 +220,14 @@ impl SiteCtxt {
             .collect::<HashSet<_>>();
 
         calculate_missing(
-            master_commits,
+            master_commits.clone(),
             queued_pr_commits,
             in_progress_artifacts,
             all_commits,
         )
     }
 
-    /// Get cached master-branch Rust commits.
+    /// Get cached master-branch Rust commits.  
     /// Returns cached results immediately, but if the cached value is older than one minute,
     /// updates in a background task for next time.
     pub fn get_master_commits(&self) -> Guard<Arc<MasterCommitCache>> {
--- a/site/src/load.rs
+++ b/site/src/load.rs
@@ -1,4 +1,4 @@
-use arc_swap::ArcSwap;
+use arc_swap::{ArcSwap, Guard};
 use std::collections::HashSet;
 use std::fs;
 use std::ops::RangeInclusive;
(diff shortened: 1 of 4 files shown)


site/src/load.rs
```diff
@@ -199,6 +222,16 @@ impl SiteCtxt {
             all_commits,
         )
     }
+
+    /// Download master-branch Rust commits if the cached value is older than one minute
+    pub async fn update_master_commits(&self) -> anyhow::Result<()> {
```
Mark-Simulacrum, line 227: I would prefer that we make this method return the cached result, rather than needing to call update() and then load the result from the field.

site/src/load.rs
```diff
@@ -199,6 +222,16 @@ impl SiteCtxt {
...
+        let commits = self.master_commits.load();
+        if commits.updated.elapsed() > std::time::Duration::from_secs(60) {
```
Mark-Simulacrum, line 229: Ideally we'd probably return results immediately always (if available) and then start a fetch in the background if it's time to update - stale results don't really hurt us much. It might be annoying to make sure there's only one background fetch running though, not sure.

Otherwise I suspect this wouldn't help much since many requests happen more than 60 seconds apart, I'd guess

Mark-Simulacrum (commented):
Overall seems good - we don't have (but probably should add) metrics for performance on prod, so it's not known how much of a benefit this could be, but it seems like a good idea regardless.

site/src/load.rs
miwig: I have implemented this. I don't think we need to bother making sure there's only one background fetch at a time. Worst case if N people request within ~half a second, we fetch N times, as we would without caching at all.

site/src/load.rs
```diff
@@ -127,7 +135,7 @@ pub struct SiteCtxt {
     /// Index of various common queries
     pub index: ArcSwap<crate::db::Index>,
     /// Cached master-branch Rust commits
-    pub master_commits: ArcSwap<MasterCommitCache>,
+    pub master_commits: Arc<ArcSwap<MasterCommitCache>>, // outer Arc enables mutation in background task
```
miwig, line 138: This was necessary to be able to access the cache in a background task. I think it makes sense, but please let me know if there's a better way.

miwig (commented):
Please let me know if this is ok and if you want it squashed before merge.

site/src/load.rs
```diff
@@ -208,15 +208,9 @@ impl SiteCtxt {
...
-            conn.queued_commits(),
-            conn.in_progress_artifacts()
-        );
-        let master_commits = master_commits
-            .map_err(|e| anyhow::anyhow!("{:?}", e))
-            .context("getting master commit list")
-            .unwrap();
+        let (queued_pr_commits, in_progress_artifacts) =
+            futures::join!(conn.queued_commits(), conn.in_progress_artifacts());
+        let master_commits = &self.master_commits.load().commits;
```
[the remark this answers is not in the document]
Mark-Simulacrum: Hm. I would expect this to call get_master_commits as well -- is there a reason it doesn't do that?

site/src/load.rs
```diff
@@ -160,9 +188,14 @@ impl SiteCtxt {
             }
         };
 
+        let master_commits = MasterCommitCache::download()
+            .await
+            .unwrap_or(MasterCommitCache::empty()); // still run the site if we can't get the commits right now
```
Mark-Simulacrum, line 193: This should probably be a regular unwrap() -- I'd rather not change the behavior we had before in terms of the *initial* fetch of commits. In other words, stale commits are probably fine (bounded by ~60 seconds), but we should not have just non-existent data.

site/src/server.rs
```diff
@@ -277,6 +277,8 @@ impl Server {
         // Refresh the landing page
         ctxt.landing_page.store(Arc::new(None));
 
+        let _ = ctxt.get_master_commits();
```
Mark-Simulacrum, line 280: I don't think this actually makes sense: it basically cannot help us get less stale data, since if things are cached they will still be cached despite this call. If the commits aren't already cached (i.e., more than 60 seconds since the previous call), then this would mean that we may refresh too early -- before triagebot updates the master commit list -- meaning that we have *delayed* getting the new commits for a full minute.

So I think we should just delete this line.

Mark-Simulacrum:
Looks like GitHub duplicated some of my comments across replies and new comments, but hopefully they make sense -- let me know if not. I think just one comment left on removing the empty constructor for the master commits data, and then we can go from there.

site/src/load.rs
miwig: Well, before caching, the failure mode was an error when accessing the compare page, the site wouldn't outright die on startup, but if that is preferred I can make the change.

miwig:
@Mark-Simulacrum I've addressed your comments, let me know if there's more to be done.

Mark-Simulacrum (approved):

PR #1197, opened 2022-03-09
Update Rust before building a freshly pulled collector

Mark-Simulacrum:
This ensures that we can seamlessly start using new features, and as such an
updated collector binary. I've already bumped Rust manually but I believe this was
the cause of the failure to update crates to the "primary" category, as the collector
wasn't building.

### DIFF
--- a/collector/collect.sh
+++ b/collector/collect.sh
@@ -32,4 +32,7 @@ while : ; do
 
         target/release/collector bench_next $SITE_URL --self-profile --bench-rustc --db $DATABASE;
         echo finished run at `date`;
+
+        # Wait a little bit before the next run starts.
+        sleep 120
 done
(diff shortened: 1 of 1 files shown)


nnethercote (approved):

collector/collect.sh
```diff
@@ -13,6 +13,10 @@ while : ; do
         # Update and rebuild the collector.
         git pull
         git reset --hard @{upstream}
+
+        # Make sure we have a recent build, so that we can successfully build
+        # the collector.
+        rustup update
```
Kobzol, line 19: I wonder if this is "safe", since if the update happens e.g. everyday, we can get an invalid nightly that breaks something? Is nightly even required for the collector, can't we just build it with stable and update that from time to time?

Mark-Simulacrum: Nightly shouldn't break, and if the build still fails after this change we should stay up, just with a stale version (like we did before I ran the update). Moving to stable seems ok; IIRC there was some reason for using nightly here but it might not apply anymore.

PR #1201, opened 2022-03-12
Site: compress responses

### LINKED ISSUE #1187: Enable gzip on responses
The site is very slow, and much of that (>1s) seems to be network transfers, with the graphs page transferring 1.4MB and the comparison page 600kB. ce59f44351bef482a23c61d5fb1339102b71e8f3 removed gzip on responses with the comment "In prod, nginx can do this for us", but as far as I can tell it was never enabled or got disabled at some point. #742 and #745 have also previously suggested enabling gzip (and ditching MessagePack), but as far as I can tell, regardless of whether MessagePack is used, [just enabling gzip should be a compression win](https://www.peterbe.com/plog/msgpack-vs-json-with-gzip).

miwig:
Closes #1187. I've chosen [Brotli](https://crates.io/crates/brotli) as the compression method, which is [supported in all major browsers](https://caniuse.com/?search=brotli) and should be [better and faster than gzip](https://expeditedsecurity.com/blog/nginx-brotli/). 

For now, compression is applied to all POST request responses (including the compare page), and to the graph/graphs requests.

It should be a performance win, especially on slower connections, but it's hard to tell how much without #1194 running in prod. It would also be great to have a recent dump of the database for local testing.


site/src/server.rs
```diff
@@ -313,6 +317,19 @@ async fn serve_req(server: Server, req: Request) -> Result<Response, ServerError
...
     let path = path.as_str();
 
+    let allow_compression = req
+        .headers()
+        .get(hyper::header::ACCEPT_ENCODING)
+        .map_or(false, |e| e.to_str().unwrap().contains("br"));
+
+    let compression = if allow_compression {
+        let mut brotli = BrotliEncoderParams::default();
+        brotli.quality = 4;
```
Mark-Simulacrum, line 327: Could this get a comment of some kind for why 4 was chosen?

site/src/server.rs
```diff
@@ -615,6 +643,35 @@ where
...
+            response.body(hyper::Body::from(compressed)).unwrap()
+        }
+    }
+}
+
+fn compress_bytes(bytes: &Vec<u8>, brotli_params: &BrotliEncoderParams) -> Vec<u8> {
+    let mut compressed = Vec::with_capacity(bytes.len());
+    BrotliCompress(
+        &mut Cursor::new(&bytes),
+        &mut Cursor::new(&mut compressed),
```
Mark-Simulacrum, line 668: I would not expect the cursors to be necessary here -- does `BrotliCompress(&bytes[..], &mut compressed, brotli_params)` not work? Can you paste the error message?

site/src/server.rs
```diff
@@ -615,6 +643,35 @@ where
...
+            let response = response.header(
+                hyper::header::CONTENT_ENCODING,
+                hyper::header::HeaderValue::from_static("br"),
+            );
+            response.body(hyper::Body::from(compressed)).unwrap()
+        }
+    }
+}
+
+fn compress_bytes(bytes: &Vec<u8>, brotli_params: &BrotliEncoderParams) -> Vec<u8> {
```
Mark-Simulacrum, line 664:
```suggestion
fn compress_bytes(bytes: &[u8], brotli_params: &BrotliEncoderParams) -> Vec<u8> {
```

Typically `&Vec<T>` is taken as `&[T]` instead.

miwig:
I've implemented your suggestions. Let me know if there's more :)

Mark-Simulacrum (approved):
It might be worth migrating to plain JSON for all our responses, now that we're compressing them I imagine that might be not that much of a size/performance hit, particularly in browsers where JSON loading is pretty well optimized (compared to the JS library we're using).

Issue #1184, opened 2022-02-25
Improve perf.rlo frontend experience


rylev:
In https://github.com/rust-lang/rustc-perf/pull/1181 @Kobzol mentions that we are at the limit of even what Vue2 can offer us on the compare page. This is despite the fact that most other pages on the compare site are built by concatenating HTML strings which can be very hard to understand and extend. 

I propose we do the following to improve our ability to iterate on the frontend experience:
* Update to Vue3
* Componentize common elements as Vue components
* Introduce the usage of View on all pages that have more than just basic dynamic content (e.g., perhaps everything but the help page)

Kobzol:
The frontend now uses Vue 3, `npm`, `parcel` and `TypeScript`, and the whole frontend has been refactored and split into components. I think that this goal has been fully achieved.

PR #1213, opened 2022-03-18
Upgrade to Vue3

rylev:
As a first step in #1184 this PR upgrades to Vue3. Largely, the upgrade was seamless other than needing to move away from using global functions on the `Vue` object in favor of the localized app object created with `createApp`. 

Additionally, I moved the CSS and JS to separate files since the comparison.html is quite large, and moving to separate files gives us additional opportunity to share code.


site/static/compare.css
```diff
@@ -0,0 +1,276 @@
...
+    .summary-wide {
+        width: 20%;
+    }
+}
+.category-title {
+    font-weight: bold;
+    font-size: 1.2em;
+    margin: 10px;
+    text-align: center;
+}
```
bjorn3, line 276: nit: missing trailing newline

rylev:
Closing in favor of #1255

PR #1215, opened 2022-03-19
Cache static responses

miwig:
Allowing the browser to cache responses to requests for static content should save another 200-800ms per page load, depending on connection speed.  

In the first commit, the current git commit id is compiled into the binary and used as an `ETag` header in responses. The browser will cache responses for one minute, after which new requests for the same file will include an `If-None-Match` header with the ETag, to which the server then replies `304 Not Modified` if the ETag matches, instead of resending the file.

On a fast connection, this doesn't save much time though (since the files are small anyway and request time is dominated by time-to-first-byte), which is why the second commit enables non-html content to be cached for longer by introducing versioned file names for cache-busting. To accomplish this, static content is served from memory instead of from disk, and occurrences of the string `{rustc_perf_version}` are replaced with the git commit id.

This could of course still be smarter, i.e. by versioning links with the file hash instead of the git commit id, but I've tried to keep complexity as low as possible while still hopefully gaining a nice speed boost. Please let me know what you think.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1880,6 +1880,7 @@ dependencies = [
  "tokio",
  "toml",
  "url",
+ "uuid",
 ]
 
 [[package]]
@@ -2201,6 +2202,15 @@ dependencies = [
  "idna",
  "matches",
  "percent-encoding",
+]
+
+[[package]]
+name = "uuid"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bc5cf98d8186244414c848017f0e2676b3fcb46807f6668a97dfe67359a3c4b7"
+dependencies = [
+ "getrandom",
 ]
 
 [[package]]
--- a/site/Cargo.toml
+++ b/site/Cargo.toml
@@ -44,6 +44,7 @@ tar = "0.4"
 inferno = { version="0.10", default-features = false }
 mime = "0.3"
 prometheus = "0.12"
+uuid = { version = "0.8.2", features = ["v4"] }
 
 [target.'cfg(unix)'.dependencies]
 jemallocator = "0.3"
--- a/site/src/server.rs
+++ b/site/src/server.rs
@@ -3,14 +3,15 @@ use brotli::BrotliCompress;
 use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::path::Path;
+use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering as AtomicOrdering};
 use std::sync::Arc;
 use std::time::Instant;
 use std::{fmt, fs, str};
 
 use futures::{future::FutureExt, stream::StreamExt};
-use headers::CacheControl;
-use headers::{Authorization, ContentType, Header};
+use headers::{Authorization, CacheControl, ContentType, ETag, Header, HeaderMapExt, IfNoneMatch};
+use http::header::CACHE_CONTROL;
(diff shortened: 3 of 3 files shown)


miwig:
Ok, so CI is failing because it's not building from the git repository, but that could be fixed by using something else instead of git_version!() as the version tag, e.g. an env-var set by a build script.

rylev (approved):
This is interesting, but I'm not sure the added complexity is worth it. We have some theoretical numbers on how much this helps performance, but do we have any indication that this will actually help in practice?

miwig:
I assume that compare.html is the page that's accessed the most so I'll focus on that:
On my reasonably fast home connection in Germany, `shared.js` and `perf.js` take on average 190ms (>99% of which is latency) to load concurrently, so that's what would be saved if the cache is hot. It's honestly not a lot, but it's still 40% of the time to `DOMContentLoaded`, and 20% of total page load time (including the actual comparison data), so the page might feel a little bit snappier :)

Mark-Simulacrum:
I would prefer to avoid embedding the git commit hash -- can we generate a UUID or similar on startup and just use that as the etag? That'll also avoid problems with local development where you may not commit but still want to see changes made.

For the static assets I suspect the better solution would probably be to add cache headers to them (IIRC, there's something like stale-while-revalidate, which should mean that the results can be used while the cache goes to check for X seconds) and then try putting perf.rust-lang.org behind a CDN, which should be relatively straightforward to do.

(Maybe just starting with the CDN is better.)

miwig:
Getting `/perf/info` also delays page load by 190ms every time because it's done sequentially. Might be worth just including that data in `/perf/get` and `/perf/graphs`.

miwig:
>(IIRC, there's something like stale-while-revalidate, which should mean that the results can be used while the cache goes to check for X seconds)

I initially dismissed that simply because the `headers` crate doesn't have it built-in, and versioned links with long cache duration seemed appealing :smile:, but you're right, that is a much simpler solution that should be beneficial even without a CDN.

> can we generate a UUID or similar on startup and just use that as the etag?

A `Last-Modified` header is probably the correct solution then.

miwig:
@Mark-Simulacrum I have pushed a simplified version using a `Last-Modified` header with the file modification date and `stale-while-revalidate`.

Mark-Simulacrum:
I am wary of file modification dates - my experience is that container filesystems (like we use in prod) often end up with wrong or at least low granularity dates for various reasons. Maybe we can use start time of the server, instead? Ideally I'd prefer to have something that's definitely right and gets "fixed" on deploys even if they don't obviously touch files.

miwig:
> I am wary of file modification dates - my experience is that container filesystems (like we use in prod) often end up with wrong or at least low granularity dates for various reasons. Maybe we can use start time of the server, instead? Ideally I'd prefer to have something that's definitely right and gets "fixed" on deploys even if they don't obviously touch files.

I went back to your earlier suggestion and used a random UUID.

Mark-Simulacrum (approved):
On a pretty good network (gigabit) I get the following results currently on compare page:

image

and on index:

image

I'm thinking we'll re-evaluate after merging and revert particularly if we start seeing trouble with cache invalidation (e.g., awkward mixes of JS and HTML or similar). Given this is just affecting static files (which have hopefully the fastest response times, since they're doing ~zero computation, just reading a few kilobytes out over the network), my hope is that this is a relatively cheap and easy change to make, though likely with limited impact as well.

PR #1223, opened 2022-03-21
Remove .git directory when downloading a crate from git(hub)

Kobzol:
The `.git` directory should not be kept there.

### DIFF
--- a/collector/src/main.rs
+++ b/collector/src/main.rs
@@ -1252,8 +1252,28 @@ fn download_from_git(target: &Path, url: &str) -> anyhow::Result<()> {
         .status()
         .expect("Git clone failed");
     generate_lockfile(tmpdir.path());
+
+    // Save downloaded commit SHA to keep information about downloaded revision
+    let git_sha = get_git_sha(tmpdir.path())?;
+    std::fs::write(tmpdir.path().join("git-commit.txt"), git_sha)?;
+
+    // Remove .git directory to avoid messing with nested git repositories
+    if let Err(error) = std::fs::remove_dir_all(tmpdir.path().join(".git")) {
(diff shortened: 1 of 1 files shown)


collector/src/main.rs
```diff
@@ -1252,6 +1252,9 @@ fn download_from_git(target: &Path, url: &str) -> anyhow::Result<()> {
         .status()
         .expect("Git clone failed");
     generate_lockfile(tmpdir.path());
+
+    std::fs::remove_dir_all(tmpdir.path().join(".git")).expect("Could not delete .git directory");
```
rylev, line 1256: I think it would be better to just log an error with `eprintln!` if deleting the directory fails instead of crashing. After all, including `.git` isn't incorrect, it's just a bit wasteful. What do you think?

Kobzol: I stumbled upon the `.git` directory when I was adding a new benchmark. If the `.git` directory is left there, it will be problematic to work with the directory using `git`, for example if I wanted to add the downloaded benchmark to the `rustc-perf` suite. But in most cases crates downloaded from `git` probably won't be added to the repository anyway, so it's not a big deal.

Kobzol:
Good idea, I added it.

rylev (approved):
Should we wait until all the currently open new benchmarks are merged and then add the git-commit.txt to them?

Kobzol:
I assume that pretty much all of the benchmarks currently in the repository are downloaded not from git(hub), but from `crates.io`, which makes it a bit redundant to add the commit to them I suppose.

The download from git is more useful for local profiling and downloading e.g. personal projects from github that are mentioned in `rustc` issues as regressions.

rylev:
Ah good point! I'll merge then 😊

PR #1225, opened 2022-03-21
Fix graphs labels

miwig:
The graphs page selects the axis label based on `state.absolute`, but at some point that went away, so the label always reads "% change from baseline", even when absolute values are shown. I've changed it to select based on `state.kind`, and did some other stuff along the way:
* cpu-clock now labeled as "CPU time" instead of "Wall time execution"
* wall-time now just "Wall time" instead of "Wall time execution"
* "Number of" replaced by "(count)"
* max-rss unit changed to "kB" instead of "kb", which could be misunderstood to mean "kilobit"
* "baseline" clarified as "first"

Some of these are also partly because uPlot doesn't support multi-line axis labels (see: https://github.com/leeoniya/uPlot/issues/675), so I tried to make them a bit shorter to fit better.

### DIFF
--- a/site/src/selector.rs
+++ b/site/src/selector.rs
@@ -479,8 +479,8 @@ impl StatisticSeries {
                 SeriesResponse {
                     series: StatisticSeries {
                         artifact_ids: ArtifactIdIter::new(artifact_ids.clone()),
-                        points: if metric == *"cpu-clock" {
-                            // Convert to seconds -- perf reports this measurement in
+                        points: if metric == *"cpu-clock" || metric == *"task-clock" {
+                            // Convert to seconds -- perf reports these measurements in
                             // milliseconds
                             points
                                 .into_iter()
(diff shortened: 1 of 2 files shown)


miwig:
cpu-clock and wall-time do have units. Do you mean the ones that didn't have labels at all before? (task-clock(:u) and cpu-clock:u) I only edited what was already there, but I suppose I could add labels for the rest, too. Although the :u variants don't seem to have any data present...

miwig:
@Mark-Simulacrum I've added label+unit for task-clock, and made task-clock display in seconds instead of ms, just like cpu-clock. I've not done anything for task-clock:u and cpu-clock:u, which have no data (and as far as I can tell from 2630cf74b52d06ea79e22fc38822e51c6af93051 haven't ever been collected) but still appear in pstat_series somehow...

rylev (approved):
This looks good to me. Thanks!

PR #1230, opened 2022-03-22
Add triage for this week

Mark-Simulacrum:
Will post TWIR shortly.

### DIFF
--- a/triage/2022-03-22.md
+++ b/triage/2022-03-22.md
@@ -1,0 +1,27 @@
+# 2022-03-22 Triage Log
+
+Very quiet week for performance, with just one statistically significant change
+landing in the last week. This change was a regression, though primarily in
+stress tests, and was a result of a soundness fix.
+
+Triage done by **@simulacrum**.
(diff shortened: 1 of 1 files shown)


triage/2022-03-22.md
```diff
@@ -0,0 +1,27 @@
...
+- Minimal impact on primary benchmarks, and as a soundness fix naturally pretty
+  justified.
+
+#### Improvements
+
+- No improvements.
+
+#### Mixed
+
+- No mixed results.
```
rylev, line 27: Are we no longer including the untriaged PRs?

Mark-Simulacrum: I am dropping them, yeah, the list in the generation is just some subset of 30 but there's >100 now IIRC and I don't think anyone is actively investing in that triage.

rylev: Ok at some point we should talk about how we want to handle regressions going forward.

rylev (approved):

PR #1234, opened 2022-03-23
Move the `piston-image` benchmark from primary to stable

rylev:
`image-0.24.1` is now a primary benchmark, and so we can move `piston-image` to just be a stable benchmark.


nnethercote (approved):
Thanks!

PR #1244, opened 2022-03-25
Remove serde (in favor of serde-1.0.136)

rylev:
A follow up to #1233 - We remove the old `serde` crate which has been superseded by `serde-1.0.136`.


nnethercote (approved):
I was going to say you forgot to remove `serde` from the README.md, but I see you did that when you added `serde-1.0.136`.

PR #1251, opened 2022-03-27
Add `hyper-0.14.18` as a new benchmark

lqd:
This adds `hyper-0.14.18` as a replacement for the `hyper-2` benchmark.

Build times comparison:

|   | `hyper-2`  |  `hyper-0.14.18` |
|---|---:|---:|
| check  | 13.73s | 10.97s |
| debug  | 14.50s | 12.71s |
| opt  | 14.59s | 17.12s |

As for the leaf crate itself, it's 6-10x smaller than before: 

image

I think the structure may be different with respect to the default features, and that realistic uses in dependent crates need to opt into some of the client/server/http1/http2/etc features. It is bigger than a helloworld, but I don't know if we'd want to enable some of these or go with the default features here ?


nnethercote:
@lqd: can you measure the final crate time with some of those features you suggested? If that makes a big difference, good, if not, perhaps it's not worth keeping this benchmark.

lqd:
Sure.

I've tried `--features client,http1,http2,server,stream`, and didn't go with the `full` feature since that also depends on `tokio`. That looks quite alright.

Here are updated timings, for full compilation (since some of these flags incur new dependencies):

|   | `hyper-2`  |  `hyper-0.14.18` | `hyper-0.14.18` w/ features |
|---|---:|---:|---:|
| check  | 13.73s | 10.97s | 15.89s |
| debug  | 14.50s | 12.71s | 17.37s |
| opt  | 14.59s | 17.12s | 20.02s | 

And for the leaf crate:

|   | `hyper-0.14.18`  |  `hyper-0.14.18` w/ features |
|---|---:|---:|
| check  | 0.36s | 2.55s |
| debug  | 0.48s | 3.23s |
| opt  | 0.94s | 7.10s |

nnethercote:
Sounds good. Can you mention the features being enabled in the `collector/benchmarks/README.md` description?

lqd:
I've amended the scaffolding commit to update the `perf-config.json` files and readme.

Here's how the results look now: 

image

nnethercote (approved):

PR #1254, opened 2022-03-28
Comparison page table tweaks and fixes

rylev:
This fixes two issues in the comparison page:
* The headings for the table weren't centered but the data was. It looks off IMO
* The labels for the raw data columns were not rendering correctly.

### Before
image

### After
image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -219,7 +219,7 @@
         }
 
         .benches th {
-            text-align: left;
+            text-align: center;
             word-break: break-word;
             width: 25%;
             min-width: 50px;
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Great, looks much better :) The wrong label was my bad from a previous PR, thanks for the fix.
Maybe I would also remove `word-wrap: break-word;` from the `<th>` element to avoid this:
image
But it's not a big deal.

I suppose that we should merge the Vue3 changes asap to avoid hoarding conflicts in the UI.

lqd:
We should likely right-align all numbers instead of centering, as it makes scanning and comparisons between rows harder, especially without thousand-separators.

Kobzol:
This is how it would look with right align:
image

It's probably a good idea to do this for the numeric columns.

PR #1255, opened 2022-03-28
Upgrade to Vue3

rylev:
This supersedes #1213 and does the upgrade without breaking the comparison.html page into separate files which makes review harder. 

Most of the changes are just needing to move code around. The changes themselves are pretty minimal:

* Vue3 works with an app instance instead of global functions that mutate a global app
* We request the data after the app has mounted and directly apply the data to the `this` instance inside of `mounted`

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -285,7 +285,7 @@
             text-align: center;
         }
     </style>
-    <script src="https://cdn.jsdelivr.net/npm/vue@2/dist/vue.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/vue@3/dist/vue.global.prod.js"></script>
 </head>
 
 <body>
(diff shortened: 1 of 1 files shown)


Kobzol:
Clicking on `Do another comparison ▶` does not work for me (it doesn't do anything, no additional errors in the console).

rylev:
@Kobzol good catch! I fixed the issue by moving the code that adds the click handlers to after the app is mounted. It now works!

site/static/compare.html
```diff
@@ -1040,7 +1042,14 @@ <h2>Comparing <span id="stat-header">{{stat}}</span> between <span id="before">{
             window.location.search = params.toString();
         }
 
-        loadState(makeData);
+        app.mount('#app');
+
+        document.getElementById("filters-toggle").onclick = (e) => {
```
Kobzol, line 1047: Maybe this should be in `mounted` rather than just after calling `mount` (I'm not sure if `mount` is synchronous?). But it seems to work.

Kobzol (approved):
LGTM

site/static/compare.html
rylev: These elements are not controlled by vue so there's not really a need to do this in `mounted`.

PR #1261, opened 2022-03-30
Align numeric columns to right in compare table

Kobzol:
Continuation of https://github.com/rust-lang/rustc-perf/pull/1254.
image
image

The `% Change` header looks a bit weird. This is how it looks right aligned:
image

`#content td` was removed, since it was overwriting specificity of other rules and it was only really used in benches and bootstrap (there is no other `<td>`).

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -206,11 +206,11 @@
             height: 10px;
         }
 
-        #content td {
+        #bootstrap th {
             text-align: center;
         }
 
-        #bootstrap th {
+        #bootstrap td {
             text-align: center;
         }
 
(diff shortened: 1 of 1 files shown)


rylev (commented):
I wonder if it would make more sense to just right align the raw numbers. The percentage change and significance factor are pretty easy to scan even when centered (and we could conceivably always display them as NN.NN if we wanted them to occupy the same width).

Kobzol:
I agree that it's most important for the raw numbers. The significance factor isn't that important IMO, but for the % change it's easier for me to judge the size difference if it's aligned to the same edge. But it's a small thing.

rylev:
@Kobzol are you planning on changing the alignment of *just* the raww data or do you still think the right alignment for all of the numeric columns is the right way to go?

Kobzol:
I still think that right aligning the numeric column with % change would look a bit better, but it would also make the header look silly, and the % change is quite readable as is for the moment. So I'll only modify the raw numbers column in this PR.

rylev (approved):

PR #1266, opened 2022-04-01
Refactor handling of historical data

rylev:
This refactor came from realizing that we weren't using the `BenchmarkVarianceDescription` at all and that what we were calling `BenchmarkVariances` is actually just historical data we use for many different calculations (some related to variance and others not). 

We now more appropriately label this data as "historical" which makes the code much easier to read. 

I also updated the glossary to remove terms for concepts we don't use any more and added terms for things we do use.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -35,11 +35,10 @@ The following is a glossary of domain specific terminology. Although benchmarks 
 ## Analysis
 
 * **test result delta**: the difference between two test results for the same metric and test case.
-* **significant test result delta**: a test result delta that is above some threshold that we have determined to be an actual change in performance and not noise. 
-* **noisy test case**: a test case for which the non-significant test result deltas are on average above some threshold. Non-noisy data tends to be very close to zero, and so a test case that produces non-significant test result deltas that are not close enough to zero on average are classified as noisy.
+* **significance threshold**: the threshold at which a test result delta is considered "significant" (i.e., a real change in performance and not just noise). This is calculated using some statistical measure (see the code for how this is currently being done).
+* **significant test result delta**: a test result delta above the significance threshold.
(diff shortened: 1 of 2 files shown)


docs/glossary.md
```diff
@@ -35,11 +35,10 @@ The following is a glossary of domain specific terminology. Although benchmarks
 ## Analysis
 
 * **test result delta**: the difference between two test results for the same metric and test case.
-* **significant test result delta**: a test result delta that is above some threshold that we have determined to be an actual change in performance and not noise. 
-* **noisy test case**: a test case for which the non-significant test result deltas are on average above some threshold. Non-noisy data tends to be very close to zero, and so a test case that produces non-significant test result deltas that are not close enough to zero on average are classified as noisy.
+* **significance threshold**: the threshold at which a test result delta is considered "significant" (i.e., a real change in performance and not just noise). This is calculated using some statistical measure (see the code for how this is currently being done).
```
Kobzol, line 38: Maybe we could either provide a link to the location in code where this happens or include the current implementation's name (IQR) here? I suppose that it won't change that much so that it would be hard to synchronize with the glossary.

Kobzol (approved):
Great, it is indeed now much more readable, I was looking for some notion of historical data in the code and the name "variance" was quite confusing.

docs/glossary.md
rylev: I'm unsure about this. I purposefully didn't include it for fear of it getting out of sync, but perhaps you're right that it won't change often enough for that really to be a concern.

Kobzol: Well, apparently the glossary **did** get out of sync historically, since you now had to update it :D But I think that with proper code reviews we can manage to keep it in sync.

rylev: Ok let's give it a try. 😊

PR #1267, opened 2022-04-01
Fix `get_error` query

rylev:
As has been noted, the status page shows old benchmarks in the section on build success. This is because the query for errors is incorrect.

The code previously made the assumption that the set of benchmarks that have ever had an error and the set of benchmarks run for the artifact in question are the same, but this has long not been true (and is very, very not true now).

First, we need to change the sql query: When doing a left join, all rows from the left table (in this case error_series) will be in the resulting table with the rows nulled out where the inner join condition is not met. This is not what we want. Instead, we want an inner join which only selects row where the join condition is met. 

However, the frontend still makes the assumption that the status page shows all the benchmarks that were run for the previous commit. But we can no longer rely on the `get_error` data giving us that data.

There are two ways to go about fixing this:
* Add a query for figuring out what benchmarks were run for a given artifact id. Then combine those artifacts with the data about which of them errored out. This would keep the front end experience the same (only with the correct data)
* Change the frontend experience to only display errors. The successfully built benchmarks will not be shown.

Because option 2 is simpler, I went with that for now, but we can discuss whether we want to actually go with option 1.

### DIFF
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -104,8 +104,7 @@ pub trait Connection: Send + Sync {
         series: u32,
         artifact_row_id: ArtifactIdNumber,
     ) -> Option<QueryDatum>;
-    async fn get_error(&self, artifact_row_id: ArtifactIdNumber)
-        -> HashMap<String, Option<String>>;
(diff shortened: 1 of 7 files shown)


Mark-Simulacrum (approved):
I'm good with merging this in the meantime though.

PR #1268, opened 2022-04-01
Add missing newline and remove duplicate header from PR summary

Kobzol:
Fixes [this](https://github.com/rust-lang/rust/pull/95171#issuecomment-1085849415).

I'm pretty sure that the main problem was missing newline after the short summary.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -706,11 +706,12 @@ async fn categorize_benchmark(
         generate_short_summary(secondary.as_ref(), is_master_commit);
 
     let mut result = format!(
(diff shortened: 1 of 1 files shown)


lqd (approved):
The newlines could also be added to the format string but that's no big deal. 

👍

Kobzol:
It was a raw literal, so I couldn't use `\n` and would have to add actual empty newlines which would look weird. Anyway, let's see if it helped.

Kobzol:
Merging to fix this asap, hopefully it helps.

PR #1272, opened 2022-04-04
Correctly calculate largest change independent of its direction in PR summary table

Kobzol:
The previous logic was wrong, as could be seen e.g. [here](https://github.com/rust-lang/rust/pull/95555#issuecomment-1087365963).


rylev (approved):
Whoops! Sorry for not catching this before! 👍

PR #1274, opened 2022-04-04
Improve summary table and GitHub messaging

rylev:
This includes several small tweaks:
* Don't include footnotes defining terms in the triage report. Overall, it looks off, and if we decide we really need to define terms we should do it once per triage report.
* Only show errored benchmark runs in the GitHub message (it's not really relevant for the triage report). 
* Show both primary and secondary benchmarks that error (we were previously only showing primary).
* Tweak the GitHub message when there are not enough changes to be significant. It is possible that there are individual changes that on their own are deemed relevant, but the comparison as a whole isn't relevant. The new messaging makes this a bit clearer by saying the "changes [are] not relevant" (i.e., point to the whole) rather than "no relevant changes".

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -755,6 +755,7 @@ pub struct Comparison {
     pub b: ArtifactDescription,
     /// Statistics based on test case
     pub statistics: HashSet<TestResultComparison>,
+    /// A map from benchmark name to an error which occured when building `b` but `a`.
     pub new_errors: HashMap<String, String>,
 }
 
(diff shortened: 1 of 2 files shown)


site/src/github.rs
```diff
@@ -700,7 +700,25 @@ async fn categorize_benchmark(
...
+                    .errors_in()
+                    .iter()
+                    .map(|benchmark| format!("- {benchmark}"))
+                    .collect::<Vec<_>>()
+                    .join("\n")
+            };
+            write!(
+                result,
+                "\nThe following benchmark(s) failed to build:\n{}{}\n",
+                list_errored_benchmarks(primary),
```
Kobzol, line 717: We currently duplicate the same errors into `primary` and `secondary` (see `summarize_by_category`). This would result in each error being printed twice.

So either this logic should be reverted or we should properly split the errors by their category. TBH I don't think that we need to distinguish build errors by category, so I would just revert this change.

Kobzol (changes requested):

site/src/comparison.rs
```diff
@@ -760,6 +755,7 @@ pub struct Comparison {
     pub b: ArtifactDescription,
     /// Statistics based on test case
     pub statistics: HashSet<TestResultComparison>,
+    /// A map from benchmark name to an error which occured when building `b` but `a`.
```
Kobzol, line 758: `b` but `a`?

Kobzol (approved):

PR #1276, opened 2022-04-05
Always show errors when present

rylev:
Previously when there were no relevant changes to show, we didn't show new benchmark errors. For example, in [this PR](https://github.com/rust-lang/rust/pull/95337#issuecomment-1088407867), the perf result showed as not being relevant and thus hid the fact that one of the benchmarks was not building. It always makes sense to show new benchmark errors even if the perf results are otherwise not relevant. This change makes it so.

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -125,7 +125,10 @@ pub async fn handle_compare(
         })
         .collect();
 
-    let mut new_errors = comparison.new_errors.into_iter().collect::<Vec<_>>();
+    let mut new_errors = comparison
+        .newly_failed_benchmarks
+        .into_iter()
+        .collect::<Vec<_>>();
     new_errors.sort();
     Ok(api::comparison::Response {
         prev,
@@ -172,8 +175,6 @@ pub struct ComparisonSummary {
     num_improvements: usize,
     /// The cached number of comparisons that are regressions
     num_regressions: usize,
-    /// Which benchmarks had errors
-    errors_in: Vec<String>,
 }
 
 impl ComparisonSummary {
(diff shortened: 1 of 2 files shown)


site/src/github.rs
```diff
@@ -675,8 +690,8 @@ async fn categorize_benchmark(
     if primary.is_none() && secondary.is_none() {
         return (
             format!(
-                "This benchmark run did not return any relevant results.\n\n{}",
-                DISAGREEMENT
+                "This benchmark run did not return any relevant results.\n\n{}{}",
+                DISAGREEMENT, errors
```
Kobzol, line 694: Maybe we can combine `DISAGREEMENT` + `errors` into some kind of `footer` variable, since it's used in two places? So that we don't forget to update it on both places if something in the footer changes.

site/src/comparison.rs
```diff
@@ -798,21 +791,21 @@ impl Comparison {
...
+            .newly_failed_benchmarks
             .into_iter()
             .partition(|(b, _)| category_map.get(&b.as_str().into()) == Some(&Category::Primary));
 
         let primary = Comparison {
             a: self.a.clone(),
             b: self.b.clone(),
             statistics: primary,
-            new_errors: primary_errors,
+            newly_failed_benchmarks: primary_errors,
```
Kobzol, line 802: It seems to me that the partition here is now useless, since the summary doesn't even access the fields? But on the other hand it keeps the comparisons in a consistent state, although it's not used at this moment.

rylev: Yes, I have an idea how to clean this up, but I'd like to do it in a separate PR.

Kobzol (approved):

rylev:

PR #1281, opened 2022-04-06
Simplify artifact comparison relevance

rylev:
Based on [discussion](https://github.com/rust-lang/rustc-perf/pull/1277#issuecomment-1090142028) in #1277, this PR drastically simplifies how we think about relevance of an artifact comparison. If there are any relevant test result comparisons, then the artifact comparison as a whole is deemed relevant. 

We should expect that more items will appear on the triage report. 

@nnethercote - I like this change for the GitHub messages we post, but I'm less keen on this for the triage report. Perhaps for the triage report we can be a bit smarter. For example, if the primary summary has any comparisons OR the secondary summary has any medium or above test result comparison or 10 or more small ones. This would at least mean the triage filer doesn't need to filter out those that are just composed of small changes to secondary benchmarks. Thoughts?

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -146,29 +146,31 @@ pub async fn handle_compare(
 async fn populate_report(
     ctxt: &SiteCtxt,
     comparison: &Comparison,
-    report: &mut HashMap<Option<Direction>, Vec<String>>,
+    report: &mut HashMap<Direction, Vec<String>>,
 ) {
-    if let Some(summary) = ComparisonSummary::summarize_comparison(comparison) {
-        let relevance = summary.relevance_level();
-        if relevance.at_least_somewhat_relevant() {
-            if let Some(direction) = summary.direction() {
-                let entry = report
-                    .entry(relevance.very_relevant().then(|| direction))
-                    .or_default();
+    let benchmark_map = ctxt.get_benchmark_category_map().await;
+    let (primary, secondary) = comparison.clone().summarize_by_category(benchmark_map);
+    // Get the combined direction of the primary and secondary summaries
+    let direction = match (primary.direction(), secondary.direction()) {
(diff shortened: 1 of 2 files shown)


site/src/comparison.rs
```diff
@@ -146,37 +146,39 @@ pub async fn handle_compare(
...
-        }
-    }
+    let benchmark_map = ctxt.get_benchmark_category_map().await;
+    let (primary, secondary) = comparison.clone().summarize_by_category(benchmark_map);
+    // Get the combined direction of the primary and secondary summaries
+    let direction = match (primary.direction(), secondary.direction()) {
+        (Some(d1), Some(d2)) if d1 != d2 => Direction::Mixed,
+        (Some(Direction::Improvement), Some(_) | None) => Direction::Improvement,
+        (Some(Direction::Regression), Some(_) | None) => Direction::Regression,
+        (Some(Direction::Mixed), Some(_) | None) => Direction::Mixed,
```
Kobzol, line 158:
```suggestion
        (Some(d), Some(_) | None) => d,
```

Kobzol (approved):
:+1: I think that it's a good idea to make this simpler first and only if we later find that there is some problem with it, we can introduce some additional logic. Deciding based on # and magnitude of relevant benchmarks in primary/secondary category sounds reasonable as a potential future step.

nnethercote (approved):
I agree with @Kobzol: let's try it and see how it goes, we can adjust if necessary.

It would be good to make sure the triagers know about this change.

rylev:
cc @Mark-Simulacrum @pnkfelix - Just so you're aware, this change will cause many more PRs to show up on the triage report as perf improvements or regressions. I imagine we'll likely want to reinstate some logic that limits what shows up on the triage report but we're going back to a simple state (show anything that causes any relevant changes) and then rebuilding from there. I'll cc you on any additional changes.

PR #1286, opened 2022-04-08
Create new ctfe-stress-5 which builds in half the time as ctfe-stress-4

rylev:
As part of the benchmark overhaul, we're making a new ctfe stress test that builds in about half the time (since there's not much reason for it to take as long as it currently does).

In a follow-up, we will remove ctfe-stress-4.

The ctfe-stress-4 benchmark was copied over to ctfe-stress-5, two annotations for nightly features which have since been stabilized were removed, and the number of ctfe iterations was reduced by a factor of about ~12.

ctfe-stress-4
* check: 4.8s
* build: 4.7sc
* build --release: 4.4s

ctfe-stress-5
* check: 2.3
* build: 2.3sc
* build --release: 2.3s


nnethercote (approved):
Lovely, thank you.

PR #1291, opened 2022-04-11
Ensure link to compare page is always present

### LINKED ISSUE #1289: Restore link to perf run results in non relevant runs
Anyway, work was done and this link don't cost anything to post, but require some jumps around in case if someone actually want to look at results.

Plus, isn't posted results says about `instructions` metric only?

For example: https://github.com/rust-lang/rust/pull/95435

rylev:
Fixes #1289 

After some refactoring, we accidentally removed the link to the compare page when the results of a perf run were not relevant (see [here](https://github.com/rust-lang/rust/pull/95914#issuecomment-1094642321) for an example). This adds it back and simplifies how we build up the message to be posted on GitHub.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -602,35 +602,41 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit: 
     please file an issue in [rust-lang/rustc-perf](https://github.com/rust-lang/rustc-perf/issues/new).";
     let footer = format!("{DISAGREEMENT}{errors}");
 
-    if !primary.is_relevant() && !secondary.is_relevant() {
-        return format!("This benchmark run did not return any relevant results.\n\n{footer}");
-    }
-
-    let primary_short_summary = generate_short_summary(&primary);
-    let secondary_short_summary = generate_short_summary(&secondary);
-
-    let mut summary = format!(
-        r#"
-- Primary benchmarks: {primary_short_summary}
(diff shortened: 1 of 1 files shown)


site/src/github.rs
```diff
@@ -602,35 +602,41 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit:
...
-    let direction = primary.direction().or(secondary.direction());
-    let next_steps = next_steps(primary, secondary, direction, is_master_commit);
+        write_summary_table(&primary, &secondary, true, &mut message);
 
-    format!(
-        "Finished benchmarking commit ({sha}): [comparison url]({comparison_url}).
+        let direction = primary.direction().or(secondary.direction());
+        let next_steps = next_steps(primary, secondary, direction, is_master_commit);
+        write!(&mut message, "\n{footer}\n{next_steps}").unwrap();
+    };
```
Kobzol, line 637:
```suggestion
    }
```

Kobzol (approved):
Great :) I wonder if we should also generate `-perf-regression` even if no relevant results were found? E.g. a previous run might have generated a regression, and another one (after some change) could return no relevant changes.

rylev:
> Great :) I wonder if we should also generate `-perf-regression` even if no relevant results were found? E.g. a previous run might have generated a regression, and another one (after some change) could return no relevant changes.

That's a great point. I don't believe we ever did that though so let's tackle that in another PR.

PR #1293, opened 2022-04-13
Remove redundant partitioning in `summarize_by_category`

rylev:
This gets rid of some unnecessary work we were doing.


Kobzol (approved):
Looks much cleaner now :+1:

PR #1296, opened 2022-04-14
Add link to benchmark code in summary page

Kobzol:
This should make the benchmarks more discoverable and should provide easy answer to ponderings such as "is this a binary or a library crate?" (which btw could also be shown directly in the summary page).

There is a slight problem with this approach, because if the benchmark gets out of date, the link will result in 404. But I don't think that it's such a big deal, as usually you look at a recent perf. result and for that the link should be fine. If we wanted to solve this, we would need to embed the version of `rustc-perf` that produced this specific perf. run into the link, but I'm not sure if that's readily available (and it wouldn't work if you compared two commits benchmarked by a different `rustc-perf` version anyway).

link

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -291,6 +291,10 @@
             margin: 10px;
             text-align: center;
         }
+
+        .silent-link {
+            color: inherit;
+        }
     </style>
     <script src="https://cdn.jsdelivr.net/npm/vue@3/dist/vue.global.prod.js"></script>
 </head>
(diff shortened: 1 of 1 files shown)


lqd:
I like it myself, what do the others think ?

nnethercote (approved):

Mark-Simulacrum:
IIRC we record the rustc-perf commit (or I wanted to) into the db for each collection, but yeah, it's not worth it to pull that in I think.

Shouldn't we link to the collector readme for the benchmark though? That contains more meaningful details than the code, seems likely to be more useful.

lqd:
@Mark-Simulacrum here our use-case was more about differentiating bin targets from libs, and looking at the perf-config.json file (or the code).

rylev (approved):
Cool! I think this is fine for now. We in fact do record the version of rustc-perf that ran a particular benchmark so we could provide links that never 404, but I'm not sure it's worth it.

PR #1300, opened 2022-04-20
Add benchmark/scenario/profile filters to graphs page and graph history link to compare page

Kobzol:
Based on suggestions mentioned in [this](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/interactive.20graph.20UX) Zulip thread.

This PR basically does three things:
1) Add the option to filter by `benchmark`, `scenario` and `profile` in the `/graphs` endpoint.
2) Fixes up the `graphs` page so that it works if the returned data does not contain all scenarios/profiles
(basically a continuation of https://github.com/rust-lang/rustc-perf/pull/1196).
3) Adds a link to the compare page which shows 2 week history chart for the specific benchmark/scenario/profile/stat combination. I wasn't sure where to put the link, so right now I just put it on the `Profile` column (that's temporary).

Currently, the changes from 1) are not exposed in the `graphs` UI, the filters can only be entered via GET parameters.

I'm not sure what is the best way to present the specific graph on the compare page. Maybe it should be rendered instead on the detailed query page (this would probably require some refactoring of the current code, so that we can share the chart code better) or we can also just include a link to the chart from the detailed query (but then it would require two clicks to get to it).

graphs

### DIFF
--- a/site/static/index.html
+++ b/site/static/index.html
@@ -102,6 +102,7 @@
 
         const otherCacheStateColors = ["#8085e9", "#f15c80", "#e4d354", "#2b908f", "#f45b5b", "#91e8e1"];
         const interpolatedColor = "#fcb0f1";
+        const basicProfiles = ["Check", "Debug", "Opt"];
 
         function tooltipPlugin({ onclick, commits, isInterpolated, absoluteMode, shiftX = 10, shiftY = 10 }) {
             let tooltipLeftOffset = 0;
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum (approved):
This overall looks good to me - I think it's a good start, though the long run probably involves moving the graphs and potentially using already fetched data to render directly on the comparison page.

site/static/index.html
```diff
@@ -380,6 +381,8 @@ <h3>This may take a while!</h3>
                         });
                     }
 
+                    let indices = cacheStates[Object.keys(cacheStates)[0]].interpolated_indices;
```
rylev, line 384: Note: we should rename `cacheStates` to `scenarios` at some point.

PR #1305, opened 2022-04-21
Make raw data numbers in compare page more readable

Kobzol:
To make the number more readable, we probably want to skip unnecessary decimal part (if it's zero anyway) and add separators. Currently, I used a locale-specific implementation that is built into JS. If you don't like that it is locale-specific, I can hand-code some variant that uses some fixed separators.

In any case, it now looks much more readable, especially when combined with the right alignment.

Before:
image

After (on my Czech locale):
image


nnethercote (approved):
Thank you, thank you, thank you! Using the locale is good.

PR #1308, opened 2022-04-25
Always include next steps in GitHub message

### LINKED ISSUE #1307: If perf run completed without any relevant results, github tag `S-waiting-on-perf` didn't removed
For example, https://github.com/rust-lang/rust/pull/96359 or https://github.com/rust-lang/rust/pull/96179

rylev:
Fixes #1307 

Before this change, the next steps instructions (which include label changes and bors directives) were only showed if the results were relevant. However, we should always show next steps even if the results aren't relevant. 

This has a few interesting changes to current behavior:
* The `S-waiting-on-perf` label will be removed when the try run finishes
* Even PRs that show no relevant perf changes will still be labeled as `@bors rollup=never`
* The `perf-regression` label will be removed if the run returns no relevant perf results

All these changes are how things worked in the past, and I believe are the right behavior.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -612,7 +612,7 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit: 
     if !primary.is_relevant() && !secondary.is_relevant() {
         write!(
             &mut message,
-            "This benchmark run did not return any relevant results.\n\n{footer}"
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Looks good.

I'm not sure if adding `+S-waiting-on-review` after each try run is always correct, but that's not for the current PR.

PR #1315, opened 2022-04-28
Add space after PR number in compare page

Kobzol:
Fixes [this](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Tasks.20for.20contributors/near/280336944).

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -340,7 +340,7 @@
                 <div class="description-box">
                     <div v-if="data.prev" class="description-arrow"><a v-bind:href="prevLink">&larr;</a></div>
                     <div style="padding: 10px;">
-                        <a v-if="data.a.pr" v-bind:href="prLink(data.a.pr)">#{{data.a.pr}}</a>
+                        <span><a v-if="data.a.pr" v-bind:href="prLink(data.a.pr)">#{{data.a.pr}}</a>&nbsp;</span>
                         <span v-if="data.a.date">{{formatDate(data.a.date)}}</span>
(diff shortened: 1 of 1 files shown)


site/static/compare.html
```diff
@@ -340,7 +340,7 @@ <h2>Comparing <span id="stat-header">{{stat}}</span> between <span id="before">{
                 <div class="description-box">
                     <div v-if="data.prev" class="description-arrow"><a v-bind:href="prevLink">&larr;</a></div>
                     <div style="padding: 10px;">
-                        <a v-if="data.a.pr" v-bind:href="prLink(data.a.pr)">#{{data.a.pr}}</a>
+                        <a v-if="data.a.pr" v-bind:href="prLink(data.a.pr)">#{{data.a.pr}}&nbsp;</a>
```
Mark-Simulacrum, line 343: Should this be *inside* the link? Wouldn't that make the underline from links go under the space as well?

Kobzol: Hmm, it's possible to botch even such a simple PR :D You're right, of course.

(Reminding myself that it would be nice to have an easier way of exporting data from the live site to offline storage, so that it's easier to check these things locally. I tried it with a hardcoded number and didn't notice it).

Kobzol:
I had to wrap the link inside `<span>`, otherwise the non-breaking space introduced two spaces (because it was now a separate "text" entity followed by a new-line, which was not stripped by the browser (at least not by Chrome).

Mark-Simulacrum (approved):

PR #1316, opened 2022-05-02
Add summary table at top of triage

rylev:
This adds a table at the top of the triage report that looks just like the other tables in the report but compares the start bound with the end bound showing how much perf changed from the start of the triage period to the end. This gives readers a better impression of whether perf regressed or improved as a whole over the entire triage period.


Kobzol (approved):
What a great idea! Code changes look OK, we'll see how the design looks at the next triage report I guess :)

PR #1322, opened 2022-05-15
Add extended stats to GH summary table

Kobzol:
This is an attempt to improve the (currently non-existing) display of secondary performance stats (like max RSS or cycles) in GH summary reports. This corresponds to the "better UI/UX" item from the performance roadmap.

Sometimes changes in icounts are virtually just noise, but some other metric changes in significant ways (typically max RSS, but sometimes also cycles). This can be missed by onlookers, since to show these secondary steps, you have to manually switch to them in the compare page, which is sometimes neglected.

I think that it would be nice if we could show the information in the summary, if we find that the changes are reasonably large (this part will be difficult, as always). The PR currently shows the same summary table that is currently used for icounts also for Max RSS, if we find that there have been large changes (TODO: define what does this mean). We can also do something different than showing the full table, like saying "Watch out, it seems that RSS has changed significantly, check this link to find more.".

The first commit is just small cleanup that enables using multiple summary tables in a single PR without duplicating footer with the footnote definition.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -1,7 +1,7 @@
 use crate::api::github::Issue;
 use crate::comparison::{
-    deserves_attention_icount, deserves_attention_maxrss, write_summary_table,
-    write_summary_table_footer, ArtifactComparisonSummary, Direction, Stat,
+    deserves_attention_icount, write_summary_table, write_summary_table_footer,
+    ArtifactComparisonSummary, Direction, Stat,
 };
 use crate::load::{Config, SiteCtxt, TryCommit};
 
@@ -11,7 +11,9 @@ use reqwest::header::USER_AGENT;
 use serde::{Deserialize, Serialize};
 
 use collector::category::Category;
+
 use std::collections::{HashMap, HashSet};
+
 use std::{fmt::Write, sync::Arc, time::Duration};
 
 type BoxedError = Box<dyn std::error::Error + Send + Sync>;
@@ -578,6 +580,40 @@ fn make_comparison_url(commit: &QueuedCommit, stat: Stat) -> String {
 }
 
 async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit: bool) -> String {
+    let benchmark_map = ctxt.get_benchmark_category_map().await;
+
+    let mut message = format!(
+        "Finished benchmarking commit ({sha}): [comparison url]({comparison_url}).
+
+**Summary**: ",
+        sha = commit.sha,
+        comparison_url = make_comparison_url(&commit, Stat::Instructions)
+    );
+
+    let mut table_written = false;
+    let stats = vec![
+        ("Instruction count", Stat::Instructions, false),
+        ("Max RSS (memory usage)", Stat::MaxRSS, true),
+        ("Cycles", Stat::Cycles, true),
+    ];
+
+    for (title, stat, hidden) in stats {
+        table_written |= write_stat_summary(
+            &benchmark_map,
+            &commit,
+            ctxt,
+            &title,
+            stat,
+            hidden,
+            &mut message,
+        )
+        .await;
+    }
+
+    if table_written {
+        write_summary_table_footer(&mut message);
+    }
+
     let comparison = match crate::comparison::compare(
         collector::Bound::Commit(commit.parent_sha.clone()),
         collector::Bound::Commit(commit.sha.clone()),
@@ -589,7 +625,6 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit: 
         Ok(Some(c)) => c,
         _ => return String::from("ERROR categorizing benchmark run!"),
     };
-
     let errors = if !comparison.newly_failed_benchmarks.is_empty() {
         let benchmarks = comparison
             .newly_failed_benchmarks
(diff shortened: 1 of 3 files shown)


lqd:
Nice, thanks for working on this!

Since the secondary stats may have too much variance, we could even maybe hide them in `<details>` when we notice a reasonably large change, e.g. with a label that is tentative about the change being actually significant and not just the usual random perturbations) ?  

(I wonder if the historical significance factor could also be helpful here, to smooth out the number of times these detections happen. We can iterate later in any case, if the initial steps are not super obtrusive and misleading, which is hard to predict.)

nnethercote (approved):

rylev (changes requested):
I'm in favor of this change though it's obvious we'll need to rapidly iterate as so much of the way things currently work is tuned to instruction counts and is not adjusted here.

The one thing I'd like to see changed before we merge this is a clearer indication of max-rss being a secondary stat. As noted in my comments, the way things currently are, it will be possible for us to report "no relevant results" but then show max-rss stats. We need to make this clearer to avoid a bunch of confusion.

site/src/github.rs
```diff
@@ -630,6 +647,29 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit:
         .unwrap();
 
         write_summary_table(&primary, &secondary, true, &mut message);
+        table_written = true;
+    }
+
+    if let Some((primary, secondary)) = analyze_secondary_stat(
```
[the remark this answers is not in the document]
Kobzol: I agree. Two options come to mind:
1) Change to wording to something like "the benchmark did not return any relevant results for this metric/for metric X".
2) Just don't show the corresponding stat at all if no relevant results were found and only state "no relevant results found" if nothing was found for any stat. In a way this would make Max-RSS no longer a secondary stat, but the stat would be more equal (apart from detection thresholds).

site/src/github.rs
```diff
@@ -640,6 +680,56 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit:
     message
 }
 
+// TODO: determine how should this work
+// Should this be separate from `deserves_attention`?
```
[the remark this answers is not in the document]
Kobzol: I agree that it could be confusing, but with the way the code is currently structured, I'm not sure if this structure would help much. In order to correctly propagate the `Stat` lineage, we would have to also include it in `ArtifactComparison` which creates the summaries from a given `stat`, otherwise we would need to pass the stat manually, which doesn't guard against mistakes. And furthermore, when we analyze the summaries, we treat instructions/max rss differently, it's not unified at all. The only safety precaution that this would provide is this:
```rust
pub(crate) fn deserves_attention_icount(
    summary: Summaries,
) -> bool {
    assert_eq!(summary.stat, Stat::Instructions);
}

pub(crate) fn deserves_attention_maxrss(
    summary: Summaries,
) -> bool {
    assert_eq!(summary.stat, Stat::MaxRSS);
}
```
which doesn't seem to be worth it.

To make it really bullet-proof, we would need to design e.g. some generics parameter like this:
```rust
struct Summaries<Stat> { primary, secondary, phantom: Stat }

fn deserves_attention_icount(summary: Summaries<Stat::Instructions>) { .. .}
```
but it seems like overkill here.

Making `Stat` an enum is a good idea, I'll do it.

Kobzol:
> Since the secondary stats may have too much variance, we could even maybe hide them in <details> when we notice a reasonably large change, e.g. with a label that is tentative about the change being actually significant and not just the usual random perturbations) ?

Hiding the secondary table(s) in `<details>` sounds like a good idea. What do others think?

> One concern is that I've noticed max-rss can be noisy in a correlated way: often all the benchmarks will improve or regress slightly on one run, but it's not actually meaningful. I don't know why this is.

I have also noticed this. Maybe we can tune the change detection algorithm for this and say something like "if at least X benchmarks had large/very large change, then we consider it to be interesting".

I have refactored the code slightly. The `deserves_attention` naming is a bit confusing, since for icounts we use it for next steps on PR summaries and for inclusion in the triage, while for Max RSS we use it for deciding whether we should display the summary table in the PR body.. sigh :D

Kobzol:
I thought about this a bit more, and I really like @lqd's suggestion. Instead of coming up with arbitrary thresholds, let's just always show MaxRSS and cycles, but at the same time:
1) Hide them in `<details>`, so that they won't take that much space, but they will be "one click away".
2) Include their mean in the `<details>` header, so we quickly see whether it's even worth it to click on it.

This will make the raw text content of the comment a bit larger, but I hope that won't be a problem. It will also do a few more work on the backend, but since perf. report PRs are not that frequent, I think that should be fine.

WDYT?

rylev:
Are we always going to show max-rss and cycles even if the instruction count is not significant? So, we keep the (not quite so arbitrary but still somewhat arbitrary) threshold for instruction count but get ride of it for other stats?

Kobzol:
Good point. I suggest this:
1) Always add Max RSS and cycles table, but keep it hidden.
2) Always show icount table. If there are relevant results, show it as before.
If there are no relevant results, hide it by default (same as RSS/cycles) + add the current string "the run did not find any relevant results".

I think that we should treat icounts a bit specially, as they are usually the most stable/least noisy stat and one that we primarily try to optimize.

Showing the icount table (even though it will be hidden by default) if there are no results in it seems a bit wasteful, but I don't consider that to be a big problem and I would rather do that than add another special case to the code. And also we can further convince the reader that the string "no relevant results were found" indeed means that the table is "empty". I think that this mostly only happens for icounts. I don't ever recall seeing a perf. result where cycles/max RSS didn't have at least a single relevant result (it might happen of course, but I think that it's quite rare).

Kobzol:
Just a reminder (I remembered this wrongly): currently there is no real heuristic for not showing the icount table in PR summaries (`deserves_attention` is not used for this!). The only case where the table is not shown is when the relevant result count is zero.

Kobzol:
As an example, I hand-crafted it [here](https://github.com/Kobzol/rustc-perf/issues/1#issuecomment-1129905682). If there are no relevant results, there is no point in showing the empty table.

rylev:
I think this is fine as long as we treat max-rss the same. If there are 0 results for max-rss, we should also not show an empty table.

Kobzol:
We could create some Markdown page in the `rustc-perf` repository with a short "Hitchhiker's guide to interpreting perf results" and link to it from the summary.

Kobzol:
I changed the code to always include Max RSS and cycles. Yes, I do realize that the instruction comparison and summaries are calculated twice, but it was just so painful to modify the code to remove the duplication that I left it like this :D Until we decide if we want to do this.

site/src/github.rs
```diff
@@ -566,23 +570,61 @@ async fn post_comparison_comment(ctxt: &SiteCtxt, commit: QueuedCommit, is_maste
...
-        "https://perf.rust-lang.org/compare.html?start={}&end={}",
-        commit.parent_sha, commit.sha
+    let benchmark_map = ctxt.get_benchmark_category_map().await;
+
+    let mut message = format!(
+        "Finished benchmarking commit ({sha}): [comparison url]({comparison_url}).
+
+**Summary**: ",
+        sha = commit.sha,
+        comparison_url = make_comparison_url(&commit, Stat::Instructions)
```
rylev, line 590: Does it make sense to link here now?

site/src/github.rs
```diff
@@ -594,33 +636,60 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit:
...
+    let comparison = match crate::comparison::compare(
+        collector::Bound::Commit(commit.parent_sha.clone()),
+        collector::Bound::Commit(commit.sha.clone()),
+        stat,
+        ctxt,
+    )
+    .await
+    {
+        Ok(Some(c)) => c,
+        _ => panic!(), //return Err("ERROR categorizing benchmark run!".to_owned()),
```
rylev, line 670: I don't think we want to leave it like this.

rylev (changes requested):
A few small things need to be addressed, but otherwise I'm ok with merging almost as is. We can then refactor.

site/src/github.rs
Kobzol: Well, maybe it's not strictly needed, but I think that people are used to it by now and having a visible link at the beginning of the summary seems like a useful thing. But I can remove it if you want.

Kobzol:
I refactored the stat summaries to remove duplication and the panic call.

site/src/github.rs
```diff
@@ -636,46 +626,65 @@ async fn summarize_run(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit:
...
+            "Instruction count",
+            Stat::Instructions,
+            false,
+            inst_comparison,
+        ),
+        (
+            "Max RSS (memory usage)",
+            Stat::MaxRSS,
+            true,
+            calculate_stat_comparison(ctxt, &commit, Stat::MaxRSS).await?,
```
rylev, line 645: nit: you could also set these to `None`, and in the loop check if there `None` and if they are run `calculate_stat_comparison` using the provided `stat`

Kobzol: It came to mind, but it seemed more "dirty" to me to do it with such a "mutable" way :) I can change it if you want.

nnethercote:
In the "Max RSS" section is it worth having a note that RSS measurements can be noisy?

Kobzol:
Directly to the PR comment or to the proposed page with a guide for interpreting the results?

nnethercote:
Either sounds ok, as long as it's somewhere that's fairly easy to find.

rylev:
I think maybe using `###` instead of `##` headers might be nicer, thoughts?

### With `##`
image


### With `###`
image

Kobzol:
That looks better, indeed.

PR #1330, opened 2022-05-19
Add newline before stat title in GH summary

Kobzol:
This is a hotfix for https://github.com/rust-lang/rustc-perf/pull/1322, as it has some [weird](https://github.com/rust-lang/rust/pull/97033#issuecomment-1131372445) formatting. Hopefully this fixes it.


site/src/github.rs
```diff
@@ -654,7 +654,7 @@ async fn summarize_run(
 
     for (title, stat, hidden, comparison) in stats {
         message.push_str(&format!(
-            "## [{title}]({})\n",
+            "\n## [{title}]({})\n",
```
rylev, line 657: I think the more correct fix is to ensure that each stat has a blank line after it no? 

```suggestion
            "## [{title}]({})\n\n",
```

rylev (commented):
Not sure if your suggestion or mine is the better way to fix it, thoughts?

(FYI: I manually edited the broken comment linked in this PR)

site/src/github.rs
Kobzol: Did this fix the broken comment in the PR? From my experience, it's fine if a header doesn't have a blank line after it in Markdown, so I thought that the problem is that there's no blank line between the details and the next header. But if this fixed the comment then sure, let's do this. I couldn't fix the comment manually because I don't have rights for it, so I couldn't experiment with it.

rylev: Ah no you're right. I think it's that the `</details>\n` closing tag needs to be `</details>\n\n`.

rylev: I also think we should remove the `**Summary:**` line.

It would be good if we could have tests for all of this...

Kobzol: Tests would be nice, I'll try to think about how we can separate the rendering from providing data to enable tests.

Kobzol:
I removed the summary heading, changes headers to `<h3>` (`###`) and added the newline.

rylev (approved):

PR #1332, opened 2022-05-19
Prefer the term metric over stat

rylev:
Currently, the code uses the term "stat" ambiguously. Sometimes it refers to a specific number statistic (e.g., 57,341,606,128 instruction counts recorded), and sometimes it refers to a class of statistic (e.g., "instruction counts"). [The glossary](https://github.com/rust-lang/rustc-perf/blob/master/docs/glossary.md) disambiguates these two using the term "stat(istic)" for the former and "metric" for the later. This PR brings the code into alignment with the glossary. 

Note: the API still uses "stat" for what is now called "metric". Since change the API is a little more error prone, I suggest we save that for a follow-up.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -1,7 +1,7 @@
 use crate::api::github::Issue;
 use crate::comparison::{
     deserves_attention_icount, write_summary_table, write_summary_table_footer, ArtifactComparison,
-    ArtifactComparisonSummary, Direction, Stat,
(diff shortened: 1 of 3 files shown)


site/src/github.rs
```diff
@@ -632,21 +632,21 @@ async fn summarize_run(
     let stats = vec![
```
Kobzol, line 632:
```suggestion
    let metrics = vec![
```

Kobzol (approved):

PR #1334, opened 2022-05-24
Make summary table vertical

Kobzol:
Based on [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Tables.20bad.20for.20TWIR.20format). The current look of the table can be seen [here](https://github.com/Kobzol/rustc-perf/issues/1#issuecomment-1135679996).

The table is now vertical.

I refactored the code to build a sort of a 2D matrix of values (`Vec<Vec<String>>`), so that it is easier to change the horizontal/vertical layout or column/row order in the future.

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -479,52 +479,20 @@ pub fn write_summary_table(
             .unwrap_or_else(|| "N/A".to_string())
     }
 
-    let columns = [
-        "          ", // we want at least 10 spaces to accommodate "count[^1]"
-        "Regressions 😿 <br />(primary)",
-        "Regressions 😿 <br />(secondary)",
-        "Improvements 🎉 <br />(primary)",
-        "Improvements 🎉 <br />(secondary)",
-        "All 😿 🎉 <br />(primary)",
-    ];
-    let counts: Vec<usize> = columns
-        .iter()
-        .map(|s| {
-            // Unicode emojis usually have width equal to ~2 chars (maybe? :) ).
-            let count = s.chars().count();
-            if s.chars().any(|c| !c.is_whitespace()) {
-                count + 1
-            } else {
-                count
-            }
-        })
-        .collect();
-    for column in &columns {
-        write!(result, "| {} ", column).unwrap();
-    }
-    result.push_str("|\n");
-    for &count in &counts {
-        write!(result, "|:{}:", "-".repeat(count)).unwrap();
-    }
-    result.push_str("|\n");
+    // (label, mean, max, count)
+    let mut column_data = vec![];
 
-    let mut render_row = |row: Vec<String>| {
-        debug_assert_eq!(row.len(), columns.len());
-        for (column, &count) in row.into_iter().zip(&counts) {
-            write!(result, "| {:<1$} ", column, count).unwrap();
-        }
-        result.push_str("|\n");
-    };
-    render_row(vec![
-        format!("count{}", if with_footnotes { "[^1]" } else { "" }),
-        primary.num_regressions.to_string(),
-        secondary.num_regressions.to_string(),
-        primary.num_improvements.to_string(),
-        secondary.num_improvements.to_string(),
-        (primary.num_regressions + primary.num_improvements).to_string(),
+    // label
+    column_data.push(vec![
+        "Regressions 😿 <br /> (primary)".to_string(),
+        "Regressions 😿 <br /> (secondary)".to_string(),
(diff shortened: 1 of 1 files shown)


site/src/comparison.rs
```diff
@@ -476,55 +476,23 @@ pub fn write_summary_table(
...
-    render_row(vec![
-        format!("count{}", if with_footnotes { "[^1]" } else { "" }),
-        primary.num_regressions.to_string(),
-        secondary.num_regressions.to_string(),
-        primary.num_improvements.to_string(),
-        secondary.num_improvements.to_string(),
-        (primary.num_regressions + primary.num_improvements).to_string(),
+    // label
+    column_data.push(vec![
+        "Worse (primary)".to_string(),
```
rylev, line 487: I personally find "Worse" and "Better" to be more confusing. How about "Improve" and "Regress"?

site/src/comparison.rs
```diff
@@ -538,7 +506,7 @@ pub fn write_summary_table(
             Some(secondary.arithmetic_mean_of_improvements())
         }),
         if primary.is_empty() {
-            "N/A".to_string()
```
rylev, line 541: I think "N/A" is more clear...

rylev (commented):
I like the layout change, but most of the changes to the content of the cells don't feel like improvements to me...

site/src/comparison.rs
Kobzol: I found improvements/regressions fine, but there was some interest in changing them (https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Tables.20bad.20for.20TWIR.20format/near/283322938, https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Tables.20bad.20for.20TWIR.20format/near/282641528). :man_shrugging:

site/src/comparison.rs
Kobzol: What I don't like about it is that it produces visual clutter, it's harder to find the real changes among a see of `N/A`. Compare:

|            | mean[^1] | max | count[^2] |
|:----------:|:--------:|:---:|:---------:|
| Worse (primary) | 20.0%    | 20.0% | 1         |
| Worse (secondary) |          |     | 0         |
| Better (primary) | -50.0%   | -50.0% | 1         |
| Better (secondary) |          |     | 0         |
| All (primary) | -15.0%   | -50.0% | 2         |


|            | mean[^1] | max | count[^2] |
|:----------:|:--------:|:---:|:---------:|
| Worse (primary) | 20.0%    | 20.0% | 1         |
| Worse (secondary) |    N/A      |  N/A   | 0         |
| Better (primary) | -50.0%   | -50.0% | 1         |
| Better (secondary) |    N/A      |  N/A   | 0         |
| All (primary) | -15.0%   | -50.0% | 2         |

But same as with all the other things, I don't have any hard preference here.

rylev (changes requested):
Looks good. Just one thing to address about row width.

site/src/comparison.rs
```diff
@@ -1314,11 +1323,13 @@ mod tests {
...
             ],
             r#"
-|            | Regressions 😿 <br />(primary) | Regressions 😿 <br />(secondary) | Improvements 🎉 <br />(primary) | Improvements 🎉 <br />(secondary) | All 😿 🎉 <br />(primary) |
-|:----------:|:------------------------------:|:--------------------------------:|:-------------------------------:|:---------------------------------:|:------------------------:|
-| count[^1]  | 3                              | 0                                | 0                               | 0                                 | 3                        |
-| mean[^2]   | 146.7%                         | N/A                              | N/A                             | N/A                               | 146.7%                   |
-| max        | 200.0%                         | N/A                              | N/A                             | N/A                               | 200.0%                   |
+|            | mean[^1] | max | count[^2] |
+|:----------:|:--------:|:---:|:---------:|
+| Regressions 😿 <br /> (primary) | 146.7%   | 200.0% | 3         |
```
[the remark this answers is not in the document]
Kobzol: TBH, I'm reluctant to optimize for plain text given how frequently is the table changing (and that it will also probably change again in the near future). But if you consider it important, I'll just set a fixed width for the columns.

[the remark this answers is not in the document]
Mark-Simulacrum: https://docs.rs/tabled/latest/tabled/ may be worth considering in the future as a dependency to ease our own work to build these.

Issue #1337, opened 2022-05-25
unable to select some comparison metrics: Failed to deserialize request


klensy:
Unable to select few comparison metrics, for example, `context-switches` with error:
```
Failed to deserialize request: Error("unknown variant `context-switches`, expected one of `instructions:u`, `cycles:u`, `faults`, `max-rss`, `task-clock`, `wall-time`, `cpu-clock`", line: 1, column: 126)
```

* Open some compare link, for example https://perf.rust-lang.org/compare.html?start=f80e454450c891c075617d23532c6eb317f4471d&end=6ac8adad1f7d733b5b97d1df4e7f96e73a46db42&stat=instructions:u
* Click `Do another comparison` and select other comparison metric,  `context-switches`, for example.
* Click submit, error will appear, but expected instead to show page with that metric.

This worked recently, so broken in last few commits, i guess.

klensy:
maybe #1322 ?

Kobzol:
Yeah this was most probably caused by my PR, I haven't seen this metric before. I'll fix it today.

klensy:
Not only this, one, some others too, so, please check other ones.

Kobzol:
Closing for now, let me know if there are any other metrics missing.

PR #1338, opened 2022-05-25
Generalize metrics

Kobzol:
My recent PR (https://github.com/rust-lang/rustc-perf/pull/1322) has broken displaying various metrics in the compare page (https://github.com/rust-lang/rustc-perf/issues/1337). This PR adds all (currently) known metric variants to the `Metric` enum.

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -45,7 +45,7 @@ pub async fn handle_triage(
     let mut before = start.clone();
 
     let mut num_comparisons = 0;
-    let metric = Metric::Instructions;
+    let metric = Metric::InstructionsUser;
     let benchmark_map = ctxt.get_benchmark_category_map().await;
     loop {
         let comparison =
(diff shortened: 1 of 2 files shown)


site/src/comparison.rs
```diff
@@ -186,28 +186,31 @@ pub enum Metric {
     Instructions,
     #[serde(rename = "cycles:u")]
     Cycles,
-    #[serde(rename = "faults")]
+    #[serde(rename = "faults:u")]
```
rylev, line 189: Should this be rename or alias?

rylev (commented):
I think there are certainly some metric types that we consider first class (e.g., instructions, max-rss, and cycles) while the others we don't normally have much logic associated with and it's fine to just pass around the string. It might make sense to limit the enum to the first class metric types and then have an `Other` variant for everything else

Kobzol:
Oh, I thought that we couldn't use that because we need something like that for the whole enum, but I can just probably use a deserializer function for the single variant and count on the fact that `serde` will try to deserialize it as the last resort. That should work, thanks, I'll try it.

Mark-Simulacrum:
A custom deserializer seems fine to me.

Kobzol:
I implemented it with a custom deserializer.

Kobzol:
We decided with @Mark-Simulacrum to just explicitly enumerate the set of known metrics for now and expand them in the future.

Mark-Simulacrum:
Hm, it looks like the database check is no longer succeeding, but it's also failing on master? Going to go ahead and remove it from the branch protection rule for now, and we can try to investigate and fix it in the future.

Mark-Simulacrum (approved):

PR #1348, opened 2022-06-21
Store artifact sizes into the database

Kobzol:
It would be really nice to track the size of generated artifacts (especially the size of binaries) over time. There are a lot of things that we could do in this direction, but probably the simplest one to begin with is to generate stats corresponding to the artifact sizes that are produced by the self profiler.

With this small change, we can start tracking artifact sizes for the currently existing benchmarks. This is just a small first step, of course, but it should already provide us with interesting data.

I used the `size:` prefix for the stats, to differentiate these values from the existing metrics. Maybe we could add some unified prefix to the compile time metrics? Later we can add yet another prefix for metrics gathered from compiled binaries.

Related issue: https://github.com/rust-lang/rustc-perf/issues/145

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -206,6 +206,18 @@ pub enum Metric {
     TaskClockUser,
     #[serde(rename = "wall-time")]
     WallTime,
+    #[serde(rename = "size:codegen_unit_size_estimate")]
+    CodegenUnitSize,
+    #[serde(rename = "size:dep_graph")]
+    DepGraphSize,
+    #[serde(rename = "size:linked_artifact")]
+    LinkedArtifactSize,
+    #[serde(rename = "size:object_file")]
+    ObjectFileSize,
+    #[serde(rename = "size:query_cache")]
+    QueryCacheSize,
+    #[serde(rename = "size:work_product_index")]
+    WorkProductIndexSize,
 }
 
 impl Metric {
@@ -222,6 +234,12 @@ impl Metric {
             Metric::TaskClock => "task-clock",
             Metric::TaskClockUser => "task-clock:u",
             Metric::WallTime => "wall-time",
+            Metric::CodegenUnitSize => "size:codegen_unit_size_estimate",
+            Metric::DepGraphSize => "size:dep_graph",
+            Metric::LinkedArtifactSize => "size:linked_artifact",
+            Metric::ObjectFileSize => "size:object_file",
+            Metric::QueryCacheSize => "size:query_cache",
+            Metric::WorkProductIndexSize => "size:work_product_index",
         }
     }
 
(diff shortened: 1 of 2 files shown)


site/src/comparison.rs
```diff
@@ -180,34 +185,39 @@ async fn populate_report(
...
-    Faults,
     #[serde(rename = "max-rss")]
     MaxRSS,
-    #[serde(rename = "task-clock")]
-    TaskClock,
-    #[serde(rename = "wall-time")]
-    WallTime,
-    #[serde(rename = "cpu-clock")]
-    CpuClock,
+    Custom(String),
```
Mark-Simulacrum, line 196: I think if we do this, I would like a comment or perhaps even a wrapper around the String in a private module that (tries to) guarantee we're not going to end up with `.to_string() == "size:some_custom_size"` down the line. I'd really rather avoid string-y typing for these metrics if we can, and it's not clear to me that an unbounded set of metrics is a good idea. If we can't justify the set of metrics enough to *name them*, then maybe they shouldn't be collected -- after all, we do pay some amount of overhead the larger the set is (in mental complexity at least, though infrastructure cost too).

Kobzol: I wanted to create explicit enum variants for enums used explicitly in code (currently cycles, instructions and max RSS) and use an opaque string wrapper for the rest. I use a string wrapper, because the metrics are coming from "unbounded" sources at this moment:
- Artifact sizes are read from a self-profile JSON, which can basically contain arbitrary data, as far as `rustc-perf` is concerned.
- Perf. metrics (with the exception of wall time, cycles, instructions and cpu-clock) are also read as an arbitrary string from the output of `perf`.

Things are also a bit more complicated by the fact that some metrics exist in two very similar variants (e.g. `cycles` and `cycles:u`), and even if we wanted to unify these somehow, they are stored in historical data, so we probably need to support them.

So I see three ways of resolving this:
- Use the proposed way of using an opaque wrapper for unknown metrics.
- Make a list of known metrics and ignore the rest.
- Make a list of known metrics and crash if an unknown one is encountered (in that case we should add it to the `Metric` enum).

Kobzol:
> It would be nice to have the stuff reported by size (.text, .data, .bss, total size), would that be difficult?

That would indeed be nice, but it delves into the territory of using special gathering/visualisation for size data. I could add code that would measure the text/data/bss sections individually, but this would create yet more metrics. Currently, this PR adds about 6-7 new `size:` metrics and the metric list in the compare page is getting a bit too unwieldy I think. Maybe we should somehow rethink the UI or create some separate categories for compile time/size/(runtime) metrics to distinguish these things better.

Kobzol:
I rebased this over https://github.com/rust-lang/rustc-perf/pull/1338 and added new explicit `size:` metrics.

Mark-Simulacrum (approved):

PR #1349, opened 2022-06-28
Exclude the rust src directory from inclusion in the rustc-perf workspace

Mark-Simulacrum:
This should fix the build error in the collector for the rustc benchmark, hopefully.

cc @jyn514

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 members = ["collector", "site", "database", "intern"]
-exclude = ["collector/benchmarks"]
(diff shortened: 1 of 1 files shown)


jyn514:
Do you have a link to the error? Why are we excluding src/ ?

Mark-Simulacrum:
https://perf.rust-lang.org/status.html has the error message; I decided a more blanket exclude is probably more future proof.

jyn514 (approved):

lqd:
Thanks for the fix, Mark 🙏

PR #1363, opened 2022-07-21
Rustup

oli-obk:
`rustup` seems to have changed its exit code, which broke rustc-perf for me


Kobzol:
This is the third PR that tries to fix this :joy: https://github.com/rust-lang/rustc-perf/pull/1361 https://github.com/rust-lang/rustc-perf/pull/1362 It was already fixed :)

PR #1366, opened 2022-07-23
Teach rustc-fake to tell the truth about subprocess exit statuses

yaahc:
This resolves the ci failure observed in https://github.com/rust-lang/rust/pull/99431

turns out the impl of `fn exec()` in rustc-perf's `rustc-fake` is incorrect on windows and only reports a failure when the subprocess fails to spawn, but if it spawns correctly it always reports success rather than checking the exit status of the subprocess. This causes the probe in anyhow to succeed and enable `#[cfg(backtrace)]` which then causes anyhow to fail to compile, but because it was inside of `rustc-fake` it reports success and continues! This happens quite a bit and we get some very exciting compiler errors when `RUST_LOG=collector=trace` is set.

### DIFF
--- a/collector/src/rustc-fake.rs
+++ b/collector/src/rustc-fake.rs
@@ -468,8 +468,7 @@ fn process_self_profile_output(prof_out_dir: PathBuf, args: &[OsString]) {
 fn exec(cmd: &mut Command) {
     let cmd_d = format!("{:?}", cmd);
     match cmd.status() {
-        Ok(status) if !status.success() => std::process::exit(1),
-        Ok(_) => {},
+        Ok(status) => std::process::exit(status.code().unwrap_or(1)),
         Err(e) => panic!("failed to execute `{}`: {}", cmd_d, e),
     }
 }
--- a/collector/src/rustc-fake.rs
+++ b/collector/src/rustc-fake.rs
@@ -467,8 +467,10 @@ fn process_self_profile_output(prof_out_dir: PathBuf, args: &[OsString]) {
 #[cfg(windows)]
 fn exec(cmd: &mut Command) {
     let cmd_d = format!("{:?}", cmd);
-    if let Err(e) = cmd.status() {
-        panic!("failed to execute `{}`: {}", cmd_d, e);
+    match cmd.status() {
+        Ok(status) if !status.success() => std::process::exit(1),
+        Ok(_) => {},
+        Err(e) => panic!("failed to execute `{}`: {}", cmd_d, e),
     }
 }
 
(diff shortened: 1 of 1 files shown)


collector/src/rustc-fake.rs
```diff
@@ -467,8 +467,10 @@ fn process_self_profile_output(prof_out_dir: PathBuf, args: &[OsString]) {
 #[cfg(windows)]
 fn exec(cmd: &mut Command) {
     let cmd_d = format!("{:?}", cmd);
-    if let Err(e) = cmd.status() {
-        panic!("failed to execute `{}`: {}", cmd_d, e);
+    match cmd.status() {
+        Ok(status) if !status.success() => std::process::exit(1),
+        Ok(_) => {},
+        Err(e) => panic!("failed to execute `{}`: {}", cmd_d, e),
```
yaahc, line 473: This might also be a good idea and a more accurate diagnostic

```suggestion
        Err(e) => panic!("failed to spawn `{}`: {}", cmd_d, e),
```

yaahc:
nika has also suggested we may also want to make it so `exec` never returns in any case, and change the return type of the fn to `!`

Mark-Simulacrum (approved):

collector/src/rustc-fake.rs
```diff
@@ -467,8 +467,10 @@ fn process_self_profile_output(prof_out_dir: PathBuf, args: &[OsString]) {
 #[cfg(windows)]
 fn exec(cmd: &mut Command) {
     let cmd_d = format!("{:?}", cmd);
-    if let Err(e) = cmd.status() {
-        panic!("failed to execute `{}`: {}", cmd_d, e);
+    match cmd.status() {
+        Ok(status) if !status.success() => std::process::exit(1),
+        Ok(_) => {},
```
eddyb, line 472:
```suggestion
        Ok(status) => std::process::exit(status.code().unwrap_or(1)),
```

FWIW ^^ that was @mystor's original suggestion, which would only use `1` when the status can't be converted into an `i32`.

collector/src/rustc-fake.rs
```diff
@@ -467,8 +467,9 @@ fn process_self_profile_output(prof_out_dir: PathBuf, args: &[OsString]) {
 #[cfg(windows)]
 fn exec(cmd: &mut Command) {
```
RalfJung, line 468: FWIW using return type `-> !` (like the Unix version) would have caught this

eddyb: Ahh I forgot that part when I posted @mystor's suggestion (https://github.com/rust-lang/rustc-perf/pull/1366#discussion_r928057956).

PR #1372, opened 2022-07-31
Individual builds for all rolled up PRs

rylev:
Every time bors merges a rollup PR, we will now create rustc artifacts for all of the individual PRs in the roll up.

This is accomplished by creating a merge commit between the master commit at the time the rollup was merged and the head of each rollup PR. Each of these merge commits is then pushed to the try-perf where a try build CI run will be kicked off. 

In and of itself, this does not help with roll up perf problems as this only builds the artifacts - it does not perform any performance run. After we have this working, we'll need a way to kick off perf runs for a particular rolled up PR.


site/src/github.rs
```diff
@@ -346,6 +426,30 @@ pub async fn create_commit(
...
+        .clone()
+        .expect("needs github API token");
+    let url = format!("{}/issues/{}", repository_url, number);
+    let response = client
+        .get(&url)
+        .header(USER_AGENT, "perf-rust-lang-org-server")
+        .basic_auth("rust-timer", Some(timer_token))
+        .send()
+        .await
+        .context("cannot get commit")?;
```
Mark-Simulacrum, line 448: Wrong context here, probably

site/src/request_handlers/github.rs
```diff
@@ -25,96 +27,197 @@ pub async fn handle_github(
...
+        return Ok(github::Response);
+    }
+
+    let previous_master = &push.before;
+    let rollup_merges = push
+        .commits
+        .iter()
+        .rev()
+        .skip(1) // skip the head commit
+        .take_while(|c| c.message.starts_with("Rollup merge of "))
```
Mark-Simulacrum, line 51: I think this probably should be a filter - the commits don't have a single linear ordering due to the merges, so I don't think we can rely on it being the first (or last) N.

Did you test this perhaps?

site/src/request_handlers/github.rs
```diff
@@ -25,96 +27,197 @@ pub async fn handle_github(
...
+    for rollup_merge in rollup_merges {
+        // Fetch the rollup merge commit which should have two parents.
+        // The first parent is in the chain of rollup merge commits all the way back to `previous_master`.
+        // The second parent is the head of the PR that was rolled up. We want the second parent.
+        let commit = get_commit(&client, &ctxt, repository_url, &rollup_merge)
+            .await
+            .map_err(|e| format!("Error getting rollup merge commit: {e:?}"))?;
+        assert!(
+            commit.parents.len() == 2,
+            "What we thought was a merge commit was not a merge commit"
```
Mark-Simulacrum, line 63: Please add the SHA of the commit to this assert.

site/src/request_handlers/github.rs
```diff
@@ -25,96 +27,197 @@ pub async fn handle_github(
...
+        .take_while(|c| c.message.starts_with("Rollup merge of "))
+        .map(|c| &c.sha);
+
+    for rollup_merge in rollup_merges {
+        // Fetch the rollup merge commit which should have two parents.
+        // The first parent is in the chain of rollup merge commits all the way back to `previous_master`.
+        // The second parent is the head of the PR that was rolled up. We want the second parent.
+        let commit = get_commit(&client, &ctxt, repository_url, &rollup_merge)
+            .await
+            .map_err(|e| format!("Error getting rollup merge commit: {e:?}"))?;
```
Mark-Simulacrum, line 60: Probably should include the SHA here too.

Mark-Simulacrum:
> After we have this working, we'll need a way to kick off perf runs for a particular rolled up PR.

The existing rust-timer build $SHA support should work okay for this, I think. We might want to have the automation added here leave a comment about the queued builds with the rollup PR to pushed commit mapping though, since it'll be annoying to get that otherwise - I think that's not yet implemented? (I may have missed it)

site/src/request_handlers/github.rs
rylev: I believe this is correct. The commits seem to always be (in reverse order) - the merge commit for the entire roll up, each individual merge commit for the rolled up PRs, and then the individual commits.

site/src/github.rs
```diff
@@ -32,42 +33,129 @@ pub async fn get_authorized_users() -> Result<Vec<usize>, String> {
...
+pub async fn rollup_pr(client: &client::Client, message: &str) -> Result<Option<u32>, String> {
+    if !message.starts_with("Auto merge of") {
+        return Ok(None);
+    }
+
+    let number = ROLLUP_PR_NUMBER
+        .captures(&message)
+        .and_then(|c| c.get(1))
+        .map(|m| {
+            println!("{}", m.as_str());
```
Mark-Simulacrum, line 102: Leftover print? We could log it, but that should go through tracing.

site/src/github.rs
```diff
@@ -32,42 +33,129 @@ pub async fn get_authorized_users() -> Result<Vec<usize>, String> {
...
+
+    Ok(issue
+        .labels
+        .iter()
+        .any(|l| l.name == "rollup")
+        .then(|| issue.number))
+}
+
 // Returns the PR number
 pub async fn pr_and_try_for_rollup(
```
Mark-Simulacrum, line 126: Presumably we'll be deleting this entirely soon, but that can be a follow-up PR.

Mark-Simulacrum (approved):
One small nit left

site/src/request_handlers/github.rs
Mark-Simulacrum: Ok, seems okay for now. Can always adjust in the future.

PR #1378, opened 2022-08-01
Process rollup in the background

rylev:
GitHub has a timeout of 10 seconds for its web hooks which does not seem configurable. This moves processing to the background. This might make debugging slightly more tedious, but other than that I don't think this is worse than the satus quo.


Mark-Simulacrum (approved):
Yeah, it does mean that we might get killed by ECS if there's a deployment in progress, but the overlap there is hopefully pretty small. We might want a command that manually triggers the logic so you can run it on a rollup if needed.

PR #1392, opened 2022-08-11
Add branch-misses and cache-misses to stats collected by `perf stat`

Kobzol:
These metrics are quite fundamental in my opinion, and I think that adding them to `perf stat` shouldn't really affect a lot of things, since it already gathers a lot of other stats. Even though these metrics will probably be quite noisy, I think that they can still be useful in the dashboard.

`branch-misses` and `cache-misses` are in fact aliases for `perf`, which chooses some inner metric to calculate these. I think that these alises should be available on the perf runner.

### DIFF
--- a/collector/src/rustc-fake.rs
+++ b/collector/src/rustc-fake.rs
@@ -83,7 +83,7 @@ fn main() {
                     .env("LC_NUMERIC", "C")
                     .arg("-x;")
                     .arg("-e")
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum (approved):
Confirmed availability on the runner.

I'll also note we have a finite amount of metrics available that require hardware counters (iirc, 5), but we're not hitting that limit quite yet -- only instructions, cycles, and now branch and cache misses are in that number I believe.

PR #1357, opened 2022-07-17
Explicitly mark commits as being try or master instead of relying on their date

Kobzol:
This should help with providing more accurate data in the dashboard and also enable seeing the 30 day metric graph for PRs.
I'm not sure of all the consequences of this change though.

PR #1393, opened 2022-08-11
Insert try commit date into DB

Kobzol:
This is a follow-up to https://github.com/rust-lang/rustc-perf/pull/1357, it looks like I missed one place where `try` commit dates were not being inserted into the database.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -760,11 +760,7 @@ where
         let (name, date, ty) = match artifact {
             ArtifactId::Commit(commit) => (
                 commit.sha.to_string(),
-                if commit.is_try() {
-                    None
-                } else {
-                    Some(commit.date.0)
-                },
+                Some(commit.date.0),
                 if commit.is_try() { "try" } else { "master" },
             ),
             ArtifactId::Tag(a) => (a.clone(), None, "release"),


rylev (approved):

site/src/request_handlers/graph.rs
```diff
@@ -140,9 +145,13 @@ async fn create_graphs(
     }))
 }
 
-fn artifact_ids_for_range(ctxt: &SiteCtxt, start: Bound, end: Bound) -> Vec<ArtifactId> {
+fn master_artifact_ids_for_range(ctxt: &SiteCtxt, start: Bound, end: Bound) -> Vec<ArtifactId> {
     let range = ctxt.data_range(start..=end);
-    range.into_iter().map(|c| c.into()).collect()
+    range
+        .into_iter()
+        .filter(|c| matches!(c.r#type, CommitType::Master))
```
Mark-Simulacrum, line 152: We probably want to explicitly include the endpoints (start/end) if they're commit SHAs, even if they are try commits. The comparison page generates links like https://perf.rust-lang.org/index.html?start=2022-06-12&end=b3f4c3119957aa0a250cab08ab586b7a9a680ef1&benchmark=ctfe-stress-5&profile=opt&scenario=full&stat=instructions:u

for which the end bound is the end commit on the compare page, and that can (often) be a try commit. It's not a big deal to drop it, but it seems annoying and probably not what we want.

Kobzol: Good catch, thanks. Hopefully it is fixed now.

PR #1394, opened 2022-08-11
Take all commits that look like rollup merges

### LINKED ISSUE #1390: GH perf bot message: number of perf builds in rollup reported by perf bot != PR's in rollup
https://github.com/rust-lang/rust/pull/100356 contains 8 PRs in rollup, but perf bot message talks about only one PR https://github.com/rust-lang/rust/pull/100356#issuecomment-1210363384

probably #1372 so @rylev ?

rylev:
Fixes #1390 

Previously the handler for unrolling rollups assumed that the commits were ordered bor's auto merge, all roll up merges, all individual commits (in reverse order), but this is not always true. In https://github.com/rust-lang/rust/pull/100356 for example, an individual commit made it in between roll up merge commits.

This simplifies the code to not make this assumption.

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -51,8 +51,7 @@ async fn handle_push(ctxt: Arc<SiteCtxt>, push: github::Push) -> ServerResult<gi
         let rollup_merges = commits
             .iter()
             .rev()
-            .skip(1) // skip the head commit
-            .take_while(|c| c.message.starts_with("Rollup merge of "));
+            .filter(|c| c.message.starts_with("Rollup merge of #"));
         let result = unroll_rollup(
             ci_client,
             main_repo_client,


site/src/request_handlers/github.rs
```diff
@@ -51,8 +51,7 @@ async fn handle_push(ctxt: Arc<SiteCtxt>, push: github::Push) -> ServerResult<gi
         let rollup_merges = commits
             .iter()
             .rev()
-            .skip(1) // skip the head commit
-            .take_while(|c| c.message.starts_with("Rollup merge of "));
+            .filter(|c| c.message.starts_with("Rollup merge of #"));
```
Mark-Simulacrum, line 54: I am a *little* worried about what happens when we rollup a subtree merge or something similar that itself contains Rollup merge ... -- particularly since today we suffer a complete queue stoppage if rustup-toolchain-install-master fails until the DB is manually cleaned.

But I think for the most part those PRs aren't rolled up, so this is probably OK for now. We can revisit and try to get some more advanced filtering at a later point if needed.

Mark-Simulacrum (approved):

PR #1400, opened 2022-08-12
Add quick links to important metrics to compare page

Kobzol:
This should make it easier to access common metrics, and it also increases the visibility of these three metrics, that we include in the PR summaries.

image

I'd like to experiment with further changes to the compare page, for that it's nice if we can finally split it into HTML/JS/CSS parts (I can split the first commit into a separate PR if you want).


Mark-Simulacrum (approved):

Mark-Simulacrum:
Did you check this locally to make sure the relative path works for loading the CSS + JS?

Kobzol:
I was developing and testing the UI locally, it worked fine. Relative path is already used for `shared.js`, so this should be the same.

PR #1403, opened 2022-08-15
Add buttons to move between categories quickly to compare page

Kobzol:
This is useful when the tables are large.

image


site/static/compare/script.js
```diff
@@ -363,8 +373,16 @@ app.component('test-cases-table', {
...
-<div class="category-title">{{ title }} benchmarks</div>
+<div class="bench-table" :id="id">
+<div class="category-title">
+  {{ title }} benchmarks
+  <span v-if="title == 'Primary'" title="To secondary benchmarks">
+    <a href="#secondary-benchmarks">⮯</a>
+  </span>
+  <span v-else title="To primary benchmarks">
+    <a href="#primary-benchmarks">⮭</a>
+  </span>
```
rylev, line 384: I don't love that we have somewhat complex logic based on the title. It makes this component harder to understand. Can we do something like this instead?

```suggestion
  <span :title="To {title} benchmarks">
    <a :href="#{title}-benchmarks">⮯</a>
  </span>
```

rylev (changes requested):

site/static/compare/script.js
```diff
@@ -363,8 +373,16 @@ app.component('test-cases-table', {
...
-<div class="bench-table">
-<div class="category-title">{{ title }} benchmarks</div>
+<div class="bench-table" :id="id">
+<div class="category-title">
+  {{ title }} benchmarks
+  <span v-if="title == 'Primary'" title="To secondary benchmarks">
+    <a href="#secondary-benchmarks">⮯</a>
+  </span>
+  <span v-else title="To primary benchmarks">
+    <a href="#primary-benchmarks">⮭</a>
```
rylev, line 383: I don't really love the pointer emojis, but I don't have a good suggestion otherwise...

site/static/compare/script.js
Kobzol: Ah, but then I would need to add yet another prop to say if button should point up or down :)

Kobzol:
I changed the code so that it doesn't depend on the title, it's not a bit complicated though.

Issue #1328, opened 2022-05-18
Additional details per metric


rylev:
In addition to the work being discussed in #1322, it might be good to include the following additional details for each of the metrics:
* **Number of uniquely impacted benchmarks**. We currently show the number of relevant test case results, but it's also relevant to know how widespread the issue is over the total surface area of the perf suite. For example, a perf run that has 12 total relevant test case results but that impact only 1 particular benchmark can be treated differently than if 12 different benchmarks are impacted.
* **Average percent change across the entire benchmark suite**. We currently show the average percent change for the relevant test case results, but obviously a +1% change impacting 5 test cases vs a 1% change impacting 40 test case results should be treated differently. The view is now required to piece this information together themselves. 
* **Average percent change per profile**. It would be good to know if one particular profile is more impacted than others. This could simply be an extra line of detail we display when one particular profile is > 75% (or some other percentage) of the impacted test case results. 
* **Average percent change per scenario**. Same as the per profile analysis

Kobzol:
I was thinking about this and I wonder if this all should rather be shown in the compare page? There's only so much that can fit inside the PR comments, and we have already added two new sections and tables there.

Kobzol:
I think that with the recent dashboard changes, this can be now closed?

rylev:
I agree!

PR #1404, opened 2022-08-15
Add summary table to compare page

Kobzol:
I moved the compare page JS, CSS and HTML to individual pages in https://github.com/rust-lang/rustc-perf/pull/1400 to make it easier to modify the compare page. It's still pretty annoying to do large changes, since it's written in JS (Typescript would be really nice!), but I think that we can do some incremental changes.

Inspired by https://github.com/rust-lang/rustc-perf/pull/1335, this PR adds a summary table component that is now used on the compare page instead of the previous `all`/`filtered` summary lines. I think that the `all`/`filtered` distinction was mostly useless and that it's much simpler to have a single summary + a `Reset filters` button, which this PR adds.

The new table contains min/max range and the number of unique benchmarks affected. I'm not sure which icon to use for the "all" (last) row, maybe some globe? Or no icon at all.

The first row shows regressions, the second row shows improvements, and the third row shows all changes (regressions + improvements).

As a follow-up, I would like to add a collapsible section with various axis aggregations, e.g. per-scenario, per-profile, per-category, and use this new summary table in the aggregations to easily display the results.

image

Related issue: https://github.com/rust-lang/rustc-perf/issues/1328

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -274,6 +274,7 @@
                     </div>
                     <input type="checkbox" v-model="showRawData" style="margin-left: 20px;" />
                 </div>
+                <button @click="resetFilter">Reset filters</button>
             </div>
         </fieldset>
         <p v-if="dataLoading && !data">Loading ...</p>
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -262,7 +262,7 @@
                             </span>
                         </span>
                     </div>
-                    <input type="checkbox" v-model="filter.showNonRelevant" style="margin-left: 20px;" />
+                    <input type="checkbox" v-model="filter.nonRelevant" style="margin-left: 20px;" />
                 </div>
                 <div class="section">
                     <div class="section-heading"><span>Display raw data</span>
@@ -278,48 +278,16 @@
         </fieldset>
         <p v-if="dataLoading && !data">Loading ...</p>
         <div v-if="data" id="content" style="margin-top: 15px">
-            <div id="summary">
-                <div style="display: flex; justify-content: end;">
+            <div id="main-summary">
+                <summary-table :cases="filteredSummary"></summary-table>
+                <div style="position: absolute; right: 5px; top: 5px;">
                     <span class="tooltip" style="margin-right: 1em;">?
                         <span class="tooltiptext">
-                            The percents show arithmetic mean amongst regressions, amongst improvements
-                            and finally amongst all benchmarks in each category (all benchmarks or
-                            filtered benchmarks).
+                            The table shows summaries of regressions, improvements and all changes
+                            calculated from data that is currently visible (data that passes the
+                            active filters).
                         </span>
                     </span>
-                </div>
-                <div v-for="summaryPair in Object.entries(summary)" class="summary-container">
-                    <span style="font-weight: bold; margin-left: 5px; text-transform: capitalize;">{{
-                        summaryPair[0] }}:</span>
-                    <div class="summary-values">
-                        <span class="summary summary-wide positive">
-                            {{summaryPair[1].regressions.toString().padStart(3, "&nbsp;")}}
-                            <svg style="width:18px;height:18px" viewBox="0 0 24 24">
-                                <path
-                                    d="M16,6L18.29,8.29L13.41,13.17L9.41,9.17L2,16.59L3.41,18L9.41,12L13.41,16L19.71,9.71L22,12V6H16Z">
-                                </path>
-                            </svg>
-                            &nbsp;(+{{(summaryPair[1].regressions_avg).toFixed(2)}}%)
-                        </span>
-                        <span class="summary">
-                            {{summaryPair[1].unchanged.toString().padStart(3, "&nbsp;")}}
-                            <svg style="width:18px;height:18px" viewBox="0 0 24 24">
-                                <path d="M22,12L18,8V11H3V13H18V16L22,12Z"></path>
-                            </svg>
-                        </span>
-                        <span class="summary summary-wide negative">
-                            {{summaryPair[1].improvements.toString().padStart(3, "&nbsp;")}}
-                            <svg style="width:18px;height:18px" viewBox="0 0 24 24">
-                                <path
-                                    d="M16,18L18.29,15.71L13.41,10.83L9.41,14.83L2,7.41L3.41,6L9.41,12L13.41,8L19.71,14.29L22,12V18H16Z">
-                                </path>
-                            </svg>
-                            &nbsp;({{(summaryPair[1].improvements_avg).toFixed(2)}}%)
-                        </span>
-                        <span class="summary" v-bind:class="percentClass(summaryPair[1].average)">
-                            &nbsp;{{ signIfPositive(summaryPair[1].average) }}{{ (summaryPair[1].average).toFixed(2) }}%
-                        </span>
-                    </div>
                 </div>
             </div>
             <div v-if="data.new_errors.length">
--- a/site/static/compare/script.js
+++ b/site/static/compare/script.js
@@ -282,6 +282,9 @@ const app = Vue.createApp({
 
             return createUrlFromParams(createSearchParamsForMetric(metric, start, end));
         },
+        resetFilter() {
+            this.filter = createDefaultFilter();
+        }
     },
 });
 
--- a/site/static/compare/script.js
+++ b/site/static/compare/script.js
@@ -6,6 +6,29 @@ function findQueryParam(name) {
     }
 }
 
+function createDefaultFilter() {
+    return {
+        name: null,
+        nonRelevant: false,
+        profile: {
+            check: true,
+            debug: true,
+            opt: true,
+            doc: true
+        },
+        scenario: {
+            full: true,
+            incrFull: true,
+            incrUnchanged: true,
+            incrPatched: true
+        },
+        category: {
+            primary: true,
+            secondary: true
+        }
+    };
+}
+
 const app = Vue.createApp({
     mounted() {
         const app = this;
@@ -20,26 +43,7 @@ const app = Vue.createApp({
     },
     data() {
         return {
-            filter: {
-                name: null,
-                showNonRelevant: false,
-                profile: {
-                    check: true,
-                    debug: true,
-                    opt: true,
-                    doc: true
-                },
-                scenario: {
-                    full: true,
-                    incrFull: true,
-                    incrUnchanged: true,
-                    incrPatched: true
-                },
-                category: {
-                    primary: true,
-                    secondary: true
-                }
-            },
+            filter: createDefaultFilter(),
             showRawData: false,
             data: null,
             dataLoading: false
@@ -103,7 +107,7 @@ const app = Vue.createApp({
                 let nameFilter = filter.name && filter.name.trim();
                 nameFilter = !nameFilter || name.includes(nameFilter);
 
-                const relevanceFilter = filter.showNonRelevant ? true : testCase.isRelevant;
+                const relevanceFilter = filter.nonRelevant ? true : testCase.isRelevant;
 
                 return (
                     profileFilter(testCase.profile) &&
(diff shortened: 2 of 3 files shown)


site/static/compare/script.js
```diff
@@ -433,6 +387,77 @@ app.component('test-cases-table', {
...
+      <th>Count</th>
+      <th><!-- icon --></th>
+    </thead>
+    <tbody>
+        <tr class="positive">
+            <td><SummaryRange :range="cases.regressions.range" /></td>
+            <td><SummaryPercentValue :value="cases.regressions.average" /></td>
+            <td><SummaryCount :cases="cases.regressions.count" :benchmarks="cases.regressions.benchmarks" /></td>
+            <td>
+              <svg style="width:16px;height:16px" viewBox="0 0 24 24">
```
rylev, line 436: Shall we switch to the emojis we use in the markdown tables? That will not only simplify this code but also make it easier to match between the two sources.

nnethercote: Yes please! ✅

nnethercote:
I haven't looked closely at the code changes because I don't know that code (or webdev stuff in general) well. But I like the visual change. Especially showing the number of results that didn't change. (Similar to what I tried [here](https://github.com/rust-lang/rustc-perf/pull/1365#issuecomment-1194933612). I think it makes sense for this summary and the GitHub comment summary to eventually look the same.)

If you copy and paste the new summary (e.g. into a GitHub comment) does it reproduce nicely?

nnethercote (approved):
Change the emoji like @rylev said and I think this is good to go, thanks!

site/static/compare/script.js
Kobzol: Good idea!

Kobzol:
> Especially showing the number of results that didn't change.

I'm a bit confused by this. Actually, the summary line before has showed the number of results that didn't change. Now the table instead shows regressions, improvements and all changes (so number of benchmarks that didn't change is **not** shown, although it can be calculated relatively easily by taking the third row of the third column and subtracting the first and second rows).

This is how the table looks on GitHub:

<table class="summary-table"><thead><th>Range</th><th>Mean</th><th>Count</th><th></th></thead><tbody><tr class="positive"><td><div> [<span>0.30%</span>, <span>2.78%</span>] </div></td><td><span>0.82%</span></td><td><span title="133 test case(s), 26 unique benchmark(s)">133 (26)</span></td><td title="Regresions">❌</td></tr><tr class="negative"><td><div> [<span>-0.30%</span>, <span>-26.83%</span>] </div></td><td><span>-2.48%</span></td><td><span title="269 test case(s), 44 unique benchmark(s)">269 (44)</span></td><td title="Improvements">✅</td></tr><tr><td><div> [<span>2.78%</span>, <span>-26.83%</span>] </div></td><td><span>-1.39%</span></td><td><span title="402 test case(s), 44 unique benchmark(s)">402 (44)</span></td><td></td></tr></tbody></table>

We can't play with CSS or color, but it doesn't look that bad. Of course it also doesn't contain the category distinction, so it's not a full replacement for the current PR summary table.

nnethercote:
> Actually, the summary line before has showed the number of results that didn't change.

I see. So there's room for confusion there, even among rustc-perf experts. Can you add a label for each row: "Worse", "Better", "All"? (Again, that's similar to what I did in #1365.) It's a shame there isn't an emoji that clearly means "All".

> This is how the table looks on GitHub:
> Range	Mean	Count	
> [0.30%, 2.78%]
> 	0.82%	133 (26)	❌
> [-0.30%, -26.83%]
> 	-2.48%	269 (44)	✅
> [2.78%, -26.83%]
> 	-1.39%	402 (44)	

We need to be careful with the ranges here.
- [0.30%, 2.78%]  The best (least regressed) result is listed first.
- [-0.30%, -26.83%]  The best (most improved) result is listed second.
- [2.78%, -26.83%]  The best (most improved or least regressed) result is listed second.

What I did in #1365 was to always put the smallest (most negative) one first and the largest (most positive) one second. This would change the ordering of the last two rows in the table, giving:
- [0.30%, 2.78%]
- [-26.83%, -0.30%]
- [-26.83%, 2.78%]

Kobzol:
> I see. So there's room for confusion there, even among rustc-perf experts. Can you add a label for each row: "Worse", "Better", "All"? (Again, that's similar to what I did in https://github.com/rust-lang/rustc-perf/pull/1365.) It's a shame there isn't an emoji that clearly means "All".

I was trying to avoid another label column to make the table smaller. It's not a problem in the main table, but then I want to include a lot of small tables for various aggregations (per-profile, per-category etc.). As an attempt to avoid such column, what about this?

<table class="summary-table"><thead><th>Range</th><th>Mean</th><th>Count</th><th></th></thead><tbody><tr class="positive"><td><div> [<span>0.30%</span>, <span>2.78%</span>] </div></td><td><span>0.82%</span></td><td><span title="133 test case(s), 26 unique benchmark(s)">133 (26)</span></td><td title="Regresions">❌</td></tr><tr class="negative"><td><div> [<span>-0.30%</span>, <span>-26.83%</span>] </div></td><td><span>-2.48%</span></td><td><span title="269 test case(s), 44 unique benchmark(s)">269 (44)</span></td><td title="Improvements">✅</td></tr><tr><td><div> [<span>2.78%</span>, <span>-26.83%</span>] </div></td><td><span>-1.39%</span></td><td><span title="402 test case(s), 44 unique benchmark(s)">402 (44)</span></td><td>❌+✅</td></tr></tbody></table>

> We need to be careful with the ranges here.

I did that on purpose, although I realize that it's a bit controversial. To me it made sense to read it as "[smallest regression, largest regression]" and "[smallest improvement, largest improvement]". But now I realize that this interpretation is not consistent with the third row, where it's "[smallest improvement, largest improvement]", but it's not clear why it's not the opposite order.

Your ordering is more consistent. @rylev what is your opinion about the range ordering?

nnethercote:
I now realize I don't understand the "Count" column. What does "133 (26)" mean?

"❌+✅" makes me think "regressions and improvements", i.e. excludes ones that didn't change.

Kobzol:
> I now realize I don't understand the "Count" column. What does "133 (26)" mean?

If you hover on it with a mouse, it has a title, but I realize that the presentation isn't ideal (I couldn't think of a better one without adding another column though). The number in the parentheses is the number of unique benchmarks changed/affected.

> "x+white_check_mark" makes me think "regressions and improvements", i.e. excludes ones that didn't change.

That's exactly what the data contains :) It's regressions + improvements, nothing more, nothing less. To include non-relevant results, you can click on "Show non-relevant results". Other than that, there's not really any other "didn't change" results.

nnethercote:
> > "x+white_check_mark" makes me think "regressions and improvements", i.e. excludes ones that didn't change.
> 
> That's exactly what the data contains :) It's regressions + improvements, nothing more, nothing less. To include non-relevant results, you can click on "Show non-relevant results". Other than that, there's not really any other "didn't change" results.

Above you said:

> Now the table instead shows regressions, improvements and all changes (so number of benchmarks that didn't change is not shown, although it can be calculated relatively easily by taking the third row of the third column and subtracting the first and second rows).

So I thought the last row was everything. But I see now that 133 + 269 = 402. I would rather see the range and mean for every result, rather than the range and mean of all the improvements and regressions. The former gives a good idea of the effect across all the benchmarks, while the latter often looks large but that's because only a small number of benchmarks are affected.

Kobzol:
Sorry, what I said before about the row substraction was completely wrong, I probably also got confused by my table :D The third row is indeed just regressions + improvements, so all data combined.

I wanted to make the table self-contained, so that it describes a specific dataset, without considering any other data (like non-relevant results). I think that this approach is correct and that the previous approach, which mixed information from data that was not presented (non-relevant results) was not intuitive.

The table now has the property that it shows regressions, improvements, and all changes amongst the currently filtered data. So the table summarizes **exactly** the data that is shown in the "changes" table/list below. With this property, there is a very easy way of displaying changes amongst all benchmarks - just check the "Show non-relevant results" checkbox. I think that this behaviour is much more intuitive, rather than mixing relevant and non-relevant data within the same table.

nnethercote:
If I click "show non-relevant results" the bottom row of the table will change. What about the other two rows? Will they change?

Kobzol:
Yes :) The table now always a summary of the currently filtered data (benchmark name, scenario, profile, category, relevance). So if you change the relevance filter, it will update the table.

Show non-relevant results: unchecked
```
<regressions taken from relevant data>
<improvemenets taken from relevant data>
<all changes taken from relevant data>
```

Show non-relevant results: checked
```
<regressions taken from all data>
<improvemenets taken from all data>
<all changes taken from all data>
```

rylev:
> what is your opinion about the range ordering?

I agree with @nnethercote that the order should be how you would see the numbers on a number line. 

> The table now always shows a summary of the currently filtered data (benchmark name, scenario, profile, category, relevance). So if you change the relevance filter, it will update the table.

I think this makes a lot of sense and intuitively makes sense. It summarizes the visible data.

site/static/compare.html
```diff
@@ -274,53 +274,21 @@ <h2>Comparing <span id="stat-header">{{stat}}</span> between <span id="before">{
...
-                <div style="display: flex; justify-content: end;">
+            <div id="main-summary">
+                <summary-table :cases="filteredSummary"></summary-table>
+                <div style="position: absolute; right: 5px; top: 5px;">
                     <span class="tooltip" style="margin-right: 1em;">?
                         <span class="tooltiptext">
-                            The percents show arithmetic mean amongst regressions, amongst improvements
-                            and finally amongst all benchmarks in each category (all benchmarks or
-                            filtered benchmarks).
+                            The table shows summaries of regressions, improvements and all changes
```
rylev, line 287: Perhaps we can make clear that this is all the *visible* data?

site/static/compare/script.js
```diff
@@ -6,6 +6,29 @@ function findQueryParam(name) {
     }
 }
 
+function createDefaultFilter() {
+    return {
+        name: null,
+        showNonRelevant: false,
```
rylev, line 12: nit: I think this should be `nonRelevant` as all of the others also have to do with what is "shown" but do not contain the "show" verb.

site/static/compare/script.js
```diff
@@ -433,6 +387,70 @@ app.component('test-cases-table', {
...
+
+app.component('summary-table', {
+    props: ['cases'],
+    template: `
+<table class="summary-table">
+    <thead>
+      <th>Range</th>
+      <th>Mean</th>
+      <th>Count</th>
+      <th><!-- icon --></th>
```
rylev, line 428: I personally don't find showing the icon last to be very intuitive. I would expect things that label the current row to be first, but I do understand the temptation to show the interesting data sooner to the user as they (presumably) read left to right.

site/static/compare/script.js
```diff
@@ -448,6 +466,12 @@ app.mixin({
             }
             return klass;
         },
+        signIfPositive(pct) {
```
rylev, line 469: Can this be removed? It doesn't look like it's being used anymore.

Kobzol: That's actually an oversight on my part, I removed it unadvertedly. I think that the + was useful visually, so I will revert it back, if you don't mind.

site/static/compare/script.js
Kobzol: image

I don't have a strong opinion here, but as you said, I wanted to keep the most important data first, at the left.

site/static/compare.html
Kobzol: Changed the help content to make this more clear (hopefully :) ).

Kobzol:
I reverted the range order and made a few visual changes based on received feedback. I also tried to align the number in the range so that in most cases the backets (`[]`) will line up nicely, and put back the plus sign for positive numbers.
image

rylev (approved):
Thanks for working on this! I think this is certainly an improvement over the status quo. I'm hesitant to send this back for another round of bike shedding when we can just get it into master and discuss small improvements as follow ups.

Kobzol:
Let's wait for @nnethercote, if he has more comments.

nnethercote:
>  image

On the last line I'd put the cross first, given that we generally show regressions first.

I like the addition of the `+` symbol everywhere.

Otherwise, this is good enough for me. Even if it's not perfect it'll be useful to try it out for a while, see how it feels in practice.

Kobzol:
I agree, let's try it. I squashed the commits.

site/static/compare/script.js
```diff
@@ -433,6 +387,86 @@ app.component('test-cases-table', {
...
+<table class="summary-table">
+    <thead>
+      <th><!-- icon --></th>
+      <th>Range</th>
+      <th>Mean</th>
+      <th>Count</th>
+    </thead>
+    <tbody>
+        <tr class="positive">
+            <td title="Regresions">❌</td>
```
lqd, line 448:
```suggestion
            <td title="Regressions">❌</td>
```

Kobzol: Oops :) Thanks. Will fix in https://github.com/rust-lang/rustc-perf/pull/1410 if it gets merged, otherwise I'll send a separate PR.

Issue #1406, opened 2022-08-16
try commit showing up on bootstrap dashboard


Mark-Simulacrum:
Looking at the [bootstrap](https://perf.rust-lang.org/bootstrap.html) dashboard, I'm seeing https://perf.rust-lang.org/compare.html?start=40336865fe7d4a01139a3336639c6971647e885c&end=b102bd401c3f4366afba3adb1b706cec1018f7c3 show up as the big spike there, which seems wrong -- that's a try commit.

Probably we need to adjust our filtering to exclude non-master commits from that graph.

cc @Kobzol

Mark-Simulacrum:
This may also be happening for the index.html page; see https://github.com/rust-lang/rust/pull/101524#issuecomment-1264487526 https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/try.20results.20as.20if.20merged

Mark-Simulacrum:
The commit in question is marked as try in the database:

```
rustc_perf=> select * from artifact where name like '3a818e7%';
  id   |                   name                   |          date          | type
-------+------------------------------------------+------------------------+------
 55280 | 3a818e795f2c0822fa14ed34745464da45dbeff3 | 2022-10-01 16:28:01+00 | try
```

PR #1407, opened 2022-08-16
Only display master commits in bootstrap chart

Kobzol:
Fixes: https://github.com/rust-lang/rustc-perf/issues/1406


site/src/request_handlers/bootstrap.rs
```diff
@@ -12,7 +12,11 @@ pub async fn handle_bootstrap(
 ) -> ServerResult<bootstrap::Response> {
     log::info!("handle_bootstrap({:?})", body);
     let range = ctxt.data_range(body.start.clone()..=body.end.clone());
-    let commits: Vec<ArtifactId> = range.iter().map(|c| c.clone().into()).collect();
+    let commits: Vec<ArtifactId> = range
+        .iter()
```
rylev, line 16: if this is changed to `into_iter`, can we get rid of the clone two lines below?

rylev (approved):

PR #1409, opened 2022-08-16
Make metric header not be "(          )" when there is no metric

### LINKED ISSUE #1405: Perf run tables contains empty brackets in titles for merged PRs
Example: https://github.com/rust-lang/rust/pull/100237#issuecomment-1216204983
There should be metrics like `(instructions:u)`, but it is empty (or brackets should be removed).

probably #1398, so @rylev

rylev:
Fixes #1405

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -639,7 +639,7 @@ pub fn write_summary_table(
     // This code attempts to space the table cells evenly so that the data is
     // easy to read for anyone who is viewing the Markdown source.
     let column_labels = [
-        format!("{metric}"),
+        metric,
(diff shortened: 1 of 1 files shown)


Kobzol (approved):

site/src/comparison.rs
```diff
@@ -640,7 +639,7 @@ pub fn write_summary_table(
     // This code attempts to space the table cells evenly so that the data is
     // easy to read for anyone who is viewing the Markdown source.
     let column_labels = [
-        format!("({metric})"),
+        format!("{metric}"),
```
Mark-Simulacrum, line 642: nit: can just be metric (no format!) I think?

PR #1410, opened 2022-08-16
Add data aggregations to compare page

Kobzol:
This PR is based on https://github.com/rust-lang/rustc-perf/pull/1404, and adds additional aggregations for profiles, scenarios and categories, so that users can quickly see the effect on the various subgroups.

The tables have legends (icons and header) removed to conserve space. This section is designed for experts, and the legend is already located in the main summary, so I think that this is OK.

image

### DIFF
--- a/site/static/compare/script.js
+++ b/site/static/compare/script.js
@@ -35,10 +35,13 @@ const app = Vue.createApp({
         loadState(state => makeData(state, app));
 
         document.getElementById("filters-toggle").onclick = (e) => {
-            toggleFilters("filters-content", "filters-toggle-indicator");
+            toggleSection("filters-content", "filters-toggle-indicator");
         };
         document.getElementById("search-toggle").onclick = (e) => {
-            toggleFilters("search-content", "search-toggle-indicator");
+            toggleSection("search-content", "search-toggle-indicator");
+        };
+        document.getElementById("aggregations-toggle").onclick = (e) => {
+            toggleSection("aggregations-content", "aggregations-toggle-indicator");
         };
     },
     data() {
@@ -420,7 +423,7 @@ const SummaryRange = {
 <div v-if="range.length > 0">
   [<SummaryPercentValue :value="range[0]" :padWidth="6" />, <SummaryPercentValue :value="range[1]" :padWidth="6" />]
 </div>
-<div v-else>-</div>
+<div v-else style="text-align: center;">-</div>
 `, components: {SummaryPercentValue}
 };
 const SummaryCount = {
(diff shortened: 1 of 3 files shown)


rylev (commented):
Interesting! I'm unsure... this is quite a bit of information, but it does make it easier to see than clicking the filters for each "aggregation" manually to see the results in the summary field.

I wouldn't mind trying this out. I definitely think this should come *after* the summary though.

Kobzol:
It's indeed intended as a tool for "advanced users", to gauge the effect of a change on specific subsets of data. For example, I often "eye the results" or click on various filters to see how much `doc` or `check` builds were affected, whether the change is only affecting `incr` builds etc. It's definitely not intended for most users.

I don't want it to steal away space on the main page, that's why it's collapsed by default. But if even that takes too much space, we can also move the aggregations section to the very bottom, below the changes table.

Kobzol:
This is how it looks like by default (collapsed):
image

nnethercote:
I like this.

If I click on the checkboxes in the "Filters" section to de-select things, are the relevant tables removed from the "Aggregations" section?

nnethercote (approved):

Kobzol:
Yes, the aggregation summaries are dynamically updated based on the active filters, same as the main summary.

rylev:
@Kobzol any thoughts to the following?

> I definitely think the aggregations should come *after* the summary though.

Kobzol:
By "summary", you mean the new summary table or the long list of benchmark changes (Primary benchmarks, Secondary benchmarks)? Do you still want to move the aggregations even though they are collapsed by default?

If yes, I don't have a problem with that. It will be a bit less discoverable, but it's for experts anyway, so that shouldn't be an issue.

rylev:
By "summary" I mean the summary table. Even though it is collapsed by default, it makes more sense to me that it would come after the section that is more commonly viewed. That way the information flows from most general to most specific as the user scrolls down the page. However, feel free to ignore me if you disagree.

site/static/compare/script.js
```diff
@@ -434,39 +437,107 @@ const SummaryCount = {
...
+  <div class="aggregation-section">
+    <div class="header">Scenario</div>
+    <div class="groups">
+      <div class="group" v-for="scenario in ['full', 'incr-full', 'incr-unchanged', 'incr-patched']">
+        <div class="group-header">{{ scenario }}</div>
+        <summary-table :summary="calculateSummary('scenario', scenario)" :withLegend="false"></summary-table>
+      </div>
+    </div>
+  </div>
+  <div class="aggregation-section">
```
rylev, line 517: Probably not worth it but we can reduce the amount of code here by making each aggregation section its own component.

rylev (approved):

PR #1417, opened 2022-08-19
Add a metric containing the size of generated documentation

Kobzol:
This week's triage report said something like "we don't track the size of generated documentation in rustc-perf". Well, I figured, why don't we? It seems easy enough.

CC @GuillaumeGomez @jsha Is this metric interesting for `rustdoc` developers? Currently I store the size of the whole `target/doc` directory, should we store something else?

### DIFF
--- a/collector/src/execute.rs
+++ b/collector/src/execute.rs
@@ -848,6 +848,12 @@ impl<'a> Processor for BenchProcessor<'a> {
                 if let Some(ref profile) = res.1 {
                     store_artifact_sizes_into_stats(&mut res.0, profile);
                 }
+                if let Profile::Doc = data.profile {
+                    let doc_dir = data.cwd.join("target/doc");
+                    if doc_dir.is_dir() {
+                        store_documentation_size_into_stats(&mut res.0, &doc_dir);
+                    }
+                }
 
                 match data.scenario {
                     Scenario::Full => {
@@ -892,6 +898,44 @@ impl<'a> Processor for BenchProcessor<'a> {
             }
         }
     }
+}
+
+fn store_documentation_size_into_stats(stats: &mut Stats, doc_dir: &Path) {
+    match get_file_count_and_size(doc_dir) {
+        Ok((count, size)) => {
+            stats.insert("size:doc_files_count".to_string(), count as f64);
+            stats.insert("size:doc_bytes".to_string(), size as f64);
+        }
+        Err(error) => log::error!(
+            "Cannot get size of documentation directory {}: {:?}",
+            doc_dir.display(),
+            error
+        ),
+    }
+}
+
+/// Counts the number of files and the total size of all files within the given `path`.
+/// File size is counted as the actual size in bytes, i.e. the size returned by
+/// [std::path::Path::metadata].
+///
+/// Returns (file_count, size).
+pub fn get_file_count_and_size(path: &Path) -> std::io::Result<(u64, u64)> {
+    let (count, size) = if path.is_dir() {
+        let mut file_count = 0;
+        let mut total_size = 0;
+        for entry in fs::read_dir(&path)? {
+            let path = entry?.path();
+            let (count, size) = get_file_count_and_size(&path)?;
+            file_count += count;
+            total_size += size;
+        }
+        (file_count, total_size)
+    } else if path.is_file() {
+        (1, path.metadata()?.len())
+    } else {
+        (0, 0)
+    };
+    Ok((count, size))
 }
 
 fn store_artifact_sizes_into_stats(stats: &mut Stats, profile: &SelfProfile) {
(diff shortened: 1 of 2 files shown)


collector/src/execute.rs
```diff
@@ -894,6 +900,17 @@ impl<'a> Processor for BenchProcessor<'a> {
     }
 }
 
+fn store_documentation_size_into_stats(stats: &mut Stats, doc_dir: &Path) {
+    match fs_extra::dir::get_size(doc_dir) {
```
Kobzol, line 904: We might also just copy-paste the `get_size` function from `fs_extra`, it's self-contained.

GuillaumeGomez:
It sounds like a very good idea! I think storing the number of file and also the mean (total size divided by number of files) would help us have a good idea of how things are evolving.

Kobzol:
I also thought about storing the number of files, that sounds useful.

But calculating some derived data (like the mean) is a bit out of scope I think. We already have ~20 metrics and I don't think that we should add derived calculations to them, it seems like we should add some other approach for that (maybe just calculating it in the dashboard). What do you think @Mark-Simulacrum?

GuillaumeGomez:
If we store both the number of files and the total size, there is no need to store the mean, it's easy to compute on demand.

Mark-Simulacrum:
Definitely agree with not storing it - and not quite sure it makes sense for the dashboard to compute, but obviously something that users can pretty easily do themselves.

I think number of files is unlikely to be too interesting (our benchmarks are largely constant so I'm not sure how often we're going to see a shift there), but storing it seems fine.

Kobzol:
@GuillaumeGomez Does `rustdoc` perform changes that vary the number of files (for a documented crate that stays the same) often enough that it's worth it to save the count?

collector/src/execute.rs
Mark-Simulacrum: I'm not sure on licensing if we copy/paste - let's try to just reimplement here. We should also have a comment describing whether we're collecting disk space size or "actual" size (i.e., does a 100 byte file take up 4KB on at least ext4)?

Kobzol: I think that we should collect the actual size to have precise measurements (otherwise small changes could be lost e.g. due to rounding to 4 KiB).

I reimplemented the function and added a comment that describes the size that is gathered. I also wrote a simple test for it.

Kobzol:
I added also file count tracking, it's simple enough since we already gather the file size. I'm not sure if the `size:` prefix makes sense for the file count.

collector/src/execute.rs
```diff
@@ -894,6 +900,44 @@ impl<'a> Processor for BenchProcessor<'a> {
...
+        ),
+    }
+}
+
+/// Counts the number of files and the total size of all files within the given `path`.
+/// File size is counted as the actual size in bytes, i.e. the size returned by
+/// [std::path::Path::metadata].
+///
+/// Returns (file_count, size).
+pub fn get_file_count_and_size(path: &Path) -> std::io::Result<(u64, u64)> {
```
Kobzol, line 922: This doesn't handle symlinks. Do `doc` directories contain symlinks? :thinking: @GuillaumeGomez

GuillaumeGomez:
> @GuillaumeGomez Does `rustdoc` perform changes that vary the number of files (for a documented crate that stays the same) often enough that it's worth it to save the count?

No but new types gets added in the std/core libraries. Considering the total number of files, it shouldn't matter much but it's better to be able from one look to tell why there is a small change in the total size.

site/src/comparison.rs
```diff
@@ -231,6 +231,12 @@ pub enum Metric {
     LlvmBitcodeSize,
     #[serde(rename = "size:llvm_ir")]
     LlvmIrSize,
+    /// Total bytes of a generated documentation directory
+    #[serde(rename = "size:doc_bytes")]
+    DocByteSize,
+    /// Number of files inside a generated documentation directory.
+    #[serde(rename = "size:doc_files")]
+    DocFilesSize,
```
rylev, line 239: `DocFilesCount` seems much clearer (especially in isolation)

I think `size:doc_files` should be find if we think `size:doc_file_count` is too long.

rylev (commented):
I have one small nit, but otherwise looks great!

site/src/comparison.rs
Kobzol: I agree, that's a better name, I have renamed it.

Kobzol:
@Mark-Simulacrum @GuillaumeGomez any further comments? :)

collector/src/execute.rs
GuillaumeGomez: It does for old macros but it's all I can think of.

GuillaumeGomez:
Looks good to me!

collector/src/execute.rs
Kobzol: It's probably not super important, if there won't be a gazillion of them. They probably point to files within the doc folder anyway and symlinks don't take a lot of space.

GuillaumeGomez: Yes, it's completely fine.

collector/src/execute.rs
```diff
@@ -894,6 +900,44 @@ impl<'a> Processor for BenchProcessor<'a> {
     }
 }
 
+fn store_documentation_size_into_stats(stats: &mut Stats, doc_dir: &Path) {
+    match get_file_count_and_size(doc_dir) {
+        Ok((count, size)) => {
+            stats.insert("size:doc_files".to_string(), count as f64);
```
rylev, line 906: I will say that I still feel like `size:doc_files_count` would be worth the long name... But this is not a blocking concern.

rylev (approved):
Looks good!

collector/src/execute.rs
Kobzol: Ok, I have renamed it :)

PR #1418, opened 2022-08-19
Add artifact size to compare page quick links

Kobzol:
Also adds a short summary of the metric. I don't think that it's obvious to most people what metric corresponds to the artifact (binary) size of the compiled crate (well, I got it wrong :D ), so we might as well add it to the quick links.

image

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -48,10 +48,11 @@
         <div class="quick-links">
             <div>Quick links:</div>
             <div v-for="metric in [
-                ['Instructions', 'instructions:u'],
-                ['Cycles', 'cycles:u'],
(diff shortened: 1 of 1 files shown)


Kobzol:
Artifact is a bit general, so maybe it could be something like "Binary size"?

rylev:
@Kobzol Artifact size is already being used to track the sizes of all build artifacts (e.g., incremental compilation caches). If this is just the end binary size, I agree that we should make this clearer.

Kobzol:
The "technical" term is "linked artifact", I think that binary size is a friendly enough name for this. I changed the name of the quick link to `Binary size`.

rylev (approved):

Issue #1419, opened 2022-08-22
Dashboard page is out-of-date again


kamulos:
See #771 

The newest release 1.63.0 and the newest beta is not available in the https://perf.rust-lang.org/dashboard.html page

Mark-Simulacrum:
OK -- I would like to see an automated fix for this. I think @Kobzol you recently adjusted our next_commit code to take a proper commit, maybe you'd be interested in taking on adjusting that to take an ArtifactId or so? That way the backend can just use the normal queue rather than ad-hoc manual queueing.

Once that's done I think the best bet is to do something similar to the way we discover master commits, but use https://static.rust-lang.org/manifests.txt instead, putting the last beta and last stable release into the queue unless they've already been benchmarked (I don't think we need to backfill so that would be enough).

PR #1420, opened 2022-08-22
Add support for benchmarking published artifacts in `BenchNext`

Kobzol:
https://static.rust-lang.org/manifests.txt is used to download all stable and beta releases. Those that are not yet in DB are then returned from the `next_artifact` endpoint and benchmarked in `BenchNext`.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1419

### DIFF
--- a/collector/src/api.rs
+++ b/collector/src/api.rs
@@ -1,9 +1,9 @@
-pub mod next_commit {
-    use database::Commit;
+pub mod next_artifact {
+    use database::ArtifactId;
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-    pub struct NextCommit {
-        pub commit: Commit,
+    pub struct NextArtifact {
+        pub artifact: ArtifactId,
         pub include: Option<String>,
         pub exclude: Option<String>,
         pub runs: Option<i32>,
@@ -11,6 +11,6 @@ pub mod next_commit {
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
     pub struct Response {
-        pub commit: Option<NextCommit>,
+        pub artifact: Option<NextArtifact>,
     }
 }
--- a/collector/src/api.rs
+++ b/collector/src/api.rs
@@ -1,12 +1,15 @@
 pub mod next_artifact {
-    use database::ArtifactId;
+    use database::Commit;
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-    pub struct NextArtifact {
-        pub artifact: ArtifactId,
-        pub include: Option<String>,
-        pub exclude: Option<String>,
-        pub runs: Option<i32>,
+    pub enum NextArtifact {
+        Release(String),
+        Commit {
+            commit: Commit,
+            include: Option<String>,
+            exclude: Option<String>,
+            runs: Option<i32>,
+        },
     }
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
--- a/collector/src/main.rs
+++ b/collector/src/main.rs
@@ -2,8 +2,9 @@
 
 use anyhow::{bail, Context};
 use clap::Parser;
+use collector::api::next_artifact::NextArtifact;
 use collector::category::Category;
-use database::{ArtifactId, Commit, CommitType};
+use database::{ArtifactId, Commit, CommitType, Pool};
 use log::debug;
 use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use std::collections::HashMap;
@@ -834,7 +835,7 @@ enum Commands {
         self_profile: SelfProfileOption,
     },
 
-    /// Benchmarks the next commit for perf.rust-lang.org
+    /// Benchmarks the next commit or release for perf.rust-lang.org
     BenchNext {
         /// Site URL
         site_url: String,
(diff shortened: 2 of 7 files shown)


collector/src/main.rs
```diff
@@ -995,113 +996,66 @@ fn main_result() -> anyhow::Result<i32> {
...
-
-            let mut benchmarks = get_benchmarks(
-                &benchmark_dir,
-                next.include.as_deref(),
-                next.exclude.as_deref(),
-            )?;
-            benchmarks.retain(|b| b.category().is_primary_or_secondary());
+            if let NextArtifact {
+                artifact: ArtifactId::Tag(tag),
+                ..
```
Mark-Simulacrum, line 1017: We should match on the enum/struct to avoid silently ignoring future artifact IDs if we add them.

site/src/load.rs
```diff
@@ -213,6 +215,45 @@ impl SiteCtxt {
...
+        for line in artifact_list.lines() {
+            if let Some(artifact) = parse_published_artifact_tag(&line) {
+                if !tested_artifacts.contains(artifact.as_str())
+                    && !in_progress_artifacts.contains(&artifact)
+                {
+                    artifacts.push(artifact);
+                }
+            }
+        }
+        Ok(artifacts)
```
Mark-Simulacrum, line 254: This should probably take only the last N artifacts we plan to consider, to avoid trying to benchmark 1.0 or similar (which won't work).

Kobzol: I tried it locally and it starts with `beta-2016-03-23` :) And benchmarking that indeed doesn't work. I'll limit it to last 10 entries.

Kobzol:
I changed the code to take last `10` not yet tested artifacts, but I realized that if some of them are currently missing in the DB, they will be benchmarked now and that could fail if they are too old.

So maybe we should just take ~50 newest entries, regardless whether they are already tested or not?

collector/src/main.rs
```diff
@@ -995,113 +996,67 @@ fn main_result() -> anyhow::Result<i32> {
...
-            let mut benchmarks = get_benchmarks(
-                &benchmark_dir,
-                next.include.as_deref(),
-                next.exclude.as_deref(),
-            )?;
-            benchmarks.retain(|b| b.category().is_primary_or_secondary());
+            match next {
+                NextArtifact {
+                    artifact: ArtifactId::Tag(tag),
+                    ..
```
rylev, line 1018: Does it make sense to model `NextArtifact` as an enum instead of a struct given that it seems that the  include,  exclude, and runs fields only make sense for `NextArtifact`s that have commit `ArtifactId`s.

Kobzol: You're right, I was just lazy to do it before :sweat_smile: I changed it now.

site/src/load.rs
```diff
@@ -244,13 +244,16 @@ impl SiteCtxt {
...
-                    artifacts.push(artifact);
-                }
+        for artifact in artifact_list
+            .lines()
+            .rev()
+            .filter_map(parse_published_artifact_tag)
+            .take(20)
+        {
+            if !tested_artifacts.contains(artifact.as_str())
+                && !in_progress_artifacts.contains(&artifact)
```
rylev, line 254: Can this be moved to a `filter` call? then the whole loop would just be a iterator.

collector/src/api.rs
```diff
@@ -1,16 +1,19 @@
...
+            include: Option<String>,
+            exclude: Option<String>,
+            runs: Option<i32>,
+        },
     }
 
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
     pub struct Response {
-        pub commit: Option<NextCommit>,
+        pub artifact: Option<NextArtifact>,
```
rylev, line 17: Just a note: we've been careful about changing the API since we don't have the benefit of a compiler to tell us if all the users of this API have been properly updated. Are we sure that all consumers are aware of this change?

collector/src/main.rs
```diff
@@ -995,113 +996,64 @@ fn main_result() -> anyhow::Result<i32> {
             bench_rustc,
             self_profile,
         } => {
-            println!("processing commits");
+            println!("processing artifacts");
             let client = reqwest::blocking::Client::new();
-            let response: collector::api::next_commit::Response = client
-                .get(&format!("{}/perf/next_commit", site_url))
+            let response: collector::api::next_artifact::Response = client
+                .get(&format!("{}/perf/next_artifact", site_url))
```
rylev, line 1002: Just want to double check - by the time this code runs, the API has been guaranteed to be updated as well, right?

site/src/load.rs
```diff
@@ -213,6 +215,51 @@ impl SiteCtxt {
         )
     }
 
+    /// Returns the not yet tested published artifacts, sorted from newest to oldest.
+    pub async fn missing_artifacts(&self) -> anyhow::Result<Vec<String>> {
```
rylev, line 219: I believe this name is incorrect. "Artifacts" refers to any version of rustc whether identified by tag or commit. But this function just returns the untested released artifacts. So maybe `missing_released_artifacts` is more accurate?

site/src/load.rs
```diff
@@ -213,6 +215,51 @@ impl SiteCtxt {
...
+
+        lazy_static! {
+            static ref VERSION_REGEX: Regex = Regex::new(r"(\d+\.\d+.\d+)").unwrap();
+        }
+
+        let conn = self.conn().await;
+
+        let index = self.index.load();
+        let tested_artifacts: HashSet<_> = index.artifacts().collect();
+        let in_progress_artifacts: HashSet<_> = conn
```
rylev, line 233:
```suggestion
        let in_progress_tagged_artifacts: HashSet<_> = conn
```

site/src/load.rs
```diff
@@ -213,6 +215,51 @@ impl SiteCtxt {
...
+    pub async fn missing_artifacts(&self) -> anyhow::Result<Vec<String>> {
+        let artifact_list: String = reqwest::get("https://static.rust-lang.org/manifests.txt")
+            .await?
+            .text()
+            .await?;
+
+        lazy_static! {
+            static ref VERSION_REGEX: Regex = Regex::new(r"(\d+\.\d+.\d+)").unwrap();
+        }
+
```
rylev, line 228:
```suggestion
```

site/src/request_handlers/next_artifact.rs
```diff
@@ -1,9 +1,22 @@
...
 
 use std::sync::Arc;
 
-pub async fn handle_next_commit(ctxt: Arc<SiteCtxt>) -> next_commit::Response {
+pub async fn handle_next_artifact(ctxt: Arc<SiteCtxt>) -> next_artifact::Response {
+    // Prefer benchmarking released artifacts first
+    match ctxt.missing_artifacts().await {
+        Ok(next_artifact) => {
+            if let Some(next_artifact) = next_artifact.into_iter().next() {
+                log::debug!("next_artifact: {next_artifact}");
```
rylev, line 11: This debug statement doesn't really mean much outside the context of the code. perhaps we can add a bit more context to this?

collector/src/api.rs
Kobzol: @Mark-Simulacrum Are there any others notable users of this API than the perf. running `bench_next` script? FWIW, the API was already recently changed in https://github.com/rust-lang/rustc-perf/pull/1399.

collector/src/main.rs
Kobzol: That's a question for @Mark-Simulacrum I suppose. But recently we have changed the API and it was fine. The bench runner is updated before running a benchmark, so that should be fine, I'm not sure how it's synchronized with also updating https://perf.rust-lang.org/ though.

Kobzol: That's a question for @Mark-Simulacrum I suppose. But recently we have changed the API and it was fine. The bench runner is updated before running a benchmark, so that should be fine, I'm not sure how it's synchronized with also updating https://perf.rust-lang.org/ though.

site/src/load.rs
Kobzol: Good point.

site/src/request_handlers/next_artifact.rs
Kobzol: Well, it was inspired by a similar comment in the same function :sweat_smile::
```rust
log::debug!(
            "next_commit: {} (missing: {})",
            commit.sha,
            missing_reason_dbg
        );
```
What additional context should be included?

site/src/load.rs
Kobzol: I don't understand this comment, the suggestion looks like I should delete these lines?

site/src/load.rs
```diff
@@ -250,6 +297,32 @@ impl SiteCtxt {
     }
 }
 
+/// Parses an artifact tag like `1.63.0` or `beta-2022-08-19` from a line taken from
+/// `https://static.rust-lang.org/manifests.txt`.
+fn parse_published_artifact_tag(line: &str) -> Option<String> {
```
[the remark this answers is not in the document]
Kobzol: Good idea, I'll add some tests.

collector/src/api.rs
Mark-Simulacrum: No, and I'm not worried about this API changing. It's not really useful unless you're trying to 1:1 match what the perf collector is doing and even then it's kind of a weird tool to use (compared to running your own database and frontend).

collector/src/main.rs
Mark-Simulacrum: Yeah, there's not technically a guarantee here - but in practice it's only a few minutes difference (the collector will pull the git repo while the frontend waits for the docker build to finish in CI on master, and then gets a deployment on the new site). But in practice that just means a brief time when the collector won't work potentially, which is fine.

site/src/load.rs
rylev: Sorry for not being clear. I believe this statis is not being used so it can be deleted.

Kobzol: Oh, of course! :facepalm: My bad, sorry.

Mark-Simulacrum (approved):

PR #1421, opened 2022-08-24
Do not return NaN or Infinity significance factors

Kobzol:
The new artifact size results are so stable that their significance thresholds are very low. This creates some numerical problems that can be seen [here](https://perf.rust-lang.org/compare.html?start=25ea5a36c6e3a4fa1f739538d9e89eb1cd747564&end=e2e3da0be1bf838c7b3199740a227236918a5a48&stat=size%3Alinked_artifact).

Here for the first 7 results, the significance factor is Infinity, while for the last two it's NaN.

This PR changes the factor calculation so that it turns NaN and Infinity into a zero (the `is_finite` call handles both NaNs and Infinity, or so it claims). It also removes the `Option`, since the function always returned `Some` and there is no other usage of it anywhere.

One problem remains though, the `is_relevant` calculation would still consider these 9 results to be relevant, even though the significance factor would be zero, since it only compares the relative change to the significance threshold, which is low. Should I add a special case that forcefully considers the benchmark to not be relevant if the significance factor is zero?

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -1154,7 +1154,7 @@ impl TestResultComparison {
         !self.is_regression()
     }
 
-    /// Whther the comparison yielded a statistically significant result
+    /// Whether the comparison yielded a statistically significant result
     pub fn is_significant(&self) -> bool {
         self.relative_change().abs() >= self.significance_threshold()
     }
@@ -1168,11 +1168,17 @@ impl TestResultComparison {
     }
 
     /// This is a numeric magnitude of a particular change.
-    fn significance_factor(&self) -> Option<f64> {
+    fn significance_factor(&self) -> f64 {
         let change = self.relative_change();
         let threshold = self.significance_threshold();
-        // How many times the treshold this change is.
-        Some(change.abs() / threshold)
+
+        // How many times the threshold this change is.
+        let factor = change.abs() / threshold;
+        if factor.is_finite() {
+            factor
+        } else {
+            0.0
+        }
     }
 
     /// Whether the comparison is relevant or not.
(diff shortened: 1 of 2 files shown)


site/src/comparison.rs
```diff
@@ -1168,11 +1168,17 @@ impl TestResultComparison {
...
-        // How many times the treshold this change is.
-        Some(change.abs() / threshold)
+
+        // How many times the threshold this change is.
+        let factor = change.abs() / threshold;
+        if !factor.is_finite() {
+            0.0
+        } else {
+            factor
+        }
```
rylev, line 1181:
```suggestion
        if factor.is_finite() {
            factor
        } else {
            0.0
        }
```

rylev (commented):
@Kobzol the `is_relevant` function also looks at magnitude. Won't that already cover very small changes?

rylev:
@Kobzol that makes sense (at least when the significance factor is `NaN`) because magnitude is partially based on significance factor. [See here](https://github.com/rust-lang/rustc-perf/blob/ebd2d1d7f4d891990e3b121fbba575877e1d4de2/site/src/comparison.rs#L1191-L1192). When significance factor is `NaN`, all changes will be considered `VeryLarge`. 

However, when significance factor is `INFINITY`, it seems like the magnitude should always be `VerySmall`. 

Changing to a significance factor of `0.0` mirrors what we see with `NaN` which might be what we want. After all if the significance threshold is `0.0` then presumably that means, this metric almost *never* changes, so shouldn't it be considered relevant when it actually does?

Kobzol:
Hmm, in theory yes, but it's quite weird to see a result being relevant when it's change is `0.00 %`. If e.g. the binary size of `ripgrep` always stays the same, and then suddenly it changes by `0.0001%`, is it relevant? I don't know.

rylev:
Since we only show figures that are bigger than 0.01%, we could hard code the `is_relevant` to false when the change is less than 0.01%. That would eliminate this issue. 

Or we change the `magnitude` method to the following:

```rust
 pub fn magnitude(&self) -> Magnitude {
        let change = self.relative_change().abs();
       if change < 0.0001 {
            return Magnitude::VerySmall;
      } 
```

Kobzol:
I agree, that seems like a good solution. I changed it.

rylev (approved):

Issue #302, opened 2018-11-30
Feature Request: Export comparison page to Markdown table


michaelwoerister:
Since comparison URLs have an expiration date (right?) it would be nice if there was an easy way of copying the data over into a GH issue -- for example a button that shows a Markdown version of the table.

Alternatively, comparisons could be marked as permanent or something. E.g. via a `permalink` button.

PR #1429, opened 2022-08-26
Add Markdown export button to compare page

Kobzol:
This PR adds a button to export the (currently filtered) data from the compare page to Markdown (it will be downloaded as a Markdown file). The Markdown file will contain the summary of all filtered results and the two changes tables (for primary and secondary benchmarks). It can be useful e.g. to paste this data to GitHub comments.

I have put the button to the Filters section, because:
1) It is opened by default
2) Here it won't eat away vertical space from the rest of the page
3) It exports the currently filtered data

But it can be moved pretty much anywhere.

image

<details>
<summary>Example exported Markdown tables</summary>

# Summary
| | Range | Mean | Count |
|:---:|:---:|:---:|:---:|
| Regressions | 0.30%, 2.78% | 0.82% | 133 |
| Improvements | -26.83%, -0.30% | -2.48% | 269 |
| All | -26.83%, 2.78% | -1.39% | 402 |

# Primary benchmarks
| Benchmark | Profile | Scenario | % Change | Significance Factor |
|:---:|:---:|:---:|:---:|:---:|

# Secondary benchmarks
| Benchmark | Profile | Scenario | % Change | Significance Factor |
|:---:|:---:|:---:|:---:|:---:|
| match-stress-enum | check | full | -26.83% | 134.17x
| match-stress-enum | check | incr-full | -26.00% | 130.00x
| match-stress-enum | debug | full | -25.20% | 126.01x
| match-stress-enum | opt | full | -25.09% | 125.46x
| match-stress-enum | debug | incr-full | -24.33% | 121.63x
| match-stress-enum | opt | incr-full | -24.25% | 121.24x
| deeply-nested-async | debug | full | -12.84% | 64.22x
| deeply-nested-async | opt | full | -12.29% | 61.44x
| deeply-nested-async | debug | incr-full | -12.15% | 60.74x
| syn | opt | incr-unchanged | -10.06% | 50.29x
| diesel | doc | full | -9.48% | 47.42x
| deeply-nested-async | opt | incr-full | -7.88% | 39.41x
| externs | debug | full | -7.16% | 35.81x
| externs | opt | full | -7.12% | 35.61x
| externs | debug | incr-unchanged | -6.93% | 34.65x
| externs | opt | incr-unchanged | -6.93% | 34.65x
| externs | debug | incr-full | -6.39% | 31.96x
| externs | opt | incr-full | -6.36% | 31.81x
| webrender-wrench | doc | full | -5.47% | 27.33x
| tokio-webpush-simple | doc | full | -5.03% | 25.13x
| token-stream-stress | doc | full | -4.93% | 24.66x
| issue-46449 | doc | full | -4.87% | 24.35x
| regression-31157 | doc | full | -4.84% | 24.21x
| syn | opt | full | -4.81% | 24.03x

</details>

Closes: https://github.com/rust-lang/rustc-perf/issues/302

### DIFF
--- a/site/static/compare/script.js
+++ b/site/static/compare/script.js
@@ -287,6 +287,9 @@ const app = Vue.createApp({
         },
         resetFilter() {
             this.filter = createDefaultFilter();
+        },
+        exportToMarkdown() {
+            exportToMarkdown(this.testCases);
         }
     },
 });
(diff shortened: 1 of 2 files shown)


site/static/compare/script.js
```diff
@@ -709,4 +712,65 @@ function submitSettings() {
     window.location.search = params.toString();
 }
 
+function exportToMarkdown(testCases) {
```
rylev, line 715: I know it might be more of a pain to add, but it feels like this should be done in Rust and exposed over an endpoint.

rylev (changes requested):

site/static/compare/script.js
Kobzol: Hmm, I'm not sure what benefits would that have :thinking: Both the summary table and the changes table are modeled after the compare page, which is presented in JavaScript.

What would be the motivation for this?

rylev: You have a point - we're already splitting the logic for what summary metrics we show across JS and Rust (i.e., mean, range, count). It would be nice if all of this lived in one place, but this PR doesn't necessarily introduce that issue. And I suppose we can't really fix that without doing server-side rendering.

rylev (approved):

site/static/compare/script.js
Kobzol: It might be useful to make the summary table on the compare page and in PR comments should be as similar as possible, although I don't think that it's strictly necessary, the comment should contain (a possibly full) subset of the information on the compare page, since that should be more detailed.

Regarding the implementation, since the tables mostly differ in their presentation, I don't think that even backend rendering would solve this. Even if we generated some summary struct in Rust that would be used to render PR comments, it would then need to be rendered in various ways (HTML/CSS, Markdown). The closest thing to unify everything would probably be to generate Markdown on the server and then just render it as HTML on the compage page, but then we would lose a lot of the customization that we have there (colors, custom presentation etc.).

It's a good point, since this adds another place that needs to be changed if we modify the tables, but I consider this button to be mostly a helper tool for PR authors/experts, that will probably be hand-modified anyway.

For now I think that it's ok to leave it like this, to avoid complicating the code a lot because of a rather simple feature.

PR #1431, opened 2022-08-28
Add hardware, software and configuration details about the CI runner to documentation

Kobzol:
Data provided by @Mark-Simulacrum.


collector/perf-runner.md
```diff
@@ -0,0 +1,61 @@
...
+Enabled.
+
+### HyperThreading
+Enabled.
+
+### CPU scaling
+Scaling governor set to `performance`.
+
+### ASLR
+Disabled with `kernel.randomize_va_space=0` in `sysctl.conf`.
```
Mark-Simulacrum, line 55: I think this only disables kernel ASLR; we need the setarch in rustc-fake to disable it in userspace.

Mark-Simulacrum (approved):

collector/perf-runner.md
```diff
@@ -0,0 +1,61 @@
...
+    </details>
+- 64 GiB RAM, 32 GiB of swap
+
+## Software
+- Kernel: `4.15.0-191-generic`
+- OS: Ubuntu `18.04`
+
+## Configuration
+### Turboboost
+Enabled.
```
klensy, line 46: So, this probably provides some degree of instability of perf run results.
https://llvm.org/docs/Benchmarking.html

collector/perf-runner.md
```diff
@@ -0,0 +1,61 @@
+# CI benchmarking machine
+The machine that actually executes the benchmarks in CI has the following configuration.
+
+## Hardware
+- 6-core AMD Ryzen 5 3600 with HyperThreading (12 hardware threads total)
+    <details>
+    <summary>Output of `lscpu`</summary>
+    
+    ```
+    rchitecture:        x86_64
```
klensy, line 10:
```suggestion
    Architecture:        x86_64
```

collector/perf-runner.md
Kobzol: Indeed it does. If we turn it off, it will probably make the benchmarks slower though, as always, it's a tradeoff.

Compile time benchmarks are relatively long, so maybe this effect can even out somehow, but for runtime benchmarks, which will probably be shorter, we might rethink this decision.

collector/perf-runner.md
Kobzol: I think that it's a global setting for all programs (https://renenyffenegger.ch/notes/Linux/fhs/proc/sys/kernel/randomize_va_space, https://docs.oracle.com/cd/E37670_01/E36387/html/ol_aslr_sec.html).

PR #1436, opened 2022-08-30
Migrate from `winapi` to `windows_sys`

sn99:
1. [Winapi is no longer actively](https://github.com/retep998/winapi-rs/issues/867 ) maintained (last commit was 9 months ago as of now and 2nd last around 2 years ago) compared to last commit to [windows-rs ](https://github.com/microsoft/windows-rs)as of 3 days ago.
2. [windows-rs](https://github.com/microsoft/windows-rs) is the official Rust crate for Windows by Microsoft and is supposed to be providing a natural and idiomatic way for Rust developers to call Windows APIs. Some crates like [socket2](https://github.com/rust-lang/socket2) have already migrated to it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -303,7 +303,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "walkdir",
- "winapi",
+ "windows-sys",
  "xz2",
 ]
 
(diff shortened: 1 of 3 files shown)


Kobzol:
Hi, thanks for the PR! Since the dependency is only used for a single error code, I wonder if we even need it and if we can't just inline the error code.

In any case, could you please revert the formatting changes? They are not relevant for this PR. Otherwise LGTM.

sn99:
Alright, I will do that.

Should I make another pull related to `grammar and formatting fixes` ?

Kobzol:
We would ideally want to have just the first commit, if you send another commit that reverts the second one, that just creates additional churn :) Here's probably the easiest way how to do it:
```
# switch to your branch
$ git checkout master
# reset to the first commit
$ git reset --hard dd743f9
# force push
$ git push --force
```

> Should I make another pull related to grammar and formatting fixes ?

I think that the formatting is fine as it is. But if you found any typos, feel free to send another PR! :)

sn99:
Ah nice, done

Kobzol (approved):
Thanks!

PR #1441, opened 2022-09-01
Add the option to ignore certain profiles in `perf-config.json`

Kobzol:
This is useful for benchmarks that only care about a narrow metric (like the resulting binary size). With this setting, they can ignore e.g. check and doc builds to:
1) Save perf.rlo time from doing non-interesting benchmarks
2) Remove duplicate regressions/improvements for things that we don't care about specifically

This feature will be used for size benchmarks (https://github.com/rust-lang/rustc-perf/pull/1413).

We don't really need to add the same field for scenarios, because ignoring scenarios is simply done by not adding any patch files.

### DIFF
--- a/collector/src/benchmark/mod.rs
+++ b/collector/src/benchmark/mod.rs
@@ -6,7 +6,7 @@ use crate::execute::{CargoProcess, Processor};
 use crate::toolchain::Compiler;
 use anyhow::{bail, Context};
 use log::debug;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::mem::ManuallyDrop;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


collector/src/benchmark/mod.rs
```diff
@@ -183,11 +187,22 @@ impl Benchmark {
     ) -> anyhow::Result<()> {
         let iterations = iterations.unwrap_or(self.config.runs);
 
-        if self.config.disabled || profiles.is_empty() {
+        let profiles: Vec<Profile> = profiles
+            .into_iter()
+            .copied()
```
rylev, line 192: nit: is the copied necessary?

rylev (approved):
👍

collector/src/benchmark/mod.rs
Kobzol: Well, it is for the return value to `Vec<Profile>`. It's not needed if it was `Vec<&Profile>`.

rylev: Ah sorry - I thought the first profiles was a `Vec<Profile>` so `into_iter` yielded `Profile` not `&Profile`. Sorry for the noise!

nnethercote (approved):

PR #1445, opened 2022-09-06
Add `ripgrep-13.0.0-tiny` benchmark

Kobzol:
This should serve as an example of a real-world binary that is optimized for binary size.

Here we can see a disadvantage of the current approach for "special" benchmarks, where we just add a new benchmark and filter unwanted profiles/scenarios - we need to duplicate the code of the benchmark that we want to compile in a special way.


Mark-Simulacrum:
I guess it might be worth asking - does ripgrep optimize for small binary size? Or are we just seeing what the compiler can do on a "random" crate?

It seems like user choices here probably matter a great deal, though I have no hard evidence either way.

Kobzol:
It doesn't, AFAIK. It's more of a test how well does `rustc` optimize a generic, reasonably sized binary, when you pass it all the available flags. I consider this to be interesting.

FWIW, these are the binary sizes:

| Command | Size |
| --- | --- |
| `ripgrep` | 29 MiB |
| `ripgrep` + `strip` | 4.1 MiB |
| `ripgrep-tiny` | 2.6 MiB |
```

nnethercote (approved):
This seems fine. Are you planning to add any more `-tiny` benchmarks? I feel like this and `helloworld` are enough for now. A small one and a larger one seems like enough to evaluate things.

On the code duplication: symbolic links? I guess Windows may be a problem? I found https://stackoverflow.com/a/59761201 it sounds like symlinks with git on Windows is possible, but requires jumping through some hoops, which might not be worthwhile.

Kobzol:
I didn't plan to add more at this moment, as you said, one trivial program and one real-world program should be enough for now I think. I'd rather not mess with symlinks in git, I think that it would bring more problems than anything else. I don't think that adding these files is a big problem, it's just something to think about if we were to create a lot of new benchmarks with similar special compilation options.

rylev (approved):

PR #1460, opened 2022-10-07
Add a BufReader benchmark for reads smaller than the buffer

saethlin:
I think this is a reasonably-real-world example of running into the regression in https://github.com/rust-lang/rust/issues/102727

The idea is that a user may drive a `BufReader` by manually accessing the buffer, then calling `consume` then `fill_buf`. The regression is hit when the wrapped `Read` impl doesn't fill the whole `BufReader` buffer in one read. Which a snappy decoder won't, because a decompressed data frame is limited to 65536 bytes.

### DIFF
--- a/collector/runtime-benchmarks/bufreader/src/main.rs
+++ b/collector/runtime-benchmarks/bufreader/src/main.rs
@@ -1,0 +1,34 @@
+use benchlib::benchmark::{black_box, BenchmarkSuite};
+use snap::{read::FrameDecoder, write::FrameEncoder};
+use std::io::{BufRead, BufReader, Write};
+
+const BYTES: usize = 64 * 1024 * 1024;
+
+fn main() {
+    let mut suite = BenchmarkSuite::new();
+
+    // Inspired by https://github.com/rust-lang/rust/issues/102727
+    // The pattern we want is a BufReader which wraps a Read impl where one Read::read call will
+    // never fill the whole BufReader buffer.
+    suite.register("bufreader-snappy", || {
+        let data = vec![0u8; BYTES];
+        move || {
+            let mut compressed = Vec::new();
+            FrameEncoder::new(&mut compressed)
+                .write_all(&data[..])
+                .unwrap();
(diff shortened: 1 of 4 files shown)


collector/runtime-benchmarks/bufreader/src/main.rs
```diff
@@ -0,0 +1,30 @@
+use benchlib::benchmark::BenchmarkSuite;
+use snap::{read::FrameDecoder, write::FrameEncoder};
+use std::io::{BufRead, BufReader, Write};
+
+const BYTES: usize = 64 * 1024 * 1024;
+
+fn main() {
+    let mut suite = BenchmarkSuite::new();
+
+    suite.register("bufreader-snappy", || {
```
Kobzol, line 10: Could you please include a short comment above this benchmark that describes what does it measure, possibly with a link to the regression issue?

collector/runtime-benchmarks/bufreader/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+                .write_all(&data[..])
+                .unwrap();
+            let mut reader = BufReader::with_capacity(BYTES, FrameDecoder::new(&compressed[..]));
+
+            while let Ok(buf) = reader.fill_buf() {
+                if buf.is_empty() {
+                    break;
+                }
+                let len = buf.len();
+                reader.consume(len);
```
Kobzol, line 24: I don't suppose that it wouldn't be optimized out, but maybe we could return something computed from the closure? Anything returned will be turned into a `black_box`. But I'm not sure if it's that important here.

Kobzol (commented):
Thank you! I left two comments.

collector/runtime-benchmarks/bufreader/src/main.rs
saethlin: I added a `black_box` call in the loop, because in real code, we'd probably do some complicated operation with the buffer inside the loop. Does that sound good to you?

Kobzol: Yes, thank you!

Kobzol (approved):
Thanks!

PR #1466, opened 2022-10-12
Fix CI package installation

Kobzol:
Some package was returning 404, let's see if updating `apt` cache helps with it.


Kobzol:
Looks like it has helped indeed.

nnethercote (approved):

PR #1471, opened 2022-10-18
Add support for comma-separated include/exclude filters to runtime benchmarks

Kobzol:
Before the CLI help claimed commas are supported, but they weren't.


Kobzol:
I have it in my TODO list, it's a bit more complicated because the benchmark names are distributed amongst several binaries. Maybe I'll move the filtering to collector from benchlib.

PR #1475, opened 2022-10-19
Test runtime benchmarks on CI

Kobzol:
Instead of just compiling them, actually run one iteration of each runtime benchmark on CI.


nnethercote (approved):
👍

Issue #1488, opened 2022-11-03
Hide previous perf bot comments


rylev:
Similar to https://github.com/rust-lang/triagebot/issues/429, it would be nice if rustc-perf comments were hidden when they become irrelevant.

Mark-Simulacrum:
https://github.com/rust-lang/rust-log-analyzer/pull/56 has a sample implementation for another of our bots.

PR #1490, opened 2022-11-03
Hide old `rust-timer` comments when a summary is posted

Kobzol:
When a perf. summary is posted to a PR, all other non-summary comments made by the bot on the PR will be minimized. I copied the code from https://github.com/rust-lang/rust-log-analyzer/pull/56 to implement this.

I have no idea how to test this though :man_shrugging:

Fixes: https://github.com/rust-lang/rustc-perf/issues/1488

### DIFF
--- a/site/src/github/comparison_summary.rs
+++ b/site/src/github/comparison_summary.rs
@@ -6,6 +6,7 @@ use crate::load::SiteCtxt;
 
 use database::{ArtifactId, QueuedCommit};
 
+use crate::github::RUST_REPO_GITHUB_API_URL;
 use std::collections::HashSet;
 use std::fmt::Write;
 
@@ -66,10 +67,7 @@ pub async fn post_finished(ctxt: &SiteCtxt) {
 ///
 /// `is_master_commit` is used to differentiate messages for try runs and post-merge runs.
 async fn post_comparison_comment(ctxt: &SiteCtxt, commit: QueuedCommit, is_master_commit: bool) {
-    let client = super::client::Client::from_ctxt(
-        ctxt,
-        "https://api.github.com/repos/rust-lang/rust".to_owned(),
-    );
+    let client = super::client::Client::from_ctxt(ctxt, RUST_REPO_GITHUB_API_URL.to_owned());
     let pr = commit.pr;
     let body = match summarize_run(ctxt, commit, is_master_commit).await {
         Ok(message) => message,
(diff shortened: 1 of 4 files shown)


site/src/github/comparison_summary.rs
```diff
@@ -57,26 +58,44 @@ pub async fn post_finished(ctxt: &SiteCtxt) {
...
         Err(error) => error,
     };
 
+    body.push_str(COMMENT_MARK_SUMMARY);
     client.post_comment(pr, body).await;
+
+    let graph_client =
+        super::client::Client::from_ctxt(ctxt, RUST_REPO_GITHUB_GRAPH_URL.to_owned());
+    for comment in graph_client.get_comments(pr).await? {
+        if comment.viewer_did_author
```
rylev, line 90: Can you explain what this `viewer_did_author` check is for? I'm not sure I fully understand it.

site/src/github/comparison_summary.rs
```diff
@@ -57,26 +58,44 @@ pub async fn post_finished(ctxt: &SiteCtxt) {
...
+) -> anyhow::Result<()> {
+    let client = super::client::Client::from_ctxt(ctxt, RUST_REPO_GITHUB_API_URL.to_owned());
     let pr = commit.pr;
-    let body = match summarize_run(ctxt, commit, is_master_commit).await {
+    let mut body = match summarize_run(ctxt, commit, is_master_commit).await {
         Ok(message) => message,
         Err(error) => error,
     };
 
+    body.push_str(COMMENT_MARK_SUMMARY);
```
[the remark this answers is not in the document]
Kobzol: Ah right. Well, there's a second approach that I can use - explicitly mark all comments that should be hidden (so awaiting, queued and insufficient privileges) and only hide these when a summary is posted. This would keep the "unrolled PRs..." comments intact.

site/src/github/comparison_summary.rs
Kobzol: My understanding is that this is true if the account/token that queries the data (so the bot) is also the original author of the comment.

rylev: This might be worth a comment somewhere

Kobzol:
Added a comment about `viewer_did_author` and changed the marking logic from allowlist to denylist.

rylev (approved):
Looks good! We can still consider in another PR hiding old perf result summaries if we can figure out how to get that logic right.

Kobzol:
I think that old summaries can actually be quite useful, but we can open a discussion about that later I guess.

Kobzol:
I suppose that there's not a practical way of testing this without merging this? :sweat_smile: @Mark-Simulacrum

Mark-Simulacrum:
I would just merge.

PR #1491, opened 2022-11-03
re-add "cycles" metric used on windows

lqd:
When trying to look at cycles on windows, the following error about a missing metric is triggered now. 

image

This PR updates the site to handle this kind of metric again.


Kobzol (approved):

lqd:
I believe CI is failing because the cargo benchmark is currently broken. So I'll merge this.

PR #1494, opened 2022-11-05
Split GraphQL client from the REST API GitHub client and use proper authentication for it

Kobzol:
This should hopefully fix comment hiding. Tested on my repo with a token with the `repo` scope.

User agent header was required, along with authentication header. I also split the clients into two parts, since their functionality doesn't have overlap and it would only lead to wrong usages if both implementations were in the same struct.

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -9,7 +9,6 @@ use serde::Deserialize;
 type BoxedError = Box<dyn std::error::Error + Send + Sync>;
 
 pub const RUST_REPO_GITHUB_API_URL: &str = "https://api.github.com/repos/rust-lang/rust";
-pub const RUST_REPO_GITHUB_GRAPH_URL: &str = "https://api.github.com/graphql";
 
 /// Comments that are temporary and do not add any value once there has been a new development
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Can you say more about what goes wrong if they're in the same struct? I would expect which API we're using to not really matter to callers...

Kobzol:
The clients have different requirements.

The REST API client needs a URL for a given repository, while the GraphQL client always works with the same URL. So if a user passes a different URL (e.g. a REST API URL) to the client and uses a GraphQL method, it will not work. Or vice versa, if a user passes a GraphQL URL and uses a REST API method, it will also not work. The two clients also use a different authentication method, which was the source of the bug in the original PR.

Since the clients have a need for different URL handling, have different authentication methods and their functionality doesn't overlap, I think that it will be best if there will be two separate structs, which makes it more difficult to use them in a wrong way. For example, now the `GraphQLClient` has only a single constructor, which disallows passing an invalid URL.

Mark-Simulacrum (approved):

PR #1500, opened 2022-11-22
Update bytehound sha after they force pushed

rylev:
CI was breaking because the bytehound repo force pushed the main branch and the sha we were pinned to no longer exists.


Kobzol (approved):
Well, LGTM if CI passes :)

PR #1510, opened 2023-01-07
chore: fix typo, Propegete -> Propagate

chansuke:


Kobzol:
Hi, typo fixes are welcome, but this code is inside one of the compile benchmark crates (the `compile-benchmarks` directory), which is not really part of the `rustc-perf`, it's just copied source code from an external crate. So I don't think that it's worth it to modify it in any way.

chansuke:
@Kobzol I appreciate your review and feedback. Ok, I will close this issue.

PR #1515, opened 2023-01-23
Store runtime benchmark results into SQLite

Kobzol:
Currently, each iteration of a runtime benchmark is stored with a fresh collection ID. This is something that we may want to change, although there is currently a unique index across (`pstat_series`, `artifact_id` and `collection_id`), so we cannot easily associate multiple metric sets with a single collection ID.

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -118,7 +118,7 @@ fn bench(
             }
             let mut tx = rt.block_on(conn.transaction());
             let (supports_stable, category) = category.db_representation();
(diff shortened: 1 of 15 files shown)


Kobzol:
ping @Mark-Simulacrum

Mark-Simulacrum (approved):
Didn't review in too much detail but this looks fine to me at a high level. I imagine we can iterate on the specifics (e.g., whether we need a separate table for these benchmarks) over time - until we add Postgres support and start collecting in production we can throw the tables away easily without worrying about migrations etc.

PR #1518, opened 2023-01-30
Add comment about unifying types to Profile

Kobzol:
Suggested by @nnethercote [here](https://github.com/rust-lang/rustc-perf/pull/1511#issuecomment-1407791939).

I didn't add the same comment to `Scenario`, as it would be mostly duplicated (and scenario is not so similar in the DB, so people probably won't be inclined to try to merge it).

I added the note as a regular comment, as it's just an implementation note and shouldn't appear in rustdoc (not that many people look at docs of `rustc-perf` I suppose :) ).


nnethercote (approved):
Thank you!

PR #1520, opened 2023-01-30
Revert "Do not hardcode stable compile-time benchmarks in dashboard"

Kobzol:
This reverts commit 3ab8ee7539a7e4ba38cc381fc9dc3c30af562d94.

https://github.com/rust-lang/rustc-perf/pull/1519 sadly broke dashboard, because apparently the benchmark directory is not contained in the `rustc-perf` Dockerfile (can I add it there?).

This is a hotfix to revert the code that caused the problem.


nnethercote (approved):

Mark-Simulacrum:
Expanding the docker file should be fine. The main reason it's not there today is just size, I think.

PR #1525, opened 2023-02-19
Display "No relevant results" on compare page where there are filtered non-relevant results

Kobzol:
This PR modifies the text on the  compare page to "No relevant results" when there are non-relevant results filtered out. In theory, we could also generalize to something like "No results (some results are filtered)", but that message would apply everytime, because there are always at least *some* results for each filter configuration. And since "Show non relevant" is the only filter checkbox that is turned off by default, I think that it makes sense to specialize for it.

Suggested [here](https://github.com/rust-lang/rust/pull/108159#issuecomment-1435389638).

### DIFF
--- a/site/static/compare.html
+++ b/site/static/compare.html
@@ -314,7 +314,8 @@
             </div>
             <test-cases-table
                 title="Primary"
-                :cases="testCases.filter(c => c.category === 'primary')"
+                :cases="primaryCases"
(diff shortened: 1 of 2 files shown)


nnethercote (approved):

rylev (approved):
The CI failure has nothing to do with this change.

PR #1533, opened 2023-03-02
Display doc benchmarks in dashboard

Kobzol:
It was missing from the dashboard.


nnethercote (approved):
Will we have historical data for doc builds?

Kobzol:
I certainly hope so, I think that we gather all data when measuring released artifacts.

PR #1539, opened 2023-03-09
Add template rendering engine to site

Kobzol:
Use Tera to render HTML templates. This will help us remove duplication (for the top link bar, general page layout, and shared Vue components).

Rendering a template takes about ~10 microseconds in release mode on my laptop. It's pretty much similar time to just reading the raw HTML file directly from disk (what was done before). Actually it could be a bit faster, as Tera now caches the template content in memory. But it shouldn't matter, as the HTML files are aggressively cached anyway using HTTP cache.

Only the help page has been ported to templates so far.

I didn't use e.g. `askama`, because that inlines the templates into the binary, which would mean that livereloading would require a rebuild of the `site` binary (which takes seconds, so it's not exactly instant for interactive development).

### DIFF
--- a/site/src/server.rs
+++ b/site/src/server.rs
@@ -2,7 +2,7 @@ use brotli::enc::BrotliEncoderParams;
 use brotli::BrotliCompress;
 use std::collections::HashMap;
 use std::net::SocketAddr;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering as AtomicOrdering};
 use std::sync::Arc;
@@ -28,6 +28,7 @@ pub use crate::api::{
 use crate::db::{self, ArtifactId};
 use crate::load::{Config, SiteCtxt};
 use crate::request_handlers;
+use crate::templates::TemplateRenderer;
 
 pub type Request = http::Request<hyper::Body>;
 pub type Response = http::Response<hyper::Body>;
@@ -336,7 +337,7 @@ async fn serve_req(server: Server, req: Request) -> Result<Response, ServerError
         None
     };
 
-    if let Some(response) = handle_fs_path(&req, path) {
+    if let Some(response) = handle_fs_path(&req, path).await {
         return Ok(response);
     }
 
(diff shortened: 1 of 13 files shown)


site/src/templates.rs
```diff
@@ -0,0 +1,37 @@
+use parking_lot::RwLock;
+use std::path::{Path, PathBuf};
+use tera::Tera;
+
+pub struct TemplateRenderer {
+    template_dir: PathBuf,
+    tera: RwLock<Tera>,
```
rylev, line 7: Just want to check that we're fine using a sync lock here in async contexts. I think this should be fine since live reloading should only be used locally where the lock shouldn't be under contention.

rylev (approved):

rylev:
Also, shouldn't we be removing the help.html static file?

site/src/templates.rs
Kobzol: That's a good point. But indeed as you said, on production environments it should never even block, as there would be only readers. I can make it async though if you'd like, it's a small change.

site/src/server.rs
```diff
@@ -608,6 +612,15 @@ fn handle_fs_path(req: &Request, path: &str) -> Option<http::Response<hyper::Bod
         Some("js") => response = response.header("Content-Type", "application/javascript"),
         _ => (),
     }
+
+    let source = match path {
+        "/help.html" => TEMPLATES
+            .render(path.trim_start_matches("/"))
```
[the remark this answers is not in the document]
Kobzol: Well, no :D But it's a preparation for another PR where I will port everything to templates.

site/src/templates.rs
rylev: If it's a small change, it might be worth to do, but I don't think it really matters.

Kobzol: I checked the rest of the code and this function anyway did blocking file I/O... so maybe it would require a larger refactor so that it would make sense. I can change it to async, but then there would still be sync code just next to it.

Kobzol:
> Also, shouldn't we be removing the help.html static file?

Yes, but the code currently checks that the HTML file exists, and it would need a larger refactoring to change that. And since I want to convert everything to templates anyway in a follow-up PR, I don't think that's worth it.

Kobzol:
Pff, `tera` depends on a specific version of `chrono` that triggers deprecation notices in the current code. I'll fix it in another commit.

PR #1542, opened 2023-03-09
Convert status and bootstrap pages to templates

Kobzol:


rylev (approved):
I didn't run this locally, but assuming things still look correct, this is fine by me.

Kobzol:
Checked with live data that it looks the same. There's now additional padding applied to all pages thanks to the use of `<body class="container">`, but I don't see it as an issue (before it was only applied to some pages). We can change it if space becomes an issue.

PR #1544, opened 2023-03-11
Fix future-incompat warnings

jyn514:
By upgrading `fs-extra` and `quick-xml`

Fixes the following warning:
```
warning: the following packages contain code that will be rejected by a future version of Rust: fs_extra v1.2.0, quick-xml v0.20.0
```


Kobzol (approved):
Thanks!

PR #1547, opened 2023-03-11
Document perf-config.json in the readme

jyn514:
"document" is a little strong for "point to the reference struct", but, close enough


Kobzol (approved):
:laughing: Fair enough :)

PR #1549, opened 2023-03-13
Optimize Dockerfile to speed up CI caching

Kobzol:
Use `cargo chef` + sparse `crates.io` registry in Dockerfile to speed up Docker (re)builds. Also add GHA cache for Docker layers and use Buildkit for building on CI.

This reduces the time to run the required Linux CI build from 10-12 minutes to 3-4 minutes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,13 @@
 FROM node:18 as frontend
 
-COPY ./site ./site
+COPY ./site/frontend ./site/frontend
 RUN cd site/frontend && npm ci
 RUN cd site/frontend && npm run check
 RUN cd site/frontend && npm run build
 
-FROM ubuntu:20.04 as build
+FROM ubuntu:20.04 as base
+
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
 RUN apt-get update -y && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


Kobzol:
First run, 10 minutes, trying the next one.

Kobzol:
Hmm, not much of a help :( We'd either need to build outside of Docker or have something like https://github.com/rust-lang/cargo/issues/2644 for this to work.

Kobzol:
I haven't. Do you have any examples? :)

mati865:
This is not that far off from what we use in a company I work for: https://github.com/benmarten/sccache-docker-test/blob/master/Dockerfile
Just some notes: this syntax is no longer experimental and it's worth to cache Cargo registry the same way as `SCCACHE_DIR`. I can try to find some time to cook PR over the weekend.

Kobzol:
This looks useful for e.g. local builds and for speeding-up CI, but ideally we would like to do it without complicating the `Dockerfile`, which is used for live deployment, and I don't want to stuff `sccache` wrappers and other things into it.

The current CI time is not terrible, I just wanted to see if there is an easy way of speeding it up when the build happens in Docker.

Kobzol:
Uhh, finally. Now rebuild takes ~4 minutes instead of 10+ minutes.

Dockerfile
```diff
@@ -1,11 +1,13 @@
 FROM node:18 as frontend
 
-COPY ./site ./site
+COPY ./site/frontend ./site/frontend
 RUN cd site/frontend && npm ci
 RUN cd site/frontend && npm run check
 RUN cd site/frontend && npm run build
 
-FROM ubuntu:20.04 as build
+FROM ubuntu:20.04 as base
```
Mark-Simulacrum, line 8: We should probably bump to 22.04 LTS at some point as well...

Mark-Simulacrum (approved):

PR #1553, opened 2023-03-15
Log preparation errors

Kobzol:
This PR adds logging for errors during preparation. If preparation of a benchmark fails in any way or panics (shouldn't happen, hopefully), the error will now be correctly propagated and logged.

This PR also removes dependency on `crossbeam-utils`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,7 +395,6 @@ dependencies = [
  "cargo_metadata",
  "chrono",
  "clap",
- "crossbeam-utils",
  "database",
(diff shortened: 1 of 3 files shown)


Kobzol:
CC @Mark-Simulacrum, now it doesn't ignore errors and also shouldn't panic on them :) Tested locally that both a compilation error and a panic within a preparation thread now logs the error into the database properly.

Mark-Simulacrum (approved):

PR #1554, opened 2023-03-15
Embed static resources and templates into the `site` binary in release mode

Kobzol:
… and live-reload them in debug mode.

This makes the `site` binary self-contained, which enables it to be more easily distributed (e.g. from GH releases).

Binary size of the `site` binary:
| Version          | Size (B) | Size after `strip` (B) |
|---------------|----------------|--------------------------|
| before this PR            | 59733152        | 19013912                |
| after this PR          | 59356184        | 18866032                |

Funnily enough, the binary is actually smaller after this PR (probably because of some dependency version or code changes).

Incremental rebuild time after a tiny change in `debug` mode:
```
before: 8.5s
after: 9.2s
```

I checked that the Docker image can be built and properly serves content even without the `static` and `templates` directories.


Kobzol:
Ping @Mark-Simulacrum.

Mark-Simulacrum (approved):

PR #1555, opened 2023-03-15
Prepare CI workflow for nightly releases

Kobzol:
This new CI workflow builds a `rustc-perf` site binary each day (night) and publishes it into a draft GH release. The release is overwritten each day, so we won't be spammed with hundreds of releases.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.66.0
+          toolchain: stable
           override: true
 
       - uses: Swatinem/rust-cache@v1
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,55 @@
+name: Nightly release
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 23 * * *"
+
+jobs:
+  publish_release:
+    name: Publish release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set version
+        run: |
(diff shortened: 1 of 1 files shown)


Kobzol:
Here is the built release: https://github.com/rust-lang/rustc-perf/releases/tag/nightly.

Kobzol:
Rebased over `master`. Once this is merged, we should be free to start using `npm;.

Kobzol:
I'd like to merge this before https://github.com/rust-lang/rustc-perf/pull/1565 so that we can provide a nightly release for users that do not want to download `npm` to build the website.

Ping @Mark-Simulacrum

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout the source code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: 1.66.0
```
Mark-Simulacrum, line 25: Is pinning this the right decision? Should we use stable here?

Mark-Simulacrum (approved):
This seems okay for now. Let's see how it works out in practice, we may want to put the site binary into our perf S3 bucket as an alternative to GHA if there's any problems with the overwriting here.

.github/workflows/nightly.yml
Kobzol: Indeed `stable` will be better, to regularly check that we can still build on it.

Kobzol:
I use a similar approach in a different repo and it works fine I think. The only issue is that the overwritten nightly release still keeps the "publication date" of the first release (so it looks like the release was published a long time ago), but that's just a cosmetic GitHub UI thing, the description of the release clearly states the date of the nightly.

PR #1556, opened 2023-03-15
Fix `EXPECT_ONLY_WRAPPED_RUSTC` for `cargo llvm-lines`

Kobzol:
And re-enable `llvm-lines` `regex` CI check.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1537


Mark-Simulacrum (approved):
I'm okay with this even if it's unfortunate we need to do it - in general it feels like a hack.

nnethercote (approved):

Kobzol:
It is a hack, but I'm not really sure how to resolve it and I don't want to go investigate how cargo llvm-lines works when we can solve it with this. Since it's just a local profiling tool and not something we use on CI for benchmarking, I think that this is fine for now.

PR #1558, opened 2023-03-18
Bump dependencies to latest versions

Mark-Simulacrum:
We had a bunch of outdated dependencies, this brings us up to date with latest versions, making necessary adjustments were there were known breaking changes.


Mark-Simulacrum:
Bytehound update seems to have changed the failure but not fixed it, I think updating probably still makes sense and we should move ahead despite that breakage (it's preexisting on master anyway).

nnethercote (approved):
I suggest commenting out the bytehound testing, to keep the job green. I never use Bytehound myself so I'm not much worried if it's broken.

PR #1565, opened 2023-03-31
Introduce frontend build system using npm, Parcel and TypeScript

Kobzol:
Since the `site` binary is now self-contained (and soon will be published on CI via [nightly releases](https://github.com/rust-lang/rustc-perf/pull/1555)), we can now add some basic frontend tooling to make it easier to change the perf.RLO UI in the future.

This PR adds a `npm` project (which is required to use common modern frontend technologies) and enables the use of TypeScript in UI components using the `Parcel` bundler. I wanted to use only TypeScript at first, but it cannot bundle `.vue` single file components, which we will hopefully use to author the UI components, out of the box. And we will also want to bundle at least (S)CSS, so I chose to use a build system. Parcel doesn't require any configuration, is fast, works pretty much out of the box and very easily allows to build the app into multiple components (which is very useful, because the UI site will be a multi-page application, not a single-page application). I tried several other bundlers, but didn't encounter anything else that would have all these properties.

Currently the transpilation step from TypeScript assumes ES2015, and e.g. keeps `async` functions and `await` in the output. I was a bit worried about this, but then I realized that browsers have supported these things for a long time now.. so it should be fine. If there would be some problem, we can always switch back to a polyfill.

(Btw the actual diff is like 150 lines of code, most of the added lines is just the `npm` lockfile :sweat_smile:).


site/frontend/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "lib": ["es5", "es2015", "dom"],
+    "target": "es5",
```
erikdesjardins, line 4: The PR description says ES2015, but this is targeting ES5. I would suggest targeting _ES2017_ or newer, so you avoid the regenerator cruft for async functions, which makes them annoying to debug.

This shouldn't lose support for any reasonable browser (at $dayjob we target ES2020). Likely any browser that does not support ES2017 today is actively insecure to expose to the internet.

Kobzol: Yeah so actually, this file only exists for type checking (npm run check). Parcel by itself doesn't do any typechecking and doesn't generate the regenerator polyfill, I checked. Maybe I should put this into README.md.

erikdesjardins: Ah, I should've noticed tsc was only used there.

Putting some basic docs about the build process in the readme would be useful, yeah

site/.gitignore
```diff
@@ -1,2 +1,4 @@
 /data
 /target
+frontend/node_modules
+frontend/dist/**/
```
the8472, line 4: you'll want to add those to dockerignore too, otherwise it'll send them to the build service when doing that locally.

Dockerfile
```diff
@@ -16,12 +16,19 @@ RUN apt-get update -y && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- \
     --default-toolchain stable --profile minimal -y
 
+RUN curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - && \
+  apt-get install -y nodejs
```
the8472, line 20: Maybe do the npm stuff in a separate stage using a node image as base?

Kobzol: That's a great idea! Thanks. I was wondering how to download `npm` more elegantly and didn't realize that a staged build is perfect for this.

rylev (approved):

PR #1570, opened 2023-04-18
Port status page script to TypeScript

Kobzol:
For these simple pages, I'm just moving them over to NPM to reduce some code duplication and remove old code. I'm not trying to refactor them heavily nor describe the server data response in super detail (it would be quite annoying for all the enum variants). The biggest benefit of TS should be in the pages that use Vue (the compare page and soon the runtime UI page(s)).

For the status page, the downloaded JS bundle size goes down from ~15k to ~3k (not that it's super important here).

### DIFF
--- a/site/frontend/package.json
+++ b/site/frontend/package.json
@@ -25,6 +25,10 @@
     "dashboard": {
       "source": "src/pages/dashboard.ts",
       "distDir": "dist/scripts"
+    },
+    "status": {
+      "source": "src/pages/status.ts",
+      "distDir": "dist/scripts"
     }
   },
   "browserslist": "> 0.5%, last 3 years, not dead"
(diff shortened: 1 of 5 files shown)


nnethercote (approved):
Seems ok to me. My only comment is that there are two uses of `any`. Can they be avoided? If not, brief comments explaining why (and what form the data takes) would be helpful.

Kobzol:
Yeah so this relates to this comment: 
> nor describe the server data response in super detail (it would be quite annoying for all the enum variants).

The Rust side uses enums which are encoded by serde in a way that is not super TypeScript friendly. I could model them (or change the serde encoding), but I'm not sure if it provides a lot of value, since it does not give us any guarantee that the incoming data will have this specific side anyway - everytime the server side changes, we need to update the TS side too. And since this data response is only used on one place in this file, I decided to only model the easy stuff and leave the rest as `any`.

Kobzol:
Added a comment explaining the usage of `any`.

PR #1573, opened 2023-04-22
Port bootstrap page to Typescript

Kobzol:
This one contains more changes than the status page. I started to make use of the new tooling to port the page to Vue. Now it dynamically refreshes itself when the selected data changes, instead of reloading the whole page. It still updates the URL though, so Go back in the browser works as expected and the URL can be shared to display the same data selection.

### DIFF
--- a/site/frontend/package-lock.json
+++ b/site/frontend/package-lock.json
@@ -12,10 +12,12 @@
         "highcharts": "6.0.7",
         "parcel": "^2.8.3",
         "sass": "^1.59.3",
+        "uplot": "^1.6.24",
         "vite": "^4.2.1",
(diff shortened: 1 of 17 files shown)


Kobzol:
After experimenting with it for a bit, I decided to throw away the "SPA aspect" for now. Turns out that with the `bfcache` optimization, it's actually much faster to go back/forward in browser history here if the page is reloaded on every data selection change. If we implemented it as an SPA, we would need to cache the downloaded data ourselves to make it as fast. That's not worth it here at this moment I think.

The code is also a bit simpler now.

rylev (approved):
Looks fine to me!

PR #1574, opened 2023-04-22
Only run `push` CI workflows on push to master

Kobzol:
I noticed that we run all (!) CI workflows twice on each PR commit, because we use an unfiltered `push` event. That's wasteful and increases latency of PR checks and also merging PRs. We should only run the `push` event on master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: CI
 on:
-  push: {}
+  push:
+    branches:
+      - master
   pull_request: {}
   schedule:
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
Typically the way to avoid this is to push branches to some other repo (e.g., your private fork). That's also nice to avoid polluting shared space for git pulls.

Kobzol:
Ah, right, that was the reason why it was running twice. Well, it still should run only on `master` anyway, but I'll try to create branches on my fork from now on.

PR #1575, opened 2023-04-22
Cache Rust artifacts of Windows CI workflows to make them faster

Kobzol:
The windows CI check is quite slow (10-12 minutes), but it doesn't use the Docker build. Therefore it should be quite straightforward to make it faster with the `Swatinem/rust-cache` action.


Kobzol:
Force pushing to test if the cache works (the first build took 6-8 minutes).

Mark-Simulacrum (approved):

PR #1576, opened 2023-04-23
replace rustc_fake with rustc_wrapper

csmoe:
1. the 1-feature.patch makes cargo compile-able on osx;
2. RUSTC=rustc_fake makes the build script which depends on `Command::new("RUSTC")` failed like [autocfg](https://github.com/cuviper/autocfg/blob/e75324cfa5fa0bec4f06f9609ab169bb7747f984/src/lib.rs#L156)


Kobzol:
Hi, could you maybe explain a little bit what is the motivation for this change (and what does it even do?)?

Kobzol:
I'm gonna close this, as it has lingered without a response for a few years. Feel free to reopen if you still want to land it.

PR #1581, opened 2023-04-29
Port compare page to Typescript

Kobzol:
This was the largest UI page to port. It was partially in Vue before, so I mostly just copied it over, refactored it slightly and cleaned up various stuff.


site/frontend/src/pages/compare/benchmarks/benchmarks.vue
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const props = defineProps<BenchmarkProps>();
+
+function Section({title, link, linkUp}: {title: string, link: string, linkUp: boolean}) {
+  return (
+    <div class="category-title">
+      {title} benchmarks
+      <span title={`To ${link} benchmarks`}>&nbsp;
+        <a href={`#${link}-benchmarks`}>{linkUp ? "⮭" : "⮯"}</a>
```
rylev, line 22: This doesn't seem to render properly in the diff?

Screenshot 2023-05-19 at 14 49 27

rylev (approved):
I didn't look at this super closely, but the overall shape looks good. I think this is an improvement over the status quo.

site/frontend/src/pages/compare/benchmarks/benchmarks.vue
Kobzol: It's fine in my source editor, probably just some visual bug.

Kobzol:
Ok, thanks! Let's ship it.

PR #1588, opened 2023-05-19
Update website build instructions

Kobzol:
Now that we have a nightly build.

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -7,13 +7,28 @@ which displays graphs to the user.
 This also contains a GitHub bot to trigger on-demand benchmarking.
 
 ## Setup
-
-The site launches on port 2346 by default, which can be overridden by setting
-the `PORT` environment variable.
-
-```
-git clone https://github.com/rust-lang/rustc-perf.git
-```
+You can build the website in multiple ways:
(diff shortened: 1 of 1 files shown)


site/README.md
```diff
@@ -7,13 +7,28 @@ which displays graphs to the user.
...
+  $ cd site/frontend
+  $ npm run build
+  
+  # Build website binary
+  $ cd ../..
+  $ cargo build --bin site --release
+  ```
+  The `--release` flag is on purpose, reducing startup time 15x from roughly 15
+  seconds to 1; this is on the production database and smaller data sets are
+  likely to load considerably faster.
```
lqd, line 31: Small datasets do indeed load super fast even without optimizations. If you're working on the website with a small database, it's probably even worth using debug builds for development, for the faster initial clean build (-40%) and subsequent faster incremental builds (-50%) as well. If only working on the js frontend of course the debug and release builds of the binary are now the same duration.

Kobzol: This is more of an "usage" section, rather than "development" section. That is in a README in the frontend directory (maybe it should be here)? For frontend development, debug is definitely the way to go, especially because it hot reloads the `dist` directory.

nnethercote (approved):

PR #1590, opened 2023-05-19
Port detailed query page to TypeScript

Kobzol:
This should be the last page to port. The only remaining Javascript file is `schedule.js`, but I don't want to touch that.


site/frontend/src/pages/detailed-query.ts
```diff
@@ -0,0 +1,319 @@
...
+  if (state.base_commit) {
+    let self_href =
+      `/detailed-query.html?sort_idx=${state.sort_idx}&commit=${state.commit}&scenario=${state.scenario}&benchmark=${state.benchmark}`;
+    let base_href =
+      `/detailed-query.html?sort_idx=${state.sort_idx}&commit=${state.base_commit}&scenario=${state.scenario}&benchmark=${state.benchmark}`;
+    txt += `<br>diff vs base ${state.base_commit.substring(0, 10)}, <a href="${base_href}">query info for just base commit</a>`;
+    txt += `<br><a href="${self_href}">query info for just this commit</a>`;
+  }
+  document.querySelector("#title").innerHTML = txt;
+  let dl_url = (commit, bench, run) => {
```
lqd, line 53: We could inline this, it has a single call-site.

site/frontend/src/pages/detailed-query.ts
```diff
@@ -0,0 +1,319 @@
...
+    return `/perf/download-raw-self-profile?commit=${commit}&benchmark=${bench}&scenario=${run}`;
+  };
+  let dl_link = (commit, bench, run) => {
```
lqd, line 56: The only thing that really changes in most of this code is the commit to use. So this and all the other closures are always used with constant "bench" and "run" from the state, we could remove them from the closure args and refer to the state's fields for example. There are other simplifications we could do inlining some of the closures and having the different "download/view" links be the reusable abstraction instead.

site/frontend/src/pages/detailed-query.ts
```diff
@@ -0,0 +1,319 @@
...
+    invocation_count,
+    incremental_load_time
+  };
+}
+
+interface Selector {
+  commit: string;
+  base_commit: string | null;
+  benchmark: string;
+  // #[serde(alias = "run_name")]
```
lqd, line 300:
```suggestion
```
We can remove that.

site/frontend/src/pages/detailed-query.ts
```diff
@@ -0,0 +1,319 @@
...
+
+async function loadData() {
+  const params = getUrlParams();
+  const selector: Selector = {
+    commit: params["commit"],
+    base_commit: params["base_commit"] ?? null,
+    benchmark: params["benchmark"],
+    scenario: params["scenario"],
+    sort_idx: params["sort_idx"] ?? "-2"
+  };
```
lqd, line 313: We probably could destructure `params`'s fields here.

site/frontend/src/utils/navigation.ts
```diff
@@ -2,12 +2,12 @@ export function createUrlParams(params: Dict<string>): URLSearchParams {
     return createUrlWithParams(params).searchParams;
 }
 
-export function createUrlWithParams(params: Dict<string>): URL {
+export function createUrlWithParams(params: Dict<any>): URL {
     const originalUrl = window.location.toString();
     const url = new URL(originalUrl);
     for (const [key, value] of Object.entries(params)) {
         if (value !== null && value !== undefined && value !== "") {
```
lqd, line 9: if values aren't strings anymore then should this check whether `toString()` returns an empty string instead ?

site/frontend/src/pages/detailed-query.ts
```diff
@@ -0,0 +1,319 @@
...
+  let bench_name = b => b.replace(/-[^-]*$/, "");
+  let scenario_filter = s => s == "full" ? "Full" :
+    s == "incr-full" ? "IncrFull" :
+      s == "incr-unchanged" ? "IncrUnchanged" :
+        s.startsWith("incr-patched") ? "IncrPatched" :
+          "???";
+  if (state.base_commit) {
+    txt += "<br>";
+    txt += `Diff: <a
+                        href="/perf/processed-self-profile?commit=${state.commit}&base_commit=${state.base_commit}&benchmark=${state.benchmark}&scenario=${state.scenario}&type=codegen-schedule"
```
lqd, line 115: Here we also could make use of the `processed_url` closure.

lqd (approved):
Nice! Glad to see the migration complete 🚀 .

I've left a couple notes, most of which could be useful if we wanted to clean this page up in the future. Maybe some day.

site/frontend/src/utils/navigation.ts
Kobzol: Good point!

Kobzol:
I performed most of the suggested changes (I closed the comments otherwise this couldn't be merged).

PR #1593, opened 2023-05-25
Fix compare page regression

Kobzol:
The URL wasn't being updated when filters have been changed. This is a regression from the TS port.

### DIFF
--- a/site/frontend/src/utils/copy.ts
+++ b/site/frontend/src/utils/copy.ts
@@ -1,0 +1,3 @@
+export function deepCopy<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value));
+}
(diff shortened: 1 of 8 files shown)


site/frontend/src/utils/copy.ts
```diff
@@ -0,0 +1,3 @@
+export function deepCopy<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value));
```
lqd, line 2: We would have to be careful about this if the filter ever stores Dates.

lqd (approved):
Works fine locally for me 👍

site/frontend/src/utils/copy.ts
Kobzol: Good point. I didn't want to include another dependency just for this one function. I suppose that this filter shouldn't care about dates, there's another filter for that.

PR #1594, opened 2023-05-25
Add code formatter for frontend

Kobzol:
I wanted to use 2 spaces for Vue templates (HTML) and 4 for everything else, but prettier [doesn't support that](https://github.com/prettier/prettier/issues/14755), so I just kept two spaces for everything. Formatting is now also checked in CI.


nnethercote (approved):

lqd:
Is CI failing because the prettier config is maybe not available within the docker image ?

PR #1598, opened 2023-05-30
Remove the now-unused ctfe step limit from the ctfe stress test

oli-obk:
This does not affect perf, it only affects whether the test builds at all. https://github.com/rust-lang/rust/pull/103877 wants to remove those limites entirely, so we first need to remove the (unstable) attribute here, in order to be able to perf test (and subsequently land) that PR

### DIFF
--- a/collector/compile-benchmarks/ctfe-stress-5/src/lib.rs
+++ b/collector/compile-benchmarks/ctfe-stress-5/src/lib.rs
@@ -1,6 +1,4 @@
 #![allow(dead_code)]
-#![feature(const_eval_limit)]
-#![const_eval_limit = "10000000"]
 use std::mem::MaybeUninit;
 
 // Try to make CTFE actually do a lot of computation, without producing a big result.


oli-obk:
I checked, and the ctfe step limit is not necessary for that perf test to run. We reduced the size of the test a while back with a bump to the trailing number of the perf test, making the attribute unnecessary. I checked, it still compiles.

Mark-Simulacrum (approved):
Sounds good - wasn't entirely clear based on your comment.

Mark-Simulacrum:
You may also need to bump the rustc-perf repo checkouts in rust-lang/rust (they're pinned by commit hash, currently: 9dfaa35193154b690922347ee1141a06ec87a199).

```
$ rg 9dfaa35193154b690922347ee1141a06ec87a199
src/ci/stage-build.py
178:        perf_commit = "9dfaa35193154b690922347ee1141a06ec87a199"

src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
58:ENV PERF_COMMIT 9dfaa35193154b690922347ee1141a06ec87a199
```

PR #1604, opened 2023-06-03
Store more runtime benchmark information into DB

Kobzol:
All relevant metrics are now stored in to the DB, and the collector progress (the list shown on the [status](https://perf.rust-lang.org/status.html) page) now also includes runtime benchmarks.


nnethercote (approved):
I'm weak on both async Rust and databases, but I don't see any problems here :)

PR #1610, opened 2023-06-05
Generalize site `pstat` querying for runtime benchmarks

Kobzol:
This PR generalizes the `site` benchmark data querying code to also make runtime benchmark data available. The generalization is now based on test cases - for compile benchmarks we have each benchmark parametrized by `(benchmark, profile, scenario)`, for runtime benchmarks it's currently only `(benchmark, )`.

To make it work on CI, this PR also implements support for runtime benchmarks for the Postgres DB backend.

### DIFF
--- a/site/src/selector.rs
+++ b/site/src/selector.rs
@@ -274,7 +274,7 @@ impl StatisticSeries {
         let index = ctxt.index.load();
         let mut statistic_descriptions = index
             .all_statistic_descriptions()
-            .filter(|&&(b, p, s, m)| {
+            .filter(|(&(b, p, s, m), _)| {
                 query.benchmark.matches(b)
                     && query.profile.matches(p)
                     && query.scenario.matches(s)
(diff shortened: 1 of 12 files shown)


site/src/selector.rs
```diff
@@ -299,13 +325,13 @@ impl StatisticSeries {
...
-            .zip(&statistic_descriptions)
+            .zip(statistic_descriptions)
             .filter(|(points, _)| points.iter().any(|value| value.is_some()))
-            .map(|(points, (&(benchmark, profile, scenario, metric), _))| {
+            .map(|(points, (test_case, metric, _))| {
                 SeriesResponse {
                     series: StatisticSeries {
                         artifact_ids: ArtifactIdIter::new(artifact_ids.clone()),
-                        points: if metric == *"cpu-clock" || metric == *"task-clock" {
+                        points: if *metric == *"cpu-clock" || *metric == *"task-clock" {
```
Kobzol, line 299: Could we just normalize these values when they are gathered (and then run a query on production DB that will normalize the existing values in the DB), and then get rid of this annoying special case?

Mark-Simulacrum: I'd be fine with that - I think the DB query might be a little annoying depending on how many of these there are but probably okay.

Mark-Simulacrum (approved):
This looks good to me, but we might want to merge ~Monday to give more bandwidth for debugging if anything breaks in production (busy most of the weekend).

PR #1615, opened 2023-06-15
Store filter and aggregations state in LocalStorage

Kobzol:
If you often examine the aggregations or switch filters on the compage page it can be annoying to open them each time the page is refreshed. This PR persists the "opened state" of these two sections in LocalStorage, to remember the choice.

### DIFF
--- a/site/frontend/src/pages/compare/header/filters.vue
+++ b/site/frontend/src/pages/compare/header/filters.vue
@@ -29,11 +29,17 @@ watch(
   },
   {deep: true}
 );
+
+function updateOpened(opened: boolean) {
+  filterOpened.value = opened;
+}
+
+const filterOpened = ref(false);
 </script>
 
 <template>
   <fieldset class="collapsible-section">
-    <Toggle :defaultOpened="false">
+    <Toggle :opened="filterOpened" @change="updateOpened">
       <template #label>Filters</template>
       <template #content>
         <div>
@@ -46,9 +52,8 @@ watch(
               <div style="width: 160px">
                 <span>Profiles</span>
                 <Tooltip
-                  >The different compilation profiles (check, debug, opt,
-                  doc).</Tooltip
-                >
+                  >The different compilation profiles (check, debug, opt, doc).
+                </Tooltip>
               </div>
             </div>
             <ul class="states-list">
@@ -84,9 +89,8 @@ watch(
                   <span class="cache-label">opt</span>
                 </label>
                 <Tooltip
-                  >Release build that produces as optimized code as
-                  possible.</Tooltip
-                >
+                  >Release build that produces as optimized code as possible.
+                </Tooltip>
               </li>
               <li>
                 <label>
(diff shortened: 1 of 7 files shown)


site/frontend/src/pages/compare/prefs.ts
```diff
@@ -0,0 +1,10 @@
+import {createStoredValue} from "../../storage";
+
+export const PREF_FILTER_OPENED = createStoredValue(
```
lqd, line 3: Nit: 

```suggestion
export const PREF_FILTERS_OPENED = createStoredValue(
```

plural, like aggregations, since there are a bunch of filters.

site/frontend/src/storage.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  get value(): T {
+    return this._value;
+  }
+
+  store(value: T) {
+    this._value = value;
+    console.debug(`Changing local preference ${this.key} to ${value}`);
```
lqd, line 24: Thanks for leaving that in so I could test more easily :)

site/frontend/src/pages/compare/header/filters.vue
```diff
@@ -29,11 +31,13 @@ watch(
   },
   {deep: true}
 );
+
+const opened = createPersistedRef(PREF_FILTER_OPENED);
```
lqd, line 35: The default value could even depend on the filters' value: if there's a filter that has been changed from its default value (via URL) then the panel should probably be open.

site/frontend/src/storage.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function createStoredValue<T>(
+  key: string,
+  defaultValue: T
+): StoredValue<T> {
+  return new StoredValue(`${prefix}.${key}`, defaultValue);
+}
+
+/**
+ * Creates a reactive variable which state will be persisted to the passed
```
lqd, line 39: nit:
```suggestion
 * Creates a reactive variable whose state will be persisted to the passed
```

IIRC

site/frontend/src/storage.ts
```diff
@@ -0,0 +1,48 @@
+import {ref, Ref, watch} from "vue";
+
+/**
+ * Value that is persisted in LocalStorage.
+ */
+class StoredValue<T> {
+  private _value: T;
+
+  constructor(private key: string, defaultValue: T) {
+    const found = localStorage.getItem(key);
```
lqd, line 10: We may need to first feature-detect whether `localStorage` is both supported and available ? (e.g. browser availability + whether the user has enabled it + some of the usage quota -- maybe more I don't know).

site/frontend/src/storage.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  get value(): T {
+    return this._value;
+  }
+
+  store(value: T) {
+    this._value = value;
+    console.debug(`Changing local preference ${this.key} to ${value}`);
+    localStorage.setItem(this.key, JSON.stringify(value));
```
lqd, line 25: Then here and the `JSON.parse(found)` above, we'd also probably need to try/catch json encoding/decoding, and clear the k/v in case of errors. There could be bugs, related to weird edge cases, values like "" or undefined and so on, or people manually changing the values in the browser, that eagerly clearing the prefs could be some kind of protection.

lqd (commented):
I think this is an improvement. It may be surprising that we'll have a single state value for all open tabs but it seems unlikely to be a real issue.

Left a few comments, r=me after that.

Issue #1622, opened 2023-06-24
Improve rust-timer rollup messages


lqd:
We can improve the messages rust-timer posts on rollup with the list of perf builds for the PRs it contains (random example [here](https://github.com/rust-lang/rust/pull/112702#issuecomment-1594841450)). 

Here are a few thoughts encountered while using the feature:
- the PRs look to be ordered by their number in descending order, which is usually the opposite of what is listed in the OP, and harder to scan.
- the table only contains the PR number but not its title, and makes it a bit hard to scan: to see which build to check, the title of the PRs is usually key when comparing the regression on a given benchmark. For example: the detailed query view of a given benchmark shows a regression in the borrowck query, so we'll look in the list of rolled up PRs for one that changed that area of the compiler. We have to read the title in the OP, remember the number, and go back to the timer message to find the sha1.
- copy-pasting is hard, the SHA1s are linked by GH and one needs to remember to paste without formatting to launch a perf run. The full command could be available in the message to be easy to copy paste.

PR #1623, opened 2023-06-25
Improve PR rollup table formatting

Kobzol:
The context is described in the linked issue. The last commit isn't mentioned in the issue, but it's something that has bothered me :)

Fixes: https://github.com/rust-lang/rustc-perf/issues/1622

This is how it should look like (simulated, as this is not straightforward to test):
image

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -29,9 +29,11 @@ pub async fn unroll_rollup(
     previous_master: &str,
     rollup_pr_number: u32,
 ) -> Result<(), String> {
+    let commit_link = |sha: &str| format!("https://github.com/rust-lang-ci/rust/commit/{sha}");
+
     let format_commit = |s: &str, truncate: bool| {
         let display = truncate.then(|| s.split_at(10).0).unwrap_or(s);
-        format!("[{display}](https://github.com/rust-lang-ci/rust/commit/{s})")
+        format!("[{display}]({})", commit_link(s))
     };
 
     // Sort rolled up commits by their PR number in ascending order, so that they have the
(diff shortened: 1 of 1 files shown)


site/src/github.rs
```diff
@@ -33,8 +33,19 @@ pub async fn unroll_rollup(
...
+    // Sort rolled up commits by their PR number in ascending order, so that they have the
+    // same ordering as in the rollup PR description.
+    let mut unrolled_builds: Vec<UnrolledCommit> =
+        enqueue_unrolled_try_builds(ci_client, rollup_merges, previous_master).await?;
+    unrolled_builds.sort_unstable_by(|a, b| {
+        a.original_pr_number
+            .parse::<u64>()
+            .and_then(|a| b.original_pr_number.parse::<u64>().map(|b| a.cmp(&b)))
+            .unwrap_or_else(|_| a.original_pr_number.cmp(b.original_pr_number))
+    });
```
lqd, line 48: If the strings don't parse correctly, the ordering will be weird anyways, so `sort_by_key` could be enough here ?

```suggestion
unrolled_builds.sort_by_key(|a| a.original_pr_number.parse::<u64>().ok());
```

(or `sort_by_cached_key`)

lqd (approved):
Nice. Thank you.

I'm not sure what changing the branch comment entails, but that seems fine.

site/src/github.rs
Kobzol: Good idea, I changedi t.

Kobzol:
Let's check if this works in the wild.

PR #1624, opened 2023-06-26
Fix Markdown table in rollup messages

Kobzol:
Should fix [this](https://github.com/rust-lang/rust/pull/113049#issuecomment-1607413376).


lqd (approved):
Manually tested the table there, and seems to work.

PR #1627, opened 2023-06-29
Type-check Vue components, not just TypeScript files

Kobzol:
Before, only TypeScript files were checked, but most of the frontend code is actually in Vue components. This PR fixes that, and also resolves a few issues found by the type-checker in Vue components.


lqd:
Looking good. r=me after rebase.

lqd (approved):

Issue #751, opened 2020-08-25
Document commands to delete artifacts from SQLite table


workingjubilee:
Original problem:
> I still can't manage to properly use rustc-perf locally. This time I got this obscure error when trying to look at per-query results when comparing a bench run with itself

And per @Mark-Simulacrum's [answer and request](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/rustc-perf.20error/near/207750179):
> You can delete from the artifact table with sqlite without too much effort
> I should document the commands to do so

PR #1632, opened 2023-07-01
Update documentation on how to remove an artifact from the database

Kobzol:
Just deleting the whole artifact seems like the easiest solution in the end.

Fixes: https://github.com/rust-lang/rustc-perf/issues/751


Kobzol:
We have tried this method and it indeed works successfully, and it was able to re-do the benchmarks for `1.69.0` and `1.70.0`.

Mark-Simulacrum (approved):

PR #1633, opened 2023-07-03
Add explanation why `tendril-0.2.4` is patched in `html5ever`

Kobzol:
Suggested [here](https://github.com/rust-lang/rustc-perf/pull/1629#issuecomment-1617410162).


nnethercote (approved):
Thanks!

PR #1640, opened 2023-07-06
Add raytracer runtime benchmark

Kobzol:
Adapted from https://github.com/jorendorff/rust-raytrace. Slightly modified for edition 2021 and made deterministic by removing randomness.

As mentioned [here](https://github.com/rust-lang/rustc-perf/pull/1638), this is added for more diversity, I'm not necessarily claiming that this benchmark should be in the final suite.


nnethercote (approved):
Nice! I like that this is a real-world kind of program.

PR #1641, opened 2023-07-06
Change how errors are stored in the DB

Kobzol:
Removes the `error_series` table, which was probably useless anyway, and adds a `benchmark` column to the `error` table instead. This will allow us to also store runtime errors in to the crate, e.g. under the name `runtime:foo`.

Tested this locally with both SQLite and Postgre. Old errors were migrated correctly and new errors were also recorded fine. Not sure for how long will the migration run, but hopefully there aren't that many errors in the database.

I had to create a new table in the migration, because it was not possible to drop the previous primary key.

### DIFF
--- a/database/src/bin/postgres-to-sqlite.rs
+++ b/database/src/bin/postgres-to-sqlite.rs
@@ -191,12 +191,12 @@ impl Table for Error {
     }
 
     fn postgres_select_statement(&self, since_weeks_ago: Option<u32>) -> String {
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
Looks like CI is failing?

Kobzol:
Thanks, fixed.

Mark-Simulacrum (approved):

Issue #1345, opened 2022-06-15
Use hardware performance counter data for the detailed/self-profile data view


wesleywiser:
Now that rustc supports using HPC data in `-Zself-profile` (rust-lang/rust#78781), it would be great to use this support on perf.rlo as well. Many of our smaller benchmarks don't run long enough for the `std::time::Instant` based profiling to work reliably which makes it hard to interpret the data when it doesn't really match the results reported on the summary page for a particular benchmark. By using the HPC data, hopefully this will improve the accuracy of detailed data view.

eddyb:
`instructions:u` is also the default for the total counts, so it seems natural to compare it, instead of time - often there is a meaningful change in the total `instructions:u`, but it's lost in time noise in the query view.

Also, assuming it (still) works, I recommend `-Z self-profile-counter=instructions-minus-irqs:u` (instead of the plain `instructions:u`), to avoid weird jitter in otherwise-deterministic queries.

eddyb:
Oh dear, it won't work yet, it's broken without adding `features = ["nightly"]` or updating the version of `measureme` that `rustc` uses (see https://github.com/rust-lang/rust/pull/78781#issuecomment-1165931887).

klensy:
> Opened [rust-lang/rust#98471](https://github.com/rust-lang/rust/pull/98471) to update measureme in rustc to resolve that.

Merged.

Kobzol:
I have tried to implement this, but I don't know how to actually read the HW counter data from the output of `-Zself-profile`. I'm using `-Zself-profile -Zself-profile-counter=instructions:u`, but the output file, when processed with `summarize summarize --json`, only contains information about time (I don't see any counter values).

bjorn3:
I think profiles either contain timestamps or instructions:u values, not both. Did you check that `summarize summarize --json` outputs values that make sense if the counters are time rather than instructions?

Kobzol:
To be honest, I'm not really sure how to recognize that. It's true that when I turn on HW counters, the times seem to be diferent by orders of magnitude. So the counter values just get stored in the nanos attribute of time?

lqd:
The PR introducing the feature in rustc did mention this as somewhat backwards compatible for tools, until they adapt to the new counters. Maybe summarize is in that category.

Kobzol:
Yeah I saw that, but somehow I expected that this adoption has already happened in these 3 years 😅 Maybe not, I'll check how the tools work.

[...]

PR #1647, opened 2023-07-08
Switch self profile to use HW counters instead of walltime

Kobzol:
This PR changes the gathered self profile to use instructions as a metric, instead of wall time. There are some problems with this:

1) This PR simply switches the interpretation of the data to instructions. However, old artifacts will have values stored in the DB as walltime. We could record the metric (instructions/walltime) in the DB, so that we show e.g. the correct units on the detailed query page table, but this does not change the fact that there will be some pairs of artifacts where one side will have walltime, and the other instructions, and these will be basically uncomparable.

2) Instructions will be added together for all threads, which might be a bit misleading where parallelism is used.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1345

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -95,8 +95,9 @@ fn main() {
                 if wrapper == "PerfStatSelfProfile" {
                     cmd.arg(&format!(
                         "-Zself-profile={}",
-                        prof_out_dir.to_str().unwrap()
+                        prof_out_dir.to_str().unwrap(),
                     ));
+                    cmd.arg("-Zself-profile-counter=instructions:u");
                     let _ = fs::remove_dir_all(&prof_out_dir);
                     let _ = fs::create_dir_all(&prof_out_dir);
                 }
--- a/site/frontend/src/pages/detailed-query.ts
+++ b/site/frontend/src/pages/detailed-query.ts
@@ -2,8 +2,8 @@ import {createUrlWithAppendedParams, getUrlParams} from "../utils/navigation";
 import {postMsgpack} from "../utils/requests";
 import {SELF_PROFILE_DATA_URL} from "../urls";
 
-function to_seconds(time) {
-  return time / 1000000000;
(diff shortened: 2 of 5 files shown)


Mark-Simulacrum (approved):
I think this is broadly reasonable - I'm not too worried about the N commits that happen in the transit time, but maybe we can temporarily add a note: message to the HTML noting that? I don't think we have easy ways of tying that back to the deployed version of rustc-perf (though I seem to recall the information being persisted into the DB), so it would just be unconditional.

Adding together across threads also seems like best we can do and is already done for the general view.

Kobzol:
I added the note to the detailed query page and added more documentation about this to the code. I also changed the formatting of values in the self profile table.

Kobzol:
I tried locally that combining `perf stat -e instructions` and `measureme` using instructions doesn't seem to affect the results (I'm not sure if it's OK in general to "nest" PMU counter recording).

@Mark-Simulacrum I don't have any more ideas on how to test this further locally. I guess that we'll just have to merge it and see what happens :) Feel free to do that.

bjorn3:
https://perf.wiki.kernel.org/index.php/Tutorial

> Furthermore, the perf_events interface allows multiple tools to measure the same thread or CPU at the same time.

Kobzol:
Yeah it definitely seems like it's supported, I just wanted to be sure that when I measure the same event (instructions) twice, it won't take two counter slots (to avoid possible multiplexing).

> Events are currently managed in round-robin fashion. Therefore each event will eventually get a chance to run. If there are N counters, then up to the first N events on the round-robin list are programmed into the PMU. In certain situations it may be less than that because some events may not be measured together or they compete for the same counter. Furthermore, the perf_events interface allows multiple tools to measure the same thread or CPU at the same time. Each event is added to the same round-robin list. There is no guarantee that all events of a tool are stored sequentially in the list.

It's not clear to me whether this (counter sharing/multiplexing) can happen when two processes measure the exact same event.

PR #1651, opened 2023-07-12
Store runtime errors into the database

Kobzol:
Stores both compilation and runtime errors of runtime benchmarks into the database. This should be the final step before we can actually execute runtime benchmarks on CI.

### DIFF
--- a/collector/src/runtime/mod.rs
+++ b/collector/src/runtime/mod.rs
@@ -1,7 +1,8 @@
-use std::io::{BufRead, BufReader};
+use std::io::{BufRead, BufReader, Cursor};
 use std::path::Path;
 use std::process::{Command, Stdio};
 
+use anyhow::Context;
 use thousands::Separable;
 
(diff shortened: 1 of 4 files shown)


collector/src/runtime/mod.rs
```diff
@@ -38,41 +39,62 @@ pub async fn bench_runtime(
...
-                    print_stats(&result);
-                    record_stats(
-                        tx.conn(),
-                        collector.artifact_row_id,
-                        &rustc_perf_version,
-                        result,
+
+        // Async block is used to easily capture all results, it basically simulates a `try` block.
+        // Extracting this into a separate function would be annoying, as there would be many
+        // parameters.
```
Mark-Simulacrum, line 53: Why an async block vs a closure?

Mark-Simulacrum (approved):
Small nit but not strictly blocking.

collector/src/runtime/mod.rs
Kobzol: There is an `await` inside the block.

Issue #1578, opened 2023-04-24
Add timeout to benchmarking


Kobzol:
Sometimes the compiler can deadlock or run for a very long time, so it would be nice if we could add a timeout so that the benchmarking run on the server doesn't have to be cancelled manually.

PR #1652, opened 2023-07-13
Add compilation benchmark timeout

Kobzol:
This PR implements basic timeout handling for compilation benchmarks. I needed to refactor some things and make them asynchronous so that we could easily have a timeout for the whole benchmark (e.g. for all configurations - incr/full/... + opt/debug/...), which I think makes more sense, because we can now say something like "everything for this benchmark cannot take more than X", and attach an error to the whole benchmark, same as we do for compilation errors.

In theory we could also implement timeout on individual compilation level, i.e. that a single configuration (benchmark/scenario/profile) cannot take more than X.

I chose the timeout value to be 30 minutes, which should more than enough to find deadlocks or endless compilations. I also tried to remove passing of runtime to `bench_compile`, but it was beyond my lifetime-fu.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1578

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2298,6 +2298,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "signal-hook-registry"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Mark-Simulacrum (approved):
This seems good to me.

PR #1654, opened 2023-07-14
Add perf. run duration estimation to PR comment

Kobzol:
Little QoL improvement that adds a rough estimate of how long it should take until a perf. run result is available. This information is added to the "Queued <commit>, future comparison URL..." comment on PRs.

It's not precise, since the duration of individual runs can vary, sometimes a perf. run can fail soon, and sometimes a PR run "jump ahead" another one, so that it takes longer. But I think that it's nice to know at least a (probable) lower bound of the waiting time, not everyone knows of the status page, and even if you know about it, this will show you its current state more quickly.

The "queued" comment is automatically hidden when the benchmark run finishes, so the outdated/stale "There are currently N other artifacts..." text won't be visible for a long time.


Kobzol:
Done, the queue text is now a link to the status page.

Mark-Simulacrum (approved):
We may want to avoid publishing nonsense about instant collection if there's not a reasonable timestamp, but this seems ok for now.

Kobzol:
I added a min. estimated time of 1 hour, if data is missing or if it's too fast for some reason.

PR #1657, opened 2023-07-15
Store artifact sizes into the DB

Kobzol:
This PR adds a new DB table that records the size of individual components (like `librustc_driver.so`, `libLLVM.so`, ...) into the database. I decided to explicitly list the components that we want to record, instead of just taking the whole `bin`/`lib` directory and somehow removing hashes from the filenames.


Mark-Simulacrum (approved):
Seems like a good starting point to me.

PR #1658, opened 2023-07-15
Add artifact size UI to compare page

Kobzol:
Adds a tab with a simple table containing artifact component size diff to the compare page.

image

### DIFF
--- a/site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
+++ b/site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
@@ -1,0 +1,152 @@
+<script setup lang="ts">
+import {ArtifactDescription} from "../types";
+import {diffClass, formatSize} from "../shared";
+
+const props = defineProps<{
+  a: ArtifactDescription;
+  b: ArtifactDescription;
+}>();
+
+// Sort binaries first, libraries later. Then within each category, sort in descending order by size
+// of the original (a) artifact component.
+const components = Object.entries(props.a.component_sizes)
+  .sort((a, b) => {
+    const aLib = a[0].startsWith("lib");
+    const bLib = b[0].startsWith("lib");
+    if (aLib && !bLib) {
+      return 1;
+    } else if (!aLib && bLib) {
+      return -1;
+    } else {
+      return b[1] - a[1];
+    }
+  })
+  .map((item) => item[0]);
+
+function formatName(component: string): string {
+  if (component.startsWith("lib")) {
+    return `${component}.so`;
+  }
+  return component;
+}
+
+function formatValue(value: number | undefined): string {
+  if (value === undefined) {
+    return "-";
+  }
+  return formatSize(value);
+}
+
+function formatChange(a: number | undefined, b: number | undefined): string {
+  if (a === undefined || b === undefined) {
+    return "-";
+  }
+  const diff = b - a;
+  const formatted = formatSize(Math.abs(diff));
+  if (diff < 0) {
+    return `-${formatted}`;
+  }
+  return formatted;
+}
+
+function formatPercentChange(
+  a: number | undefined,
+  b: number | undefined
+): string {
+  if (a === undefined || b === undefined) {
+    return "-";
+  }
+  return `${(((b - a) / a) * 100).toFixed(3)}%`;
+}
+
+function getClass(a: number | undefined, b: number | undefined): string {
+  if (a === undefined || b === undefined || a == b) {
+    return "";
+  }
+  return diffClass(b - a);
+}
+</script>
+
+<template>
+  <div class="category-title">Artifact component sizes</div>
+  <table>
+    <thead>
+      <tr>
+        <th>Component</th>
+        <th>Before</th>
+        <th>After</th>
+        <th>Change</th>
+        <th>% Change</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr v-for="component in components">
+        <td class="component">{{ formatName(component) }}</td>
+        <td>
+          <div class="aligned">
+            {{ formatValue(a.component_sizes[component]) }}
+          </div>
+        </td>
+        <td>
+          <div class="aligned">
+            {{ formatValue(b.component_sizes[component]) }}
+          </div>
+        </td>
+        <td
+          :class="
+            getClass(a.component_sizes[component], b.component_sizes[component])
+          "
+        >
+          <div class="aligned">
+            {{
+              formatChange(
+                a.component_sizes[component],
+                b.component_sizes[component]
+              )
+            }}
+          </div>
+        </td>
+        <td
+          :class="
+            getClass(a.component_sizes[component], b.component_sizes[component])
+          "
+        >
+          <div class="aligned">
+            {{
+              formatPercentChange(
+                a.component_sizes[component],
+                b.component_sizes[component]
+              )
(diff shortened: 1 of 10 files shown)


nnethercote:
It would be very helpful if there could be tooltips or something similar with brief explanations of each component.

Kobzol:
Good point, I added a short tooltip to each known artifact in the table.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -65,6 +65,29 @@ function getClass(a: number | undefined, b: number | undefined): string {
...
+
+function generateTitle(component: string): string {
+  if (component === "rustc") {
+    return `Main binary of the Rust compiler. It is a small shim that links to librustc_driver.so,
+ which contains most of the compiler logic.`;
+  } else if (component === "rustdoc") {
+    return `Tool that generates documentation of Rust crates. Links to librustc_driver.so, which provides
+ most of the compiler logic.`;
+  } else if (component === "cargo") {
+    return "Build tool that compiles crates and their dependencies.";
```
lqd, line 77: Do we want the tooltips to be what the tool is, or what the artifact is ? I'm hoping people won't learn what LLVM or cargo is from this tooltip, so e.g. "Build tool that compiles crates and their dependencies." vs "cargo's executable. Will be the one from the default toolchain if not present in a locally built toolchain" (just an example).

I also wonder if we should add the little "?" icon to show there's a tooltip, like we do elsewhere.

image

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -0,0 +1,152 @@
...
+import {ArtifactDescription} from "../types";
+import {diffClass, formatSize} from "../shared";
+
+const props = defineProps<{
+  a: ArtifactDescription;
+  b: ArtifactDescription;
+}>();
+
+// Sort binaries first, libraries later. Then within each category, sort in descending order by size
+// of the original (a) artifact component.
```
lqd, line 11: it's cute but the sorting feels a bit too clever: there's no visual separation between bins and libs (there could be an empty row) making the sections and their own sub-order hard to see and breaks scanning vertically; using one of the artifact sizes (which can evolve and result in different orderings between toolchains over time) can feel arbitrary as well as different from how we sort the other similar tables.

maybe it's fine, maybe "bins, empty row, libs -- each section alphabetically sorted" is clear enough: it's good to keep it simpler when presenting data.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -0,0 +1,152 @@
...
+        <th>Change</th>
+        <th>% Change</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr v-for="component in components">
+        <td class="component">{{ formatName(component) }}</td>
+        <td>
+          <div class="aligned">
+            {{ formatValue(a.component_sizes[component]) }}
```
lqd, line 112: since these values are formatted and rounded, it could be nice to display the full value on a tooltip: we may have no difference in the before/after columns, lost in the rounding after the 2nd digit, but some size difference in the "change" column.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function formatPercentChange(
+  a: number | undefined,
+  b: number | undefined
+): string {
+  if (a === undefined || b === undefined) {
+    return "-";
+  }
+  return `${(((b - a) / a) * 100).toFixed(3)}%`;
```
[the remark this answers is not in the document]
Kobzol: Good point. I will add a check for both undefined and `0`.

site/frontend/src/pages/compare/tabs.vue
```diff
@@ -119,6 +137,30 @@ const activeTab: Ref<Tab> = ref(props.initialTab);
...
+          {{ formatArtifactSize(totalSizeA) }} ->
+          {{ formatArtifactSize(totalSizeB) }}
+        </div>
+        <div
+          v-if="bothSizesAvailable"
+          :class="diffClass(totalSizeB - totalSizeA)"
+        >
+          {{ totalSizeB < totalSizeA ? "-" : ""
+          }}{{ formatSize(Math.abs(totalSizeB - totalSizeA)) }} ({{
+            (((totalSizeB - totalSizeA) / totalSizeA) * 100).toFixed(3)
```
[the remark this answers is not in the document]
Kobzol: I extracted the function, good idea. This code won't be called if `totalSizeA` is ever zero though.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -65,6 +65,29 @@ function getClass(a: number | undefined, b: number | undefined): string {
...
+    return `Main binary of the Rust compiler. It is a small shim that links to librustc_driver.so,
+ which contains most of the compiler logic.`;
+  } else if (component === "rustdoc") {
+    return `Tool that generates documentation of Rust crates. Links to librustc_driver.so, which provides
+ most of the compiler logic.`;
+  } else if (component === "cargo") {
+    return "Build tool that compiles crates and their dependencies.";
+  } else if (component === "librustc_driver") {
+    return `Shared library which contains the core implementation of the compiler. It is used by several
+ other tools and binaries.`;
```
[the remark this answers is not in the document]
Kobzol: Good point. I did that manually, but just because of source code aesthetics.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
Kobzol: Again, very good points :)

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
Kobzol: I added a new column to the table (Library/Binary) and changed ordering to be "binaries first, then alphabetically".

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
Kobzol: Good point, added titles.

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
```diff
@@ -64,22 +64,19 @@ function getClass(a: number | undefined, b: number | undefined): string {
 
 function generateTitle(component: string): string {
   if (component === "rustc") {
-    return `Main binary of the Rust compiler. It is a small shim that links to librustc_driver.so,
- which contains most of the compiler logic.`;
+    return `Executable of the Rust compiler. A small wrapper that links to librustc_driver.so, which provies most of the compiler logic.`;
```
lqd, line 78: tiny typo for "provides"

lqd (approved):

site/frontend/src/pages/compare/artifact-size/artifact-size-table.vue
Kobzol: Thanks!

PR #1659, opened 2023-07-17
Add timeout to master commit list download

Kobzol:
Before, there was no timeout, so if the remote side wasn't responding, or the connection became stuck somehow, the server could wait without any upper bound.


Kobzol:
Alternatively, we could apply a timeout to the whole initialization of the server (DB and index load + master commit download).

Mark-Simulacrum (approved):
Seems ok, I guess we have some functionality that should continue to work despite this being broken.

Kobzol:
The collector isn't affected by this, but the site doesn't start if this times out. So hopefully after this PR it won't happen.

PR #1661, opened 2023-07-18
Check that `Cargo.lock` doesn't change

Kobzol:
Recently I saw some cases where `Cargo.lock` can be updated during a cargo build, and I realized that it wouldn't be nice if this situation happens when building benchmarks, as something could then change the code being benchmarked. This PR adds a check that makes sure that `Cargo.lock` stays the same before and after a compile/runtime benchmark is built.

I tried it on our suite and some lockfiles are actually changed, but hopefully only because of old `Cargo.lock` manifest format. In any case, it would be nice to catch these situations in the future, I suppose.

When the lockfiles need to be changed across the board (e.g. when a new manifest format is introduced by Cargo), we'd need to update all of them in the repository, and until that perf. runs would probably fail. But I guess (hope) that this situation isn't very common.

### DIFF
--- a/collector/src/compile/execute/mod.rs
+++ b/collector/src/compile/execute/mod.rs
@@ -5,6 +5,7 @@ use crate::compile::benchmark::profile::Profile;
 use crate::compile::benchmark::scenario::Scenario;
 use crate::compile::benchmark::BenchmarkName;
 use crate::toolchain::Toolchain;
+use crate::utils::fs::EnsureImmutableFile;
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum (approved):
I think it seems pretty safe to do this, Cargo shouldn't change lockfiles often on its own. It's a bug if it does, and hopefully would be quickly caught since it'll be in the cargo bump PR that things start failing for us.

PR #1664, opened 2023-07-19
Update license field following SPDX 2.1 license expression standard

frisoft:
The new recommendation is to follow the SPDX 2.1 standard. This allows automatic license verification software to work properly. Reference: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields


Kobzol:
Hi! :) As noted e.g. [here](https://github.com/rust-lang/rustc-perf/pull/1356), the `collector/compile-benchmarks` directory contains downloaded/copy-pasted code of benchmarks downloaded from the internet, and we don't modify it.

So unless there is some specific use-case that requires this code to be changed, I'd like to avoid modifying this part of the repository.

frisoft:
You are using an old standard. The Rust community moved to the new standard. I can't see any reason to keep using the old one, but I would say it is up to you.

Kobzol:
I get your point, but as I tried to explain before, the code that you have changed is not really a part of this repository's sourcecode. You can think of it as an opaque binary artifact that is included in this repository as a black box. Once it's included, it's basically read only and serves only for benchmarking compilation time.

If you want to make these license changes, you should make PRs against the specific projects that you have modified. Although I suspect that they have already been fixed in the meantime (e.g. [Cargo](https://github.com/rust-lang/cargo/blob/263d24baca913d7ace29e2f59a3254e47272fd6b/Cargo.toml#L15)).

frisoft:
I see. My bad. You are right. In this case it does not make sense updating the license of the crates.

PR #1672, opened 2023-07-27
Separate a comparisons table component and use it in compile and runtime tabs

Kobzol:
Refactoring to reduce the amount of duplicated code in the compare page.

### DIFF
--- a/site/frontend/package-lock.json
+++ b/site/frontend/package-lock.json
@@ -1803,49 +1803,49 @@
       }
     },
     "node_modules/@vue/compiler-core": {
-      "version": "3.2.47",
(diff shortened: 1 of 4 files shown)


Kobzol:
I would still be glad for your opinion, but on second thought, maybe this is just forcing an abstraction where it doesn't belong and we shouldn't do this. If we wanted to add new complex behavior to the individual tables (and I have some ideas here), then it would be quite complicated to keep this as a unifie component.

lqd:
We can keep them separate to allow for parallel evolution until we know it would be helpful to regroup them: if we ever have to add features to a common core a few times, then we'd add the abstraction less prematurely and reduce duplication.

Kobzol:
I agree. Closing this for now.

Issue #1665, opened 2023-07-25
ICE when building match-stress


klensy:
Didn't bisected where that started, but https://perf.rust-lang.org/status.html
<details>
```
Cannot compile match-stress

Caused by:
    expected success, got exit status: 101
    
    stderr= Documenting match-stress v0.1.0 (/tmp/.tmpUt7JXo)
    warning: unclosed HTML tag `baz`
       --> src/macros.rs:39:23
        |
    39  |                       $(#[$variant_attr])*
        |                         ^^^^^^^^^^^^^^^^
        |
       ::: src/expr.rs:15:1
        |
    15  | / ast_enum_of_structs! {
    16  | |     pub enum ExprKind {
    17  | |         /// A `box x` expression.
    18  | |         pub Box(ExprBox {
    ...   |
    348 | |     }
    349 | | }
        | |_- in this macro invocation
        |
        = note: `#[warn(rustdoc::invalid_html_tags)]` on by default
        = note: this warning originates in the macro `ast_enum_of_structs` (in Nightly builds, run with -Z macro-backtrace for more info)
    help: try marking as source code
        |
    39  |                     $(`#[$variant_attr]`)*
        |                       +                +
    
    warning: unclosed HTML tag `baz`
       --> src/macros.rs:6:11
        |
    6   |           $(#[$attr])*
        |             ^^^^^^^^
        |
       ::: src/expr.rs:15:1
        |
    15  | / ast_enum_of_structs! {
    16  | |     pub enum ExprKind {
    17  | |         /// A `box x` expression.
    18  | |         pub Box(ExprBox {
    ...   |
    348 | |     }
    349 | | }
        | |_- in this macro invocation
        |
        = note: this warning originates in the macro `ast_struct` which comes from the expansion of the macro `ast_enum_of_structs` (in Nightly builds, run with -Z macro-backtrace for more info)
    help: try marking as source code
        |
    6   |         $(`#[$attr]`)*
        |           +        +
    
    thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', src/librustdoc/html/format.rs:818:37
    stack backtrace:
       0:     0x7ffff3b26fc1 - std::backtrace_rs::backtrace::libunwind::trace::hf36f05c203408ba8
                                   at /rustc/e55147822ced7759651867efebb824fff4c64a43/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
[...]

PR #1675, opened 2023-07-27
Port status page to Vue

Kobzol:
This will make it easier to make changes (e.g. to fix issue https://github.com/rust-lang/rustc-perf/issues/1665). This PR first ports the status page as it was (apart from some tiny code changes, or removing cruft that's no longer relevant, like try commits without a proper date), I will make further changes in follow-up PRs.

To test this, it will probably be easiest to use data from the live server, by changing `BASE_URL` in `site/frontend/urls.ts` to this:
```typescript
const BASE_URL = "https://perf.rust-lang.org/perf";
```

Then you can open the page locally and on the live server and compare that if it looks the same (it should) :)


lqd (commented):
> To test this, it will probably be easiest to use data from the live server

Yeah agreed, but that also means there was a limit in the number of states I could check: I couldn't easily compare benchmark errors, some states like both "collection in-progress" and "collection complete", the "empty queue" state, missing commits, etc 😓 

But it's ok. We'll keep an eye on these rarer states and see if we missed anything then. It should be easy to fix if we can reproduce issues when they happen. The main Status functionality works great, so that's the most important use cases taken care of.

lqd (approved):

PR #1678, opened 2023-07-27
Show `dev` profile metadata for benchmarks

lqd:
This PR extends #1626 to also show benchmark metadata from the `dev` profile for `Check` and `Debug` benchmarks. 

Even the dev profile has interesting information for some benchmarks: while `ripgrep` uses debuginfo in release, `exa` uses no debuginfo in debug...

### DIFF
--- a/site/build.rs
+++ b/site/build.rs
@@ -55,22 +55,11 @@ fn main() -> Result<(), Box<dyn Error>> {
         let manifest: toml::Value = toml::from_str(&manifest_contents)?;
 
         let table = manifest.as_table().unwrap();
-        let profile = table.get("profile");
-        let release = profile
-            .and_then(|p| p.as_table())
-            .and_then(|t| t.get("release"))
-            .and_then(|t| t.as_table());
-        let debug = release.and_then(|t| t.get("debug"));
-        let lto = release.and_then(|t| t.get("lto"));
-        let codegen_units = release.and_then(|t| t.get("codegen-units"));
-
+        let profiles = table.get("profile");
         let metadata = CompileBenchmarkMetadata {
             perf_config: config,
-            release_metadata: ProfileMetadata {
-                debug: debug.map(|v| v.to_string()),
-                lto: lto.map(|v| v.to_string()),
-                codegen_units: codegen_units.and_then(|v| v.as_integer().map(|v| v as u32)),
-            },
+            release_metadata: read_profile_metadata(profiles, "release"),
+            dev_metadata: read_profile_metadata(profiles, "dev"),
         };
         suite.insert(benchmark_name, metadata);
     }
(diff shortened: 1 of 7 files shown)


site/build.rs
```diff
@@ -82,3 +71,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        .and_then(|p| p.as_table())
+        .and_then(|t| t.get(profile))
+        .and_then(|t| t.as_table());
+    let debug = profile.and_then(|t| t.get("debug"));
+    let lto = profile.and_then(|t| t.get("lto"));
+    let codegen_units = profile.and_then(|t| t.get("codegen-units"));
+    ProfileMetadata {
+        debug: debug.map(|v| v.to_string()),
+        lto: lto.map(|v| v.to_string()),
+        codegen_units: codegen_units.and_then(|v| v.as_integer().map(|v| v as u32)),
```
Kobzol, line 91: INB4 we add `codegen-units = "auto"` :laughing: Maybe we should just treat everything as a black box and store it as a string. But it's not super important (and pre-existing of course, I love to nitpick my own old code).

site/frontend/src/pages/compare/compile/comparisons-table.vue
```diff
@@ -86,6 +86,11 @@ Category: ${metadata.category}
   };
   if (testCase.profile === "opt" && metadata.release_profile !== null) {
     addMetadata(metadata.release_profile);
+  } else if (
+    (testCase.profile === "check" || testCase.profile === "debug") &&
```
Kobzol, line 90: I'm not sure if we should display these details on `check` builds. Yes, they are treated as `dev` builds, but LTO/debug/CGUs are not (?) relevant for `check` builds, so this can be a bit confusing for whoever displays this data and might think that LTO did affect something for a `check` build.

Kobzol (commented):
Looking great! I had just one question about `check` builds, otherwise LGTM.

site/frontend/src/pages/compare/compile/comparisons-table.vue
lqd: I wasn't sure myself that this was a good move: I thought there may be some metadata we'll add in the future that will affect check builds (e.g. turning on debug assertions, some lints, etc).

It's probably not going to happen any time soon, so we can think about it then. I'll limit this to debug builds for now then.

site/build.rs
lqd: We could indeed. I think if we display more metadata in the future, we could simplify this a bit. For now it seems fine, as it's not clear we'll change this code in the short term?

Kobzol: Yeah that was written mostly in jest :) Let's keep it as it is for now.

Kobzol (approved):

PR #1684, opened 2023-08-02
Fetch origin HEAD on checkout

Mark-Simulacrum:
This ensures that git merge-base can find the right merge point; otherwise, our history looks to have diverged when the original checkout was made, which might be years ago.

cc @RalfJung https://github.com/rust-lang/rust/issues/101907

### DIFF
--- a/collector/src/compile/execute/rustc.rs
+++ b/collector/src/compile/execute/rustc.rs
@@ -165,19 +165,23 @@ fn checkout(artifact: &ArtifactId) -> anyhow::Result<()> {
             .status()
             .context("git fetch origin")?;
 
-        if !status.success() && matches!(artifact, ArtifactId::Tag(_)) {
+        if !status.success() {
             log::warn!(
-                "git fetch origin {} failed - trying default branch",
+                "git fetch origin {} failed, this will likely break the build",
                 artifact
             );
-            status = Command::new("git")
-                .current_dir("rust")
-                .arg("fetch")
-                .arg("origin")
-                .arg("HEAD")
-                .status()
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Before we merge we should do some testing to try to figure out the right incantation. I think HEAD might do the wrong thing here, we might need `git fetch origin master` as the invocation...

RalfJung:
Oh so this has an `origin/master` branch, but it is ancient, and new things are fetched in a way that does not update that remove? Yeah that would cause problems.

Just not having that branch is the other alternative, then my PR will fall back to the old logic.

collector/src/compile/execute/rustc.rs
```diff
@@ -165,19 +165,23 @@ fn checkout(artifact: &ArtifactId) -> anyhow::Result<()> {
...
         }
+
+        // Regardless, we fetch the default branch. Upstream Rust started using `git merge-base`
+        // recently, which (reasonably) finds the wrong base if we think e.g. origin/master
+        // diverged thousands of commits ago.
+        status = Command::new("git")
+            .current_dir("rust")
+            .arg("fetch")
+            .arg("origin")
+            .arg("HEAD")
```
RalfJung, line 182: I think you are right, this needs to be master to update the `origin/master` ref -- I just checked this locally in another git repo: `git fetch origin HEAD` fetches the latest commits but does not update where `origin/master` points to.

Mark-Simulacrum, line 182:
```suggestion
            .arg("master")
```

lqd (approved):

PR #1694, opened 2023-08-09
Move comparison table links to a separate section

Kobzol:
I wanted to do this for a long time :grin: This PR moves all the "hidden" links that were previously more or less randomly scattered across the various columns of the comparison table to an explicit list of links in the new benchmark detail section.

Before there were these links:
- Benchmark name column: link to source code
- Profile column: link to 30 day graph
- Change column: link to self profile diff
- Raw value (before): link to self profile of commit A
- Raw value (after): link to self profile of commit B

The new link section contains all links except for the 30 day graph, which is now displayed directly in the benchmark detail. I can add an explicit link to the graph there too, if it's considered useful.

This PR changes the "distance" to go to each link from one click to two click (open row, click on link). For source code, I don't think it's that important, since it's not used much I think. For the 30 day graph, it's now actually better, since the graph is displayed directly on the page. For the self-profile links, this is a "regression", so to speak, but I still consider it to be a better trade-off to make the links more discoverable.

image

After removing the links from the columns, I made the whole row clickable, and gave it a slight background tint on hover and when it is toggled. This is bikesheddable of course :)

### DIFF
--- a/site/frontend/src/components/cachegrind-cmd.vue
+++ b/site/frontend/src/components/cachegrind-cmd.vue
@@ -1,0 +1,61 @@
+<script setup lang="ts">
+import {CompileTestCase, Profile} from "../pages/compare/compile/common";
+import {computed} from "vue";
+
+const props = defineProps<{
+  commit: string;
+  testCase: CompileTestCase;
+  baseline_commit?: string;
+}>();
+
+const firstCommit = computed(() => {
+  if (props.baseline_commit !== undefined) {
+    return props.baseline_commit;
+  } else {
+    return props.commit;
+  }
+});
+
+function normalizeProfile(profile: Profile): string {
+  if (profile === "opt") {
+    return "Opt";
+  } else if (profile === "debug") {
+    return "Debug";
+  } else if (profile === "check") {
+    return "Check";
+  } else if (profile === "doc") {
+    return "Doc";
+  }
+  return "<invalid profile>";
+}
+function normalizeScenario(scenario: string): string {
+  if (scenario === "full") {
+    return "Full";
+  } else if (scenario === "incr-full") {
+    return "IncrFull";
+  } else if (scenario === "incr-unchanged") {
+    return "IncrUnchanged";
+  } else if (scenario.startsWith("incr-patched")) {
+    return "IncrPatched";
+  }
+  return "<invalid scenario>";
+}
+</script>
+
+<template>
+  <pre><code>cargo run --bin collector profile_local cachegrind \
+    +{{ firstCommit }} \<template v-if="props.baseline_commit !== undefined">
+    --rustc2 +{{ props.commit }} \</template>
+    --include {{ testCase.benchmark }} \
+    --profiles {{ normalizeProfile(testCase.profile) }} \
+    --scenarios {{ normalizeScenario(testCase.scenario) }}</code></pre>
+</template>
+
+<style scoped lang="scss">
+pre {
+  background-color: #eeeeee;
+}
+code {
+  user-select: all;
+}
+</style>
--- a/site/frontend/src/components/cachegrind-cmd.vue
+++ b/site/frontend/src/components/cachegrind-cmd.vue
@@ -43,7 +43,7 @@ function normalizeScenario(scenario: string): string {
 </script>
 
 <template>
-  <pre><code>cargo run --bin collector profile_local cachegrind \
+  <pre><code>cargo run --bin collector --release profile_local cachegrind \
     +{{ firstCommit }} \<template v-if="props.baseline_commit !== undefined">
     --rustc2 +{{ props.commit }} \</template>
     --include {{ testCase.benchmark }} \
--- a/site/frontend/src/components/cachegrind-cmd.vue
+++ b/site/frontend/src/components/cachegrind-cmd.vue
@@ -5,12 +5,12 @@ import {computed} from "vue";
 const props = defineProps<{
   commit: string;
   testCase: CompileTestCase;
-  baseline_commit?: string;
+  baselineCommit?: string;
 }>();
 
 const firstCommit = computed(() => {
-  if (props.baseline_commit !== undefined) {
-    return props.baseline_commit;
+  if (props.baselineCommit !== undefined) {
+    return props.baselineCommit;
   } else {
     return props.commit;
   }
@@ -44,7 +44,7 @@ function normalizeScenario(scenario: string): string {
 
 <template>
   <pre><code>cargo run --bin collector --release profile_local cachegrind \
-    +{{ firstCommit }} \<template v-if="props.baseline_commit !== undefined">
+    +{{ firstCommit }} \<template v-if="props.baselineCommit !== undefined">
     --rustc2 +{{ props.commit }} \</template>
     --include {{ testCase.benchmark }} \
     --profiles {{ normalizeProfile(testCase.profile) }} \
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
I think an explicit graph link is useful to cite in comments and such, and seems cheap to retain at least for now.

Kobzol:
I agree. I added the link back and moved the source code to the end:
image

Kobzol:
I think that it's more practical to open the benchmark detail by clicking on the whole row (anywhere in the row), vs opening it only by clicking on the tiny arrow on the left. And if you make the whole row clickable, then I think that it's weird that based on exactly where you click in the row, you either open the detail or open a new tab (it would basically then form a sort of a "double link" on a single click).

RalfJung:
What are these 3 "self profile" links? Currently there is a single link to a page like [this](https://perf.rust-lang.org/detailed-query.html?commit=94b77248c5aa991f2597b4a1c1a49e297ac2f3e2&base_commit=3071e0aef6dfd0a150c3fb1da0abad4ec86ca0aa&benchmark=ctfe-stress-5-debug&scenario=full&sort_idx=-11), does that correspond to one of the 3 new links?

"self profile" is very unclear as a name, I have no idea what it means.

RalfJung:
One of the very useful things on that page are the instructions for how to get cachegrind details on a perf regression. I'd never click on "self profile" if that's what I am looking for.

Also, it's not clear that "self profile" means "rustc self profile". Just "rustc profile" would be a lot more clear IMO.

>  If you enable the "Display raw data" checkbox in the filter section, you can then click on the Before column to get to the detailed query page for the original commit, or on the After column value to get to the detailed query page for the new commit. And if you click on the value in the % Change column, you will get to the detailed query page containing a diff between the two commits laughing So there were already three links, but obviously they weren't very easy to find.

Okay. Sounds like I would almost always want the "diff" page -- that should be made clear somehow.

Kobzol:
Indeed, the diff page is the most useful one, that's also why it's in the first position out of the three links.

I guess that it could be called something like "Investigate [this perf. change]", heh.

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -215,6 +243,42 @@ onMounted(() => renderGraph());
...
     </div>
+    <div class="links">
+      <div class="title bold">Links</div>
+      <ul>
+        <li>
+          <a
+            :href="graphLink(props.artifact, props.metric, props.testCase)"
+            target="_blank"
+          >
+            History graph
```
lqd, line 254: Maybe we could name that link "history: 30 days up to commit" or something, so that we can have links to graphs up to the commit, and without an end date i.e. up to now -- to easily see the future behavior after the PR with the possibility to change the graph kind.

(This doesn't need to be in this PR.)

Kobzol: Added to my TODO list.

lqd: I have a local branch doing it locally, and will open a PR soon after this one lands.

Kobzol:
I added the profiling command and modified the layout to avoid wrapping of the links:
image

RalfJung:
That's the "diff" profile, right?

I still think "detailed results" or so would capture the page contents better than "self profile". There's other stuff there besides the self profile, like flamegraphs.

Kobzol:
Yes, it's the diff. `Detailed results` sounds good.

```
- Detailed results (diff)
- Detailed results (before)
- Detailed results (after)
```
?

lqd:
Ok sure, "detailed results" if you want.

Note that the flamegraphs are only flamegraphs of *the self-profile data*.

site/frontend/src/components/cachegrind-cmd.vue
```diff
@@ -0,0 +1,61 @@
...
+    return "IncrUnchanged";
+  } else if (scenario.startsWith("incr-patched")) {
+    return "IncrPatched";
+  }
+  return "<invalid scenario>";
+}
+</script>
+
+<template>
+  <pre><code>cargo run --bin collector profile_local cachegrind \
```
lqd, line 46: I assume it's not slower or impactful to run collection in debug just to launch a few cargo and valgrind commands, rather than release? We tell people to run the collector in --release today.

Kobzol: I usually just everything in debug locally, I don't think that it should affect a lot of things. But of course, running in release is theoretically *safer*. I can add `--release` here if you want.

lqd: Yes let's, just because otherwise the detailed query page and this command would be different. We can figure out which of the two is best in the future if need be, but until then let's keep it consistent.

lqd (approved):
r=me on the change with the couple nits fixed (links order, --release, rough link naming) once @RalfJung is happy with the naming

(the "GH r+" should hold even if you add commits in the future IIRC)

RalfJung:
> Note that the flamegraphs are only flamegraphs of the self-profile data.

Ah hm. So is the profiling instruction the only non-self-profile thing there?

Is there some way we can just entirely move the profiling command out of that page? Like, the expanded per-benchmark thing with this PR already has the profiling command for the diff. Maybe that could be a little drop-down menu to also offer the profiling command for the old/new commit?

Kobzol:
Yeah it's the only non self-profile thing there. Adding a drop-down menu to the benchmark detail is a good idea, but I would still also keep the commands on the detailed query page, as it should serve as a one-stop-shop for analyzing the performance results in a greater detail.

lqd:
(This is already described in the documentation but you can also change the command locally: the diff command has both sha1s, with `--rustc2` being the benchmarked commit, and you just change the `+sha` without `--rustc2`.)

My initial suggestion was that we can just show all 3 cachegrind commands in the per-benchmark details (with less new lines) much like we do on the detailed query page.

Kobzol:
It's too wide for the current layout on a single line though:
image

It would have to be really tiny (`font-size: 8px`) to fit:
image

I added the drop-down box (sorry for the visual artifacts):
select

Kobzol:
The dropdown has been implemented and the link renamed to `Detailed results`, as Ralf has suggested. I'll merge this PR now, because we're both leaving with @lqd for a vacation, and we want to have this thing active in the next few weeks so that we can see if someone complains about the new link location.

@RalfJung If you have any other suggestions, feel free to write them here or create a new issue on the issue tracker, and I'll implement them once I get back.

RalfJung:
Thanks, this is certainly a lot better!

> I would still also keep the commands on the detailed query page, as it should serve as a one-stop-shop for analyzing the performance results in a greater detail.

Okay, so it is the "detailed results" page then, not just "self profile".

I find it confusing though that there are 3 of these pages (old, new, diff), but the diff page alone lists 3 local profile commands. Shouldn't there be one local profile command for each page?

lqd:
Noted, we'll fix that label when we update the page in the future: "new" there likely means "the commit whose results are shown on this page". 

The "base" and "new" pages are showing the results of a single commit, so there's only 1 profiling command you could run with that. Only the diff page knows the 2 commits, and therefore is able to show the 3 commands.

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -149,88 +179,209 @@ const cargoProfile = computed((): CargoProfileMetadata => {
...
+              <a
+                :href="graphLink(props.artifact, props.metric, props.testCase)"
+                target="_blank"
+              >
+                History graph
+              </a>
+            </li>
+            <li>
+              <a :href="detailedQueryLink(props.baseArtifact)" target="_blank">
+                Rustc self-profile: baseline commit
```
RalfJung, line 281: I feel like this could also be "Detailed results: baseline commit", and dito for the next line, to indicate the parallel to the other "Detailed results" page.

PR #1695, opened 2023-08-10
Add runtime benchmark cachegrind diff

Kobzol:
This PR adds the ability to compare two Cachegrind profiles of a runtime benchmark using `cg_diff`. It also refactors similar functionality for compile time benchmarks, to allow reusing the implementation and remove some unneeded temporary files. The need to have an external `rustfilt` binary installed was also removed, demangling is now performed using the `rustc-demangle` crate.


nnethercote (approved):
👍

PR #1707, opened 2023-08-15
Allow zooming with mouse wheel in graphs

Kobzol:
I took the mouse wheel zoom code from https://leeoniya.github.io/uPlot/demos/index.html.

https://github.com/rust-lang/rustc-perf/assets/4539057/03c69c61-4fae-4c2f-9f99-ad9d650f75f8

### DIFF
--- a/site/frontend/src/graph/render.ts
+++ b/site/frontend/src/graph/render.ts
@@ -147,6 +147,119 @@ function tooltipPlugin({
   };
 }
 
+// Taken from https://leeoniya.github.io/uPlot/demos/zoom-wheel.html
+function wheelZoomPlugin(opts) {
+  let factor = opts.factor || 0.75;
+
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
This seems okay. One question: does this break scrolling through the page as a whole? Do you need to stay "between" graphs or so?

Kobzol:
Yes, you do indeed need to stay "outside of the graphs" to scroll the page.

Mark-Simulacrum (approved):
I'm okay trying this but we may want to revert if the scrolling interference is too bad.

Kobzol:
Ok, let's see :)

lqd:
Once people have had a chance to experience it, let's gather feedback. (As of now, scrolling interference is somewhat noticeable for me, we'll see over time, or how others feel about it)

PR #1708, opened 2023-08-15
Show future dates in noise chart as shaded region

### LINKED ISSUE #1705: Vertical dotted line in 30 day history makes it hard to see data value
I *love* the addition of the 30 day graph to the benchmark detail that I now see in the compare page. It really helps provide some much needed context when trying to interpret some result that's been flagged.

However, the current interface highlights the specific commit under comparison with a dotted line that obscures the actual data value in the 30 day view, as shown here:

 
image

(In the  above somewhat zoomed in view, you *can* make out where there are gaps in the dottted line and thus infer a rough idea of where the value lies. But in my web browser, a bit more zoomed out, its harder to tell for sure where the value was...)

I'm torn about this: Its a really **good** idea to highlight where the commit lies.

but can we do it in a way that doesn't obscure the value its been assigned? Maybe can we put a short horizontal line-segment that intersects this vertical line at the place corresponding to the measured value for the commit? (Or maybe even just **choose a different color for the dotted line, like blue or something?**)

lqd:
This is what I had in mind when I mentioned using a shaded region instead of the dashed line. 

Here's a screenshot with multiple scenarios, to have some idea in comparison to the different foreground colors that they all have.

image

r? @Kobzol 

Closes #1705, cc @pnkfelix who opened the issue.

### DIFF
--- a/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
+++ b/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
@@ -207,6 +207,10 @@ onMounted(() => renderGraph());
         <div style="font-size: 0.8em">
           Each plotted value is relative to its previous commit
         </div>
+        <div style="font-size: 0.8em">
+          The shaded region shows values that are more recent than the
+          benchmarked commit
+        </div>
       </div>
       <div ref="chartElement"></div>
     </div>
--- a/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
+++ b/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
@@ -74,26 +74,21 @@ function getGraphRange(artifact: ArtifactDescription): GraphRange {
 }
 
 /**
- * Hook into the uPlot drawing machinery to draw a vertical line at the
- * position of the given `date`.
+ * Hook into the uPlot drawing machinery to draw a rectangle from the `date` to
+ * the end of the plot, representing the region that is the date's future.
  */
 function drawCurrentDate(opts: GraphRenderOpts, date: Date) {
   opts.hooks = {
     drawSeries: (u: uPlot) => {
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -210,7 +217,11 @@ onMounted(() => renderGraph());
       <div class="title">
         <div class="bold">{{ getGraphTitle() }}</div>
         <div style="font-size: 0.8em">
-          Each plotted value is relative to its previous commit
+          Each plotted value is relative to its previous commit.
+        </div>
+        <div style="font-size: 0.8em">
+          A shaded region shows values that are more recent than the profiled
```
Kobzol, line 223: nit: maybe benchmarked instead of profiled? But no strong objections to keeping profiled.

Kobzol (approved):
This looks awesome! I would just remove the dashed line altogether, I don't think that it provides value in addition to the shaded region (maybe could be useful to people with CVD?).

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
lqd: Sure.

lqd:
> I would just remove the dashed line altogether

Done.

Kobzol:
Hmm, I have now opened the screenshots on a different monitor and here I can barely see them :sweat_smile: While on the previous monitor it was fine. Could you please make the shade a bit less transparent? Alpha 0.1 is perfect, but 0.06 is already much better.

lqd:
> Could you please make the shade a bit less transparent?

sure

Kobzol:
Thanks!

Kobzol (approved):

PR #1717, opened 2023-09-11
Improve benchmark detail graph on mobile

Kobzol:
This improves how the graph in the benchmark detail on the compare page is displayed on mobile devices.

Note that in Chrome's device view `window.innerWidth` returns some weird value. But on mobile devices the width should be correct hopefully.


lqd (approved):
Nice! The graph is looking better in the couple devices I tried.

Issue #1713, opened 2023-08-31
Suspicious time diff at https://perf.rust-lang.org/status.html in took/expected


klensy:
Time displayed at https://perf.rust-lang.org/status.html for took/expected heavily differs for some runs, while perf result shows no difference.

For example, for run https://github.com/rust-lang/rust/pull/115388#issuecomment-1701235905

Step | Took | Expected
-- | -- | --
await-call-tree | 0m28s | 0m28s
bitmaps-3.1.0 | 1m06s | 0m59s
cargo-0.60.0 | 8m24s | 5m23s
clap-3.1.6 | 1m43s | 1m32s
coercions | 0m56s | 0m54s
cranelift-codegen-0.82.1 | 9m50s | 3m02s
ctfe-stress-5 | 1m24s | 1m11s
deeply-nested-multi | 2m43s | 0m32s
deep-vector | 1m34s | 1m24s
derive | 4m13s | 0m49s
diesel-1.4.8 | 7m15s | 2m30s
exa-0.10.1 | 3m00s | 3m00s
externs | 0m38s | 0m39s
helloworld | 0m37s | 0m37s
helloworld-tiny | 0m11s | 0m11s
html5ever-0.26.0 | 1m18s | 1m08s
hyper-0.14.18 | 2m11s | 2m10s
image-0.24.1 | 2m00s | 2m01s
issue-46449 | 1m15s | 1m16s
issue-58319 | 0m32s | 0m32s
issue-88862 | 0m32s | 0m32s
libc-0.2.124 | 1m01s | 1m02s
many-assoc-items | 0m57s | 0m59s
match-stress | 0m51s | 0m52s
projection-caching | 0m41s | 0m41s
regex-1.5.5 | 2m58s | 2m56s
regression-31157 | 0m54s | 1m01s
ripgrep-13.0.0 | 2m04s | 2m02s
ripgrep-13.0.0-tiny | 1m05s | 1m04s
runtime:bufreader | 0m11s |  
runtime:compression | 0m43s |  
runtime:css | 0m15s |  
runtime:fmt | 0m32s |  
runtime:hashmap | 0m50s |  
runtime:nbody | 0m26s |  
runtime:parsing | 0m15s |  
runtime:raytracer | 0m24s |  
runtime:svg | 0m32s |  
runtime:text-search | 0m19s |  
rustc | 13m43s |  
serde-1.0.136 | 1m33s | 1m33s
[...]

PR #1718, opened 2023-09-11
Redesign status page

Kobzol:
This PR redesigns the status page so that we can see the history of runs, to have better visibility of historical benchmark errors, and also to have a more obvious estimation of when will a given PR finish.

image

This patch can be used to hardcode data for testing locally:
<details>
<summary>Patch</summary>

```
diff --git a/site/src/request_handlers/status_page.rs b/site/src/request_handlers/status_page.rs
index e1e2563e..f2ae5f04 100644
--- a/site/src/request_handlers/status_page.rs
+++ b/site/src/request_handlers/status_page.rs
@@ -1,18 +1,46 @@
+use database::{Commit, CommitType, Date};
 use std::str;
 use std::sync::Arc;
 
 use crate::api::status;
-use crate::api::status::FinishedRun;
+use crate::api::status::{CurrentState, FinishedRun};
 use crate::db::{ArtifactId, Lookup};
-use crate::load::SiteCtxt;
+use crate::load::{MissingReason, SiteCtxt};
+use crate::server::status::BenchmarkError;
 
 // How many historical (finished) runs should be returned from the status API.
 const FINISHED_RUN_COUNT: u64 = 5;
 
 pub async fn handle_status_page(ctxt: Arc<SiteCtxt>) -> status::Response {
-    let missing = ctxt.missing_commits().await;
-
-    // FIXME: no current builds
+    let mut missing = vec![
+        (
+            Commit {
+                sha: "abcdef".to_string(),
+                date: Date::ymd_hms(2023, 1, 1, 0, 0, 0),
+                r#type: CommitType::Try,
+            },
+            MissingReason::Try {
+                pr: 123,
+                parent_sha: "foobar".to_string(),
+                include: None,
+                exclude: None,
+                runs: None,
+            },
+        ),
+        (
+            Commit {
+                sha: "abcdef2".to_string(),
+                date: Date::ymd_hms(2023, 2, 1, 0, 0, 0),
+                r#type: CommitType::Master,
+            },
+            MissingReason::Master {
+                pr: 421,
+                parent_sha: "foobar2".to_string(),
+                is_try_parent: true,
+            },
+        ),
+    ]; //ctxt.missing_commits().await;
+       // FIXME: no current builds
     let conn = ctxt.conn().await;
     let current = if let Some(artifact) = conn.in_progress_artifacts().await.pop() {
         let steps = conn
@@ -35,8 +63,76 @@ pub async fn handle_status_page(ctxt: Arc<SiteCtxt>) -> status::Response {
         None
     };
 
-    // FIXME: load at least one master commit with errors, if no master commit is in last N commits?
-    // FIXME: cache this whole thing, or write a specific SQL query for it
+    // let current: Option<CurrentState> = None;
+    let current = Some(status::CurrentState {
+        artifact: ArtifactId::Commit(Commit {
+            sha: "abcdef123".to_string(),
+            date: Date::ymd_hms(2023, 09, 11, 18, 00, 00),
+            r#type: CommitType::Try,
+        }),
+        progress: vec![
+            status::Step {
+                step: "a".to_string(),
+                is_done: true,
+                expected_duration: 120,
+                current_progress: 80,
+            },
+            status::Step {
+                step: "b".to_string(),
+                is_done: true,
+                expected_duration: 90,
+                current_progress: 120,
+            },
+            status::Step {
+                step: "c".to_string(),
+                is_done: false,
+                expected_duration: 140,
+                current_progress: 40,
+            },
+            status::Step {
+                step: "runtime:a".to_string(),
+                is_done: false,
+                expected_duration: 3,
+                current_progress: 0,
+            },
+            status::Step {
+                step: "runtime:b".to_string(),
+                is_done: false,
+                expected_duration: 4,
+                current_progress: 0,
+            },
+            status::Step {
+                step: "rustc".to_string(),
+                is_done: false,
+                expected_duration: 1000,
+                current_progress: 0,
+            },
+            status::Step {
+                step: "d".to_string(),
+                is_done: false,
+                expected_duration: 120,
+                current_progress: 0,
+            },
+        ],
+    });
+
+    if let Some(ref current) = current {
+        if let ArtifactId::Commit(ref commit) = current.artifact {
+            missing.push((
+                commit.clone(),
+                MissingReason::InProgress(Some(Box::new(MissingReason::Try {
+                    pr: 1240,
+                    parent_sha: "foo".to_string(),
+                    include: None,
+                    exclude: None,
+                    runs: None,
+                }))),
+            ));
+        }
+    }
+
+    /*// TODO: load at least one master commit with errors, if no master commit is in last N commits?
+    // TODO: cache this whole thing, or write a specific SQL query for it
     let mut finished_runs = Vec::new();
     let idx = ctxt.index.load();
 
@@ -64,7 +160,51 @@ pub async fn handle_status_page(ctxt: Arc<SiteCtxt>) -> status::Response {
             duration: collection.duration.as_secs(),
             finished_at: collection.end_time.timestamp() as u64,
         });
-    }
+    }*/
+    let mut finished_runs = vec![
+        FinishedRun {
+            artifact: ArtifactId::Commit(Commit {
+                sha: "abcdef".to_string(),
+                date: Date::ymd_hms(2023, 09, 05, 01, 02, 03),
+                r#type: CommitType::Try,
+            }),
+            pr: Some(128),
+            errors: vec![],
+            duration: 1234,
+            finished_at: 1694447945,
+        },
+        FinishedRun {
+            artifact: ArtifactId::Commit(Commit {
+                sha: "foobar".to_string(),
+                date: Date::ymd_hms(2023, 09, 04, 01, 02, 03),
+                r#type: CommitType::Master,
+            }),
+            pr: Some(256),
+            errors: vec![
+                BenchmarkError {
+                    name: "foo".to_string(),
+                    error: "ERROR".to_string(),
+                },
+                BenchmarkError {
+                    name: "bar".to_string(),
+                    error: "ERROR 2".to_string(),
+                },
+            ],
+            duration: 1462,
+            finished_at: 1694437945,
+        },
+        FinishedRun {
+            artifact: ArtifactId::Commit(Commit {
+                sha: "foobar2".to_string(),
+                date: Date::ymd_hms(2023, 09, 03, 01, 02, 03),
+                r#type: CommitType::Master,
+            }),
+            pr: Some(100),
+            errors: vec![],
+            duration: 1302,
+            finished_at: 1694427945,
+        },
+    ];
 
     status::Response {
         finished_runs,
```
</details>

### DIFF
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -69,26 +69,15 @@ function formatCommitAsHtml(commit: Commit, reason: MissingReason): string {
   return `${pullRequestUrlAsHtml(pr)} (${type}): ${url}`;
 }
 
-function formatReason(reason: any): string {
-  if (typeof reason == "string") {
-    return reason;
-  } else if (reason.InProgress) {
-    return `${formatReason(reason.InProgress)} - in progress`;
-  } else if (reason["Master"] !== undefined && reason.Master.pr) {
-    return `<a href="https://github.com/rust-lang/rust/pull/${
-      reason["Master"].pr
-    }">
-                    #${reason["Master"].pr}</a>${
-      reason.Master.is_try_parent ? " - Try commit parent" : ""
+function formatMissingReason(reason: MissingReason): string {
+  if (reason.hasOwnProperty("InProgress")) {
+    return `${formatMissingReason(reason["InProgress"])} - in progress`;
+  } else if (reason.hasOwnProperty("Master")) {
+    return `${pullRequestUrlAsHtml(reason["Master"].pr)}${
+      reason["Master"].is_try_parent ? " - Try commit parent" : ""
     }`;
-  } else if (reason["Master"] !== undefined && reason.Master.pr == 0) {
-    return "Master";
-  } else if (reason["Try"] !== undefined && reason.Try.pr) {
-    return `
-                Try for
-                <a href="https://github.com/rust-lang/rust/pull/${reason["Try"].pr}">
-                    #${reason["Try"].pr}
-                </a>`;
+  } else if (reason.hasOwnProperty("Try")) {
+    return `Try for ${pullRequestUrlAsHtml(reason["Try"].pr)}`;
   } else {
     // Should never happen, but a reasonable fallback
     return JSON.stringify(reason);
@@ -214,7 +203,7 @@ loadStatus(loading);
           <tr v-for="[commit, reason] in data.missing">
             <td>{{ new Date(commit.date).toLocaleString() }}</td>
             <td v-html="commitUrlAsHtml(commit.sha)"></td>
-            <td v-html="formatReason(reason)"></td>
+            <td v-html="formatMissingReason(reason)"></td>
           </tr>
         </tbody>
       </table>
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -2,13 +2,31 @@
 import {getJson} from "../../utils/requests";
 import {STATUS_DATA_URL} from "../../urls";
 import {withLoading} from "../../utils/loading";
-import {computed, ref, Ref} from "vue";
-import {Artifact, Commit, MissingReason, StatusResponse} from "./data";
+import {computed, ref, Ref, watch} from "vue";
+import {
+  Artifact,
+  BenchmarkError,
+  Commit,
+  FinishedRun,
+  MissingReason,
+  StatusResponse,
+} from "./data";
+import {
+  addSeconds,
+  differenceInSeconds,
+  format,
+  fromUnixTime,
+  sub,
+  subSeconds,
+} from "date-fns";
+import {CompileTestCase} from "../compare/compile/common";
+import {useExpandedStore} from "./expansion";
 
 async function loadStatus(loading: Ref<boolean>) {
   data.value = await withLoading(loading, () =>
     getJson<StatusResponse>(STATUS_DATA_URL)
   );
+  console.log(data.value);
 }
 
 function formatDuration(seconds: number): string {
@@ -19,11 +37,11 @@ function formatDuration(seconds: number): string {
 
   let s = "";
   if (hours > 0) {
-    s = `${hours}h${mins < 10 ? "0" + mins : mins}m${
+    s = `${hours}h ${mins < 10 ? "0" + mins : mins}m ${
       secs < 10 ? "0" + secs : secs
     }s`;
   } else {
-    s = `${mins < 10 ? " " + mins : mins}m${secs < 10 ? "0" + secs : secs}s`;
+    s = `${mins < 10 ? " " + mins : mins}m ${secs < 10 ? "0" + secs : secs}s`;
   }
   return s;
 }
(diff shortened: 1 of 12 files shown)


Mark-Simulacrum:
It's hard to tell based on the image - but it looks like the new page might be quite wide? Is it wider than the current page?

I often check in from a phone if there's reports of slowness or similar and the current page is pretty readable without sideways scrolling for me.

Not a blocker of course, scrolling isn't bad and I'm not sure optimizing for phones makes much sense anyway, but figured I'd raise it here.

Kobzol:
It uses a flex layout, so on mobile the timeline will be displayed below the current collection.That being said, the timeline table could be optimized for mobile more, now it requires a bit of horizontal scrolling.

Mark-Simulacrum:
Sounds great. I think a bit of scrolling is ok but we can reassess once things merge.

site/src/request_handlers/status_page.rs
```diff
@@ -2,14 +2,16 @@ use std::str;
...
 use crate::api::status;
+use crate::api::status::FinishedRun;
 use crate::db::{ArtifactId, Lookup};
 use crate::load::SiteCtxt;
 
-pub async fn handle_status_page(ctxt: Arc<SiteCtxt>) -> status::Response {
-    let idx = ctxt.index.load();
-    let last_commit = idx.commits().last().cloned();
+// How many historical (finished) runs should be returned from the status API.
+const FINISHED_RUN_COUNT: u64 = 5;
```
klensy, line 10: Can history be ~ twice bigger? 5 entries for 2 kind of runs kinda small, plus that way more easy to see any correlation in run time.

Kobzol: For future reference, it takes about ~40-50ms to load the status page on my PC/connection in the current version.

klensy:
Don't think that this can close https://github.com/rust-lang/rustc-perf/issues/1713, as source of difference not found, but step for simpler debugging.

Maybe rolling window of bench logs/artifacts can be kept, so if anything funny happens again, we can check logs?

Kobzol:
Good point. We have logs, but I'm not sure if the current ones are very helpful. FWIW, I think that the long duration is caused by DB, it has been behaving quite erratically recently. We could add some timings to the DB operations to check if that is indeed the issue.

site/frontend/src/pages/status/page.vue
```diff
@@ -2,13 +2,31 @@
...
+  subSeconds,
+} from "date-fns";
+import {CompileTestCase} from "../compare/compile/common";
+import {useExpandedStore} from "./expansion";
 
 async function loadStatus(loading: Ref<boolean>) {
   data.value = await withLoading(loading, () =>
     getJson<StatusResponse>(STATUS_DATA_URL)
   );
+  console.log(data.value);
```
lqd, line 29:
```suggestion  
```

site/frontend/src/pages/status/page.vue
```diff
@@ -19,84 +37,262 @@ function formatDuration(seconds: number): string {
...
+        expected_end: addSeconds(new Date(), timeLeft.value),
+        expected_total,
+        progress,
+      };
+    }
+  }
+  return null;
+});
+const lastFinishedRun: Ref<FinishedRun | null> = computed(() => {
+  const finished = data?.value.finished_runs;
```
lqd, line 157: it seems `value` or `finished_runs` can be null/undefined here

site/frontend/src/pages/status/page.vue
```diff
@@ -19,84 +37,262 @@ function formatDuration(seconds: number): string {
...
+      end_time: current.expected_end,
+      end_estimated: true,
+      errors: [],
+      duration: null,
+      warning: false,
+      current: true,
+    });
+  }
+
+  for (const run of data.value.finished_runs) {
```
lqd, line 228: as well as here

lqd (commented):
The frontend triggers errors on startup for me so I can't really test, unfortunately.

site/frontend/src/pages/status/page.vue
```diff
@@ -119,22 +257,42 @@ loadStatus(loading);
...
+              <td>
+                <progress
+                  max="100"
+                  :value="currentRun.progress * 100"
+                ></progress>
+              </td>
+              <td>
+                {{ format(currentRun.expected_end, "HH:mm") }}
+              </td>
+              <td>{{ timeLeft <= 0 ? "?" : formatDuration(timeLeft) }}</td>
```
lqd, line 291: maybe a dash instead of "?", or nothing.

lqd (commented):
I managed to look at the frontend with the patch, and we may need to iterate a bit after this lands in case the production data has surprising effects (e.g. actual lengths impacting layout and scrolling).

The PR number and sha columns should be aligned and not centered. (The date in the Run end column could also maybe be turned into a fixed width, or the alignment there will be randomly shifting as well).

I didn't look closely at the queries, or API changes. It looked fine at a quick glance. Maybe Mark or Ryan can take a look?

I didn't notice it at first but it seems the timeline reverses the queue order compared to a regular ordering by position. If that's the case, I'm not sure we gain much by mixing the future collections with the past collections in a single list, with the current collection duplicated between the two. Compared to, say, having 2 tables: the queue in ASC order, and history of recent collections in DESC order. That feels clearer to me, but I don't mind If others prefer it.

site/frontend/src/pages/status/page.vue
Kobzol: Good point, this should have be `data.value?` instead.

Kobzol:
> The PR number and sha columns should be aligned and not centered. (The date in the Run end column could also maybe be turned into a fixed width, or the alignment there will be randomly shifting as well).

Right aligned PR and SHA (they will be almost always the same width on rust-lang/rust, but it's still better to align).

The run end data was previously formatted with the local "locale", which I don't like that much, but I didn't want to change that. I'm not sure how to make it fixed width when the format is not fixed.

> I didn't notice it at first but it seems the timeline reverses the queue order compared to a regular ordering by position. If that's the case, I'm not sure we gain much by mixing the future collections with the past collections in a single list, with the current collection duplicated between the two. Compared to, say, having 2 tables: the queue in ASC order, and history of recent collections in DESC order. That feels clearer to me, but I don't mind If others prefer it.

Yeah it's currently basically ordered by start/end time, from newest (which can be in the future) to the oldest. The interpretation is that the next PR will "fall down" into the marked down after the current collection ends. Having two tables could also be reasonable. I just wanted to avoid the previous situation where a currently running PR would basically be in a sort of a random position within the list/table.

Kobzol:
CC @Mark-Simulacrum if you could please take a look at the DB changes (and possibly comment on the new design).

Mark-Simulacrum (approved):

Mark-Simulacrum:
Broadly happy for it to be a unified table. I think we can always edit if we don't like it, so I'm inclined to move ahead when ready.

PR #1724, opened 2023-09-20
Integrate Clippy into the project

blyxyas:
This PR integrates Clippy into the project.
A user can run the `clippy` profile with a local Clippy build, with the `--clippy` flag targetting to the Clippy executable:

```sh
$ cargo run --bin collector  -- bench_local +nightly --profiles Clippy --clippy ../clippy/target/release/cargo-clippy
```

---

But I'm having some problems with the bot integration, so some feedback and tips in those files would be very appreciated.

* Would the bot run this on every commit, where is it configured? (we don't want to run it every commit)
Thanks for the feedback :heart:

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -375,6 +375,10 @@ struct CompileTimeOptions {
     /// The path to the local rustdoc to measure
     #[arg(long)]
     rustdoc: Option<PathBuf>,
+
+    /// The path to the local clippy to measure
+    #[arg(long)]
+    clippy: Option<PathBuf>
 }
 
 #[derive(Debug, clap::Args)]
@@ -663,6 +667,7 @@ fn main_result() -> anyhow::Result<i32> {
                     None,
                     None,
                     None,
+                    None,
                     id,
                     target_triple.clone(),
                 )?;
@@ -721,6 +726,7 @@ fn main_result() -> anyhow::Result<i32> {
                     None,
                     None,
                     None,
+                    None,
                     id,
                     target_triple.clone(),
                 )?;
@@ -757,6 +763,7 @@ fn main_result() -> anyhow::Result<i32> {
                 &profiles,
                 &local.rustc,
                 opts.rustdoc.as_deref(),
+                opts.clippy.as_deref(),
                 local.cargo.as_deref(),
                 local.id.as_deref(),
                 "",
@@ -949,6 +956,7 @@ fn main_result() -> anyhow::Result<i32> {
                         profiles,
                         rustc,
                         opts.rustdoc.as_deref(),
+                        opts.clippy.as_deref(),
                         local.cargo.as_deref(),
                         local.id.as_deref(),
                         suffix,
@@ -1062,6 +1070,7 @@ fn get_local_toolchain_for_runtime_benchmarks(
     get_local_toolchain(
         &[Profile::Opt],
         &local.rustc,
+        None,
         None,
         local.cargo.as_deref(),
         local.id.as_deref(),
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -1197,9 +1197,18 @@ fn bench_published_artifact(
     let artifact_id = ArtifactId::Tag(toolchain.id.clone());
 
     let profiles = if collector::version_supports_doc(&toolchain.id) {
-        Profile::all()
+        vec![
+            Profile::Check,
+            Profile::Debug,
+            Profile::Doc,
+            Profile::Opt
+        ]
     } else {
-        Profile::all_non_doc()
+        vec![
+            Profile::Check,
+            Profile::Debug,
+            Profile::Opt,
+        ]
     };
     let scenarios = if collector::version_supports_incremental(&toolchain.id) {
         Scenario::all()
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -38,7 +38,7 @@ use collector::runtime::{
 };
 use collector::runtime::{profile_runtime, RuntimeCompilationOpts};
 use collector::toolchain::{
-    create_toolchain_from_published_version, get_local_toolchain, Sysroot, Toolchain,
+    create_toolchain_from_published_version, get_local_toolchain, Sysroot, Toolchain, ToolchainConfig,
 };
 use collector::utils::cachegrind::cachegrind_diff;
 use collector::utils::{is_installed, wait_for_future};
@@ -664,10 +664,7 @@ fn main_result() -> anyhow::Result<i32> {
                 let toolchain = get_local_toolchain(
                     &[Profile::Opt],
                     rustc,
-                    None,
-                    None,
-                    None,
-                    None,
+                    ToolchainConfig::default(),
                     id,
                     target_triple.clone(),
                 )?;
@@ -723,10 +720,7 @@ fn main_result() -> anyhow::Result<i32> {
                 let toolchain = get_local_toolchain(
                     &[Profile::Opt],
                     rustc,
-                    None,
-                    None,
-                    None,
-                    None,
+                    ToolchainConfig::default(),
                     id,
                     target_triple.clone(),
                 )?;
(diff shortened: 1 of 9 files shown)


collector/compile-benchmarks/cargo-0.60.0/src/cargo/core/profiles.rs
```diff
@@ -95,6 +95,14 @@ impl Profiles {
                 .dir_names
                 .insert(InternedString::new("doc"), InternedString::new("debug"));
 
+            profile_makers.by_name.insert(
```
Kobzol, line 98: Why did you need to change the source of the cargo benchmark? It shouldn't be required to change a benchmark to run `clippy` on it.

collector/src/bin/collector.rs
```diff
@@ -663,6 +667,7 @@ fn main_result() -> anyhow::Result<i32> {
                     None,
```
Kobzol, line 667: I think that we have gotten past the point where N `None`s in a row is an acceptable API :laughing: I know that it's mostly preexisting, but would you be willing to turn this e.g. into a builder pattern? Something like
```rust
get_local_toolchain(..., ToolchainConfig::default()) // everything `None` by default
get_local_toolchain(..., ToolchainConfig::default().rustc(...).clippy(...)) // some things filled
```
it could contain all parameters for the `get_local_toolchain` function, where it makes sense.

collector/src/compile/benchmark/profile.rs
```diff
@@ -12,13 +12,15 @@ pub enum Profile {
     Debug,
     Doc,
     Opt,
+    Clippy,
 }
 
 impl Profile {
     pub fn all() -> Vec<Self> {
```
Kobzol, line 19: This will need to be modified in `BenchNext` and `bench_published_artifact`. These two functions should not benchmark Clippy. Probably the functions should directly decide the specific profiles to benchmark, rather than delegating to these two functions (`all` and `all_non_doc`).

collector/src/toolchain.rs
```diff
@@ -400,6 +405,26 @@ pub fn get_local_toolchain(
             None
         };
 
+        let clippy =
+        if let Some(clippy) = &clippy {
+            Some(clippy.canonicalize().with_context(|| {
+                format!("failed to canonicalize rustdoc executable {:?}", clippy)
```
Kobzol, line 411:
```suggestion
                format!("failed to canonicalize clippy executable {:?}", clippy)
```

collector/src/toolchain.rs
```diff
@@ -400,6 +405,26 @@ pub fn get_local_toolchain(
...
+            })?)
+        } else if profiles.contains(&Profile::Clippy) {
+            // We need a `clippy`. Look for one next to `rustc`.
+            if let Ok(clippy) = rustc.with_file_name("clippy").canonicalize() {
```
Kobzol, line 415: I think that this should be `cargo-clippy` instead of `clippy`. It is called `cargo-clippy` in my local toolchains installed by `rustup`.

collector/src/toolchain.rs
```diff
@@ -400,6 +405,26 @@ pub fn get_local_toolchain(
...
+                debug!("found clippy: {:?}", &clippy);
+                Some(clippy)
+            } else {
+                anyhow::bail!(
+                    "'Clippy' build specified but '--clippy' not specified and no 'clippy' found \
```
Kobzol, line 420:
```suggestion
                    "'Clippy' build specified but '--clippy' not specified and no 'cargo-clippy' found \
```

site/src/request_handlers/github.rs
```diff
@@ -123,8 +123,15 @@ async fn handle_rust_timer(
         let exclude = captures.get(2).map(|v| v.as_str());
         let runs = captures.get(3).and_then(|v| v.as_str().parse::<i32>().ok());
         {
+            let profile;
```
Kobzol, line 126: I'm not sure about the integration of clippy into the main dashboard. I'd rather leave this part of the PR out for now, and just focus on data gathering.

database/src/pool/sqlite.rs
```diff
@@ -957,13 +957,14 @@ impl Connection for SqliteConnection {
         include: Option<&str>,
         exclude: Option<&str>,
         runs: Option<i32>,
+        profile: Option<&str>
     ) {
         self.raw_ref()
             .prepare_cached(
-                "insert into pull_request_build (pr, complete, requested, include, exclude, runs) VALUES (?, 0, strftime('%s','now'), ?, ?, ?)",
+                "insert into pull_request_build (pr, complete, requested, include, exclude, runs, profile) VALUES (?, 0, strftime('%s','now'), ?, ?, ?, ?)",
```
Kobzol, line 964: Why did you change SQL queries? I don't think that we need to change anything SQL related in this PR. Also, if we did, we would have to write a migration.

Kobzol (commented):
Hi, thanks for the PR! I left some comments. I wonder what was your motivation for this PR? Do you work on `Clippy` and want to measure its performance on crates?

blyxyas:
> I wonder what was your motivation for this PR? Do you work on Clippy and want to measure its performance on crates?
Yep, we wanted a tool for measuring Clippy's performance on crates.

site/src/request_handlers/github.rs
blyxyas: Would the data still be viewable outside the dashboard somewhere?

Kobzol: It would be visible only in the local DB. I would like to first merge support for benchmarking clippy. We can discuss its inclusion in the server after that :)

blyxyas:
This new commit should be just like the first one, but discarding anything server-related =^w^=

collector/src/toolchain.rs
```diff
@@ -123,6 +123,7 @@ impl SysrootDownload {
         let components = ToolchainComponents::from_binaries_and_libdir(
             sysroot_bin("rustc")?,
             Some(sysroot_bin("rustdoc")?),
+            Some(sysroot_bin("clippy")?),
```
Kobzol, line 126: As is apparent on CI, `clippy` might not always be applicable. So we should only pass it here if it actually exists. Also, it should also probably be `cargo-clippy` here.

collector/src/toolchain.rs
```diff
@@ -288,6 +292,45 @@ impl ToolchainComponents {
...
+#[derive(Clone, Copy)]
+pub struct ToolchainConfig<'a> {
+    rustdoc: Option<&'a Path>,
+    clippy: Option<&'a Path>,
+    cargo: Option<&'a Path>,
+    id: Option<&'a str>,
+}
+
+impl<'a> ToolchainConfig<'a> {
+    pub fn default() -> Self {
```
Kobzol, line 304: Somewhat ironically, clippy tells you to derive Default for this struct :laughing: https://github.com/rust-lang/rustc-perf/actions/runs/6273536371/job/17037178205?pr=1724

Kobzol (commented):
Thanks for the changes. Some problems that cause CI to fail still have to be resolved, as per the other comments.

Also, as per https://github.com/rust-lang/rustc-perf/pull/1724#discussion_r1332925816, we shouldn't include `Clippy` in `Profile::all()` for now. Let's just remove it from there, that should fix collection on the server. Later we might add it back once we're sure that clippy benchmarks can be run on the perf server.

collector/src/toolchain.rs
blyxyas: Done :cat:!

collector/src/toolchain.rs
blyxyas: Probably done, I'll check CI to see if the fix I did is correct :cat:

collector/src/toolchain.rs
```diff
@@ -123,7 +123,10 @@ impl SysrootDownload {
         let components = ToolchainComponents::from_binaries_and_libdir(
             sysroot_bin("rustc")?,
             Some(sysroot_bin("rustdoc")?),
-            Some(sysroot_bin("clippy")?),
+            match sysroot_bin("cargo-clippy") {
```
Kobzol, line 126: There is a helper `or` method for this (https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) in the stdlib, btw :) But you can keep it as it is here.

Kobzol (approved):
Great, thanks for the changes! :)

collector/src/toolchain.rs
blyxyas: Do I... *not know `std`* :fearful:. Haha I'll change it right now.

Issue #1727, opened 2023-09-27
compare page tab availability should not depend on currrently selected metric


klensy:
For example, https://perf.rust-lang.org/compare.html?start=6710e3312d54f03f6462481199b9e72367a2568b&end=5899a80ae60ec0959dcd4a7eca6eb02ce2866632&stat=instructions%3Au&tab=compile

1. Select compile-time tab and instructions metric: all tabs available.
2. Select in current tab (compile-time) max-rss metric: page refreshes and runtime tab disappears.

This behavior slightly confusing. Tabs should be always available, if some metric not available, it should reset to some default one.

Kobzol:
Right, hiding the tab was a temporary measure until we gather some runtime data. Now it should be shown, but without any entries, if there is no data for that metric. Thanks! I'll try to fix it soon.

PR #1728, opened 2023-09-27
Do not hide runtime tab when there are no results

Kobzol:
This was just a temporary precaution until we bootstrap enough data for runtime benchmarks. Now it shouldn't be needed. This PR removes the special case.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1727

### DIFF
--- a/site/frontend/src/pages/compare/page.vue
+++ b/site/frontend/src/pages/compare/page.vue
@@ -81,17 +81,15 @@ async function loadCompareData(
       )
     )
   );
-  if (response.runtime_comparisons.length > 0) {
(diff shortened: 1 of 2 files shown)


lqd (approved):

klensy:
Some docker troubles, https://github.com/rust-lang/rustc-perf/actions/runs/6343476830/job/17231414819 (looks like in main rustc repo too).

Kobzol:
Yeah, Docker's servers are down, apparently.

klensy:
Can dummy commit be pushed, so this pr will be deployed?

PR #1729, opened 2023-09-28
Add absolute chart to benchmark details

lqd:
This PR adds an absolute chart to the benchmark detail panel, to complement the relative chart:
- the relative chart easily shows cyclical noise, and the common magnitude seen on a given benchmark
- the absolute chart easily shows whether a change was noise and has returned to steady state, or is real and has stayed the same since the commit.

Screenshot

image

### DIFF
--- a/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
+++ b/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
@@ -96,7 +96,18 @@ function drawCurrentDate(opts: GraphRenderOpts, date: Date) {
   };
 }
 
-async function renderGraph() {
+// Render both relative and absolute graphs
+async function renderGraphs() {
+  // We want to be able to see noise "blips" vs. a previous artifact.
+  // The "percent relative from previous commit" graph should be the best to
+  // see these kinds of changes.
+  renderGraph("percentrelative" as GraphKind, relativeChartElement);
+  // We also want to see whether a change maintained its value or whether it was noise and has since
+  // returned to steady state. Here, an absolute graph ("raw") is best.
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -359,16 +381,20 @@ onMounted(() => renderGraph());
     flex-grow: 1;
   }
 }
+.columns + .columns {
```
Kobzol, line 384: I have been doing webdev for 12+ years and this is the first time I'm seeing this :eyes: Bloody CSS. I would probably just add a new class, like `.graphs`, and add the style to that, to make it more obvious/readable.

Kobzol (commented):
This is great, I like the horizontal layout and the fact that the graphs are both visible at once.

For me, they render below each other though:
image

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
lqd: Done

lqd:
I also reduced the graph width a bit so that both should fit in a single row consistently, which should fix the issue in your screenshot.

lqd:
Actually incredible. I pushed a `nowrap`, and hope it works for you because 3rd time's the charm.

Kobzol (approved):
Now it works, thanks! Just remove https://github.com/rust-lang/rustc-perf/compare/b873b613af29f44ab5be13466b82b765af690c8e..627d02e9a5f7d000f251f9f55f8c1dc54b3806cb#diff-813a534b1cbe87a3c3d6370efa9e279cec7c8083308f493d6dd95e12d9feab81R1, and otherwise LGTM.

PR #1736, opened 2023-10-22
fix(collector): allow unused imports in serde-1.0.136

bvanjoi:
cc @Kobzol


Mark-Simulacrum (approved):

Kobzol:
It's possible that there are also some other benchmarks that fail, as we have only tried the "training crates" on CI, but since the build didn't succeed, the full benchmark run didn't finish.

If you have a build of the compiler with the corresponding change from the PR, could you please try to run the whole compilation benchmark suite locally, to see if it works?

```bash
$ cargo build
$ cargo run --bin collector bench_local <path-to-rustc> --self-profile
```

We should make sure that it works, because after we update `rustc-perf`, we also need to send another PR to `rust-lang/rust`, to update to the modified `rustc-perf` suite for PGO/BOLT training.

bvanjoi:
> could you please try to run the whole compilation benchmark suite locally, to see if it works


I sequentially ran ﻿`cargo +stage1 build` for each project within ﻿`compile-benchmarks` and `﻿runtime-benchmarks`, and found that the ﻿`unused_imports` error occurred exclusively in ﻿`serde`.

- (`stage1` is build on https://github.com/rust-lang/rust/pull/116033)
- (The reason I refrained from using ﻿`cargo run --bin collector xxx` is due to my local environment being macos, which doesn't support ﻿`pref`.)

But there has another errors:
- build `compile-benchmarks/stm32f4-0.14.0` failed with:

```
error[E0460]: found possibly newer version of crate `std` which `rustc_version` depends on
 --> bare-metal-0.2.5/build.rs:1:1
  |
1 | extern crate rustc_version;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: perhaps that crate needs to be recompiled?
  = note: the following crate versions were found:
// ....
Some errors have detailed explanations: E0432, E0460, E0463.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `syn` (lib) due to 5 previous errors
```

- build `compile-benchmarks/style-servo` failed with:

```
error[E0460]: found possibly newer version of crate `std` which `app_units` depends on
  --> components/style_traits/lib.rs:16:1
   |
16 | extern crate app_units;

// .....
For more information about this error, try `rustc --explain E0503`.
error: could not compile `getopts` (lib) due to previous error
```

- build `compile-benchmarks/tokio-webpush-simple` failed with:

```
error[E0460]: found possibly newer version of crate `std` which `tokio_core` depends on
// .....
For more information about this error, try `rustc --explain E0460`.
error: could not compile `serde_json` (lib) due to 15 previous errors
error: could not compile `tokio-proto` (lib) due to 100 previous errors
```

- build `compile-benchmarks/webrender-2022` failed with:

```
// ....
error: cannot determine resolution for the derive macro `PeekPoke
// ....
Some errors have detailed explanations: E0412, E0432, E0433, E0460, E0463.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `webrender_api` (lib) due to 449 previous errors
```

- build `runtime-benchmarks/svg` failed with:

```
// .....
error[E0282]: type annotations needed
// .....
Some errors have detailed explanations: E0282, E0460, E0463, E0599.
For more information about an error, try `rustc --explain E0282`.
error: could not compile `resvg` (lib) due to 642 previous errors
```

Kobzol:
Yes that appears to be flaky, possibly some OS X thing. Thanks! I will create a PR on rust-lang/rust to sync rustc-perf.

PR #1742, opened 2023-11-09
Add basic support for selecting a codegen backend for compile benchmarks

Kobzol:
This PR introduces a new parameter for compilation benchmarks - a codegen backend. It adds support for it to the collector and the database. It does not yet modify the web UI, and it also doesn't add other backends than the default one (LLVM).

The database changes will require the most review scrutiny. I tested the DB migration for both SQLite and Postgre locally.

### DIFF
--- a/collector/src/compile/benchmark/codegen_backend.rs
+++ b/collector/src/compile/benchmark/codegen_backend.rs
@@ -1,0 +1,11 @@
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum, serde::Deserialize)]
+#[value(rename_all = "PascalCase")]
+pub enum CodegenBackend {
+    Llvm,
+}
+
+impl CodegenBackend {
+    pub fn all() -> Vec<CodegenBackend> {
+        vec![CodegenBackend::Llvm]
+    }
+}
--- a/database/schema.md
+++ b/database/schema.md
@@ -26,15 +26,16 @@ Here is the diagram for compile-time benchmarks:
 │ └────────────┘  └───────────────┘ │└────────────┘ │
 │                                   │               │
 │                                   │               │
-│ ┌───────────────┐  ┌──────────┐   |               │
+│ ┌───────────────┐  ┌──────────┐   │               │
 │ │ pstat_series  │  │ pstat    │   │               │
 │ ├───────────────┤  ├──────────┤   │               │ 
 │ │ id *          │◄┐│ id *     │   │               │
 └─┤ crate         │ └┤ series   │   │               │ 
   │ profile       │  │ aid      ├───┼───────────────┘
-  │ cache         │  │ cid      ├───┘
-  │ statistic     │  │ value    │
-  └───────────────┘  └──────────┘
+  │ scenario      │  │ cid      │   │
+  │ backend       │  │ value    ├───┘
+  │ metric        │  └──────────┘
+  └───────────────┘
 ```
 
 For runtime benchmarks the schema very similar, but there are different table names:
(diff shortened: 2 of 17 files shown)


collector/src/compile/benchmark/codegen_backend.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum, serde::Deserialize)]
+#[value(rename_all = "PascalCase")]
+pub enum CodegenBackend {
+    LLVM,
```
nnethercote, line 4: Standard Rust style would be to call this `Llvm`. Then you might not need the PascalCase?

database/schema.md
```diff
@@ -26,15 +26,16 @@ Here is the diagram for compile-time benchmarks:
...
-│ ┌───────────────┐  ┌──────────┐   |               │
+│ ┌───────────────┐  ┌──────────┐   │               │
 │ │ pstat_series  │  │ pstat    │   │               │
 │ ├───────────────┤  ├──────────┤   │               │ 
 │ │ id *          │◄┐│ id *     │   │               │
 └─┤ crate         │ └┤ series   │   │               │ 
   │ profile       │  │ aid      ├───┼───────────────┘
-  │ cache         │  │ cid      ├───┘
-  │ statistic     │  │ value    │
-  └───────────────┘  └──────────┘
```
nnethercote, line 37: This is changing more than just adding a backend. Were `cache` and `statistic` out of date, or copy/paste errors?

database/src/lib.rs
```diff
@@ -360,6 +360,39 @@ impl PartialOrd for Scenario {
     }
 }
 
+/// The codegen backend used for compilation.
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
+)]
+pub enum CodegenBackend {
+    /// The default LLVM backend
+    LLVM,
```
nnethercote, line 369: Again, `Llvm` is the usual style.

database/src/pool/postgres.rs
```diff
@@ -254,6 +254,13 @@ static MIGRATIONS: &[&str] = &[
         UNIQUE(aid, component)
     );
     "#,
+    r#"
+    alter table pstat_series rename column cache to scenario;
+    alter table pstat_series rename column statistic to metric;
+    alter table pstat_series add backend text not null default 'llvm';
+    alter table pstat_series drop constraint pstat_series_crate_profile_cache_statistic_key;
+    alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, metric);
+    "#,
```
nnethercote, line 263: Hmm, is this commit renaming some fields as well? Would be good if the commit log mentioned this.

nnethercote (approved):
Basically seems ok, though I didn't really look at the DB changes because I don't know anything about that.

database/src/pool/postgres.rs
Kobzol: Yes. It's specified in a comment among the changes, but now I see that only in the SQLite DB, and you're right that it should also be in the commit message. Thanks.

Kobzol:
Thanks for taking a look, I have resolved the comments.

Kobzol:
Turns our that `PascalCase` is still needed, otherwise serde will just use the lowercase version by default.

database/src/pool/postgres.rs
```diff
@@ -254,6 +254,15 @@ static MIGRATIONS: &[&str] = &[
...
     );
     "#,
+    // Add codegen backend column and add it to the unique constraint.
+    // Also rename cache to scenario and statistic to metric, while we're at it.
+    r#"
+    alter table pstat_series rename column cache to scenario;
+    alter table pstat_series rename column statistic to metric;
+    alter table pstat_series add backend text not null default 'llvm';
+    alter table pstat_series drop constraint pstat_series_crate_profile_cache_statistic_key;
+    alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, metric);
```
Mark-Simulacrum, line 264: Have we tested this migration on a local database or so? I'm a bit worried that there's (AFAICT) no index that would help with this unique constraint, right? Or if there is such an index, then maybe we need to drop the *previous* index so we're not ending up with 2 separate indices.

Creating the new index may also be pretty slow, but I suppose that's a one-time cost and shouldn't be too bad.

site/src/comparison.rs
```diff
@@ -155,6 +155,7 @@ pub async fn handle_compare(
             comparison: comparison.comparison.into(),
         })
         .collect();
+    println!("{:?}", compile_comparisons);
```
Mark-Simulacrum, line 158: nit: stray println!?

database/src/pool/postgres.rs
Kobzol: Hmm, I kind of assumed that dropping the UNIQUE constraint would also drop the corresponding index. When I tried it on a local Postgres database, the old index was removed. I tried it again now. I converted the live site data from SQLite to Postgres and then ran the migration.

The migration took 63 milliseconds, for ~13 thousand `pstat_series` entries. I guess that there's not *that much* entries in this table, so it shouldn't take too long, hopefully.

Before:
image

After:
image

Mark-Simulacrum: Hm, ok. I guess it's auto-created? Should be OK then.

Mark-Simulacrum (approved):

Kobzol:
Here goes nothing.

PR #1744, opened 2023-11-11
Add compile benchmark detail endpoint

Kobzol:
This PR adds a new endpoint for returning the data of a compile benchmark detail (the thing that is shown when you click on a specific row on the compile benchmark compare page). Before, two separate API calls had to be made.

I want to add more data to this detail endpoint in the future, so I created a new endpoint to put everything into a single place.


site/src/api.rs
```diff
@@ -110,6 +110,65 @@ pub mod graphs {
...
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
+                v.split(',')
+                    .map(|v| {
+                        serde_json::from_str::<T>(&format!("\"{v}\""))
+                            .map_err(|error| E::custom(format!("{error:?}")))
+                    })
+                    .collect::<Result<Vec<T>, _>>()
```
Mark-Simulacrum, line 158: This looks pretty weird. I think the serde_json::from_str should be replaceable with `T::deserialize(serde::de::value::StrDeserializer::new(v))` which should be more efficient and cause less confusion.

Mark-Simulacrum (approved):
Seems broadly OK to me.

site/src/api.rs
Kobzol: Thanks, this is much better.

PR #1749, opened 2023-11-17
Add a chart showing individual compilation parts to benchmark detail

Kobzol:
So that we have a better idea of how much time is spent in frontend/backend/linking, and also how are these fractions changed by individual PRs. The data is loaded from self-profile files and it's calculated based on time spent in individual self-profile queries. There is currently a super rough [heuristic](https://github.com/rust-lang/rustc-perf/commit/7b58abcd8d26e823077505aadca2bc329d2c5871#diff-4340dc3a4f2c67c4492a7280d414af9a4ac74cd96b95a289d4c295ff4def2d60R102) that tries to estimate into which part does a query belong.

chart

Depends on https://github.com/rust-lang/rustc-perf/pull/1747 and https://github.com/rust-lang/rustc-perf/pull/1744.


Kobzol:
I separated the endpoints for the graphs and for the compilation sections, so that the (slow) section loading doesn't slow down the loading of the graphs.

Mark-Simulacrum (approved):

PR #1756, opened 2023-11-26
Add option of `-Z threads`

blyxyas:
This PR adds `parallel-compiler = true` support for `rustc-perf`. It work by adding a `ZTHREADS` env. var. and passing it to RUSTFLAGS if it is set.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -155,6 +155,8 @@ The following options alter the behaviour of the `bench_local` subcommand.
 `RUST_LOG=debug` can be specified to enable verbose logging, which is useful
 for debugging `collector` itself.
 
+`ZTHREADS=<number of threads>` can be used for builds with `parallel-compiler = true` in the build's `config.toml`. This will set up that number as a flag when compiling the benchmarks.
+
 ### How to compare different versions on your own machine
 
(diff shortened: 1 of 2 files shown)


Kobzol:
Hi :) We will definitely want to add support for this in rustc-perf, but it will have to be handled in a more complex manner. We need to store the information about the number of used threads into the database, the website, the CLI interface, and also double check how we're handling threads in the collector, because we use some custom configuration for jobserver tokens, and we need to check how it works with multiple threads.

Enabling threads using env. variables is a hack that I'd prefer not to add to the collector.

blyxyas:
I totally understand. Honestly, I'm not really sure what I was thinking when I opened this PR. I'm not looking for it to get merged, it was just a "if someone wants this resource, here it is".

I have some deadlines to meet, so I won't be able to update this PR along the way, let's keep this just closed as an archival for someone who wants to have this functionality early.

Sorry for wasting your time :(

Kobzol:
> Sorry for wasting your time :(

No problem, no time was wasted :)

I think that having this PR archived is a good way for someone who also wants to quickly try out more threads with rustc-perf to find out how to do that. So thanks for that!

PR #1770, opened 2023-12-14
Add CI workflow for benchmarking Cranelift backend

Kobzol:
This PR adds a CI workflow that checks whether we can run benchmarks using the Cranelift backend. It also adds support for downloading the Cranelift backend component in `install_nexŧ`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,6 @@ jobs:
       - name: Run unit tests
         run: cargo test --all
 
-
   test_benchmarks:
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum (approved):
Approving, but let's try to deploy this in the 2nd half of the day (ET) or on the weekend so I can fix the collector if needed. The download code isn't really automated today in terms of recovery from faults.

Kobzol:
I don't think that this should affect the server `collector` at all (?). It doesn't use `install_next`.

Mark-Simulacrum:
Ah, then feel free to merge whenever. I think I misremembered - we use bench_next looks like: https://github.com/rust-lang/rustc-perf/blob/master/collector/collect.sh#L25

PR #1772, opened 2023-12-17
Add a `collector` command to examine and compare binary size statistics

Kobzol:
This PR adds a new `collector` command called `binary_stats`, which can be used to analyze the sizes of ELF binaries and libraries produced by compile benchmarks. The command could be easily modified in the future to allow analyzing arbitrary binary artifacts specified by a filesystem path.

The command uses the [`elf`](https://crates.io/crates/elf) crate to print the sizes of sections and symbols in the ELF artifact. A few heuristics are used to normalize and group section/symbol names for less output clutter.

The command can also be used to compare the binary size of a benchmark between two different versions of `rustc` (this will probably be the most common use-case) and to compare the binary size between two different codegen backends (currently Llvm/Cranelift).

Examples:
- Examine size
    image
- Diff between two compilers/codegen backends:
    image


Mark-Simulacrum:
Hm, can you say more about how you chose the elf crate? I personally usually use object (https://docs.rs/object/latest/object/) -- it's also used in std, so to some extent is more maintained. It would probably make it easier to run this tool on non-ELF platforms (e.g., macOS or Windows) as well.

Kobzol:
Googled "Rust ELF" and selected the first thing that looked maintained, documented and allowed me to solve what I needed :)

Happy to switch to `object`, if it can handle binaries and also `.rlib`s and `.so`s.

Mark-Simulacrum:
.sos should definitely work (they're also ELF). .rlib may require https://docs.rs/ar/latest/ar/ or so - they're unix archives under the hood containing (usually) a pile of .o files (ELF or Mach-O or etc depending on platform) + the metadata.

Kobzol:
Yeah, this PR already uses `ar` to split rlibs into objects and then parse them as ELF :) I'll take a look into using `object`. It shouldn't change the other parts of the PR.

Mark-Simulacrum (approved):
Broadly though this looks great to me.

nnethercote (approved):
This is great, thanks! I don't have any opinions about which ELF crate to use.

Kobzol:
I pushed a change that just switches from `elf` + `ar` to `object`, which should hopefully also support Windows/macOS binaries.

PR #1773, opened 2023-12-17
Add binary size local command shortcut to benchmark detail

Kobzol:
This PR adds a command shortcut for analyzing binary size locally using the new [`binary_stats`](https://github.com/rust-lang/rustc-perf/pull/1772) command to the benchmark detail panel.

image

### DIFF
--- a/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
+++ b/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
@@ -12,7 +12,6 @@ import {daysBetweenDates, getFutureDate, getPastDate} from "./utils";
 import {GraphRenderOpts, renderPlots} from "../../../../graph/render";
 import {GraphData, GraphKind, GraphsSelector} from "../../../../graph/data";
 import uPlot from "uplot";
-import CachegrindCmd from "../../../../components/cachegrind-cmd.vue";
 import {
   COMPILE_DETAIL_GRAPHS_RESOLVER,
   COMPILE_DETAIL_SECTIONS_RESOLVER,
@@ -23,6 +22,7 @@ import {
 } from "./detail-resolver";
 import CompileSectionsChart from "./sections-chart.vue";
 import PerfettoLink from "../../../../components/perfetto-link.vue";
+import ProfileShortcut from "./shortcuts/profile-shortcut.vue";
 
 const props = defineProps<{
   testCase: CompileTestCase;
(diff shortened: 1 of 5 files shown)


site/frontend/src/pages/compare/compile/table/shortcuts/cachegrind-cmd.vue
```diff
@@ -1,5 +1,8 @@
 <script setup lang="ts">
-import {CompileTestCase, Profile} from "../pages/compare/compile/common";
+/**
+ * This component displays a rustc-perf command for profiling a compile benchmark with Cachegrind.
+ * */
```
lqd, line 4: tiny nit: newline here, I thought this was describing the import 😓

site/frontend/src/pages/compare/compile/table/shortcuts/profile-shortcut.vue
```diff
@@ -0,0 +1,82 @@
+<script setup lang="ts">
+import {computed, ref, Ref} from "vue";
```
lqd, line 2: If we're documenting the `CachegrindCmd` component above, then let's also do it for this shortcut file as well as the binary-size shortcut file.

site/frontend/src/pages/compare/compile/table/shortcuts/binary-size-shortcut.vue
```diff
@@ -0,0 +1,53 @@
...
+      `cargo build --release`, to compare binary sizes.
+    </Tooltip>
+  </div>
+
+  <pre><code>./target/release/collector binary_stats \
+    +{{ props.baseArtifact.commit }} \
+    --rustc2 +{{ props.artifact.commit }} \
+    --include {{ testCase.benchmark }} \
+    --profile {{ normalizeProfile(testCase.profile) }} \
+    --backend {{ normalizeBackend(testCase.backend) }}</code></pre>
```
lqd, line 38: Would we want the selectbox with "Diff/Baseline commit/Benchmarked commit" choices here, like for the profiling shortcut?

lqd (approved):
r=me with 2 tiny nits

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -455,40 +433,20 @@ onMounted(() => {
...
-          Benchmarked commit
-        </option>
-      </select>
-
-      <CachegrindCmd
-        :commit="profileCommit"
-        :baseline-commit="profileBaselineCommit"
-        :test-case="testCase"
+    <div class="shortcut">
+      <template v-if="props.metric === BINARY_SIZE_METRIC">
```
lqd, line 437: (drive-by note that this check reminded me, would we also want to hide the "frontend/backend/linker" wall-time sections when displaying the size metrics?)

Kobzol: The FE/BE/Linker chart doesn't fit that well into the detail, since it's not metric-specific :/ But I wanted it to be easily discoverable. It really only makes sense currently for wall-time, or task-clock.

Maybe we could somehow highlight this fact under its title, like we do for the 30 day charts, where we explain what do they show.

site/frontend/src/pages/compare/compile/table/shortcuts/binary-size-shortcut.vue
Kobzol: I decided not to add it, to avoid complicating the code. I think that in 99 % you want to see the diff anyway, and getting the binary stats should be much faster than running cachegrind, so I think that it's not worth it to run the command without the diff.

Kobzol:
Added and expanded comments :+1:

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
lqd: I'm finding a bit less helpful than I hoped it would be (as discussed in a recent regression);  we could make a bunch of improvements there. Hiding the chart on the size metrics seems like a straightforward one we probably should do (as it's very big)

site/frontend/src/pages/compare/compile/table/shortcuts/binary-size-shortcut.vue
lqd: Seems fine. And it's also easy to understand once you're not a bigger with rustc-perf that `--rustc2` is the common pattern for diffs.

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
Kobzol: Ok, I'll send a PR, right after this one gets merged.

PR #1774, opened 2023-12-18
Hide section chart for binary size metric

Kobzol:
And also include a small note that the chart is metric agnostic.


lqd (approved):
Sweet, thank you!

PR #1775, opened 2023-12-18
Make fraction chart relative to its corresponding artifact

Kobzol:
Before, it was normalized w.r.t. the longest artifact (duration), which was not very intuitive.
image


Mark-Simulacrum (approved):

lqd (approved):
Cool! 

I still wonder if we should normalize the after percentages to the before total duration, to be able to compare these relative sections.

PR #1781, opened 2023-12-23
Simplify handling of self-profile data in `collector`

Kobzol:
This PR performs several changes to the way we handle self-profile files in the `collector`, in order to simplify it:
- Self-profile data handling is moved from `rustc-fake` to `collector`.
- Instead of invoking the `summarize` binary, the `analyzeme` crate is used directly to parse the self-profile results.
- Handling of the old "3 file format" self-profile data was removed. This format cannot be read by `analyzeme` since 2020, so even if we were able to upload this format to S3 from the `collector` (for some veeeery old `rustc` versions), it could not be read by the site, which uses the `stable` version of `analyzeme` (currently `0.11`).
- The `measureme` binaries are no longer installed on the `collector`, since the binaries are no longer used during benchmarking.

### DIFF
--- a/collector/src/compile/execute/mod.rs
+++ b/collector/src/compile/execute/mod.rs
@@ -8,13 +8,14 @@ use crate::compile::benchmark::BenchmarkName;
 use crate::toolchain::Toolchain;
 use crate::utils::fs::EnsureImmutableFile;
 use crate::{async_command_output, command_output, utils};
+use analyzeme::ArtifactSize;
 use anyhow::Context;
 use bencher::Bencher;
-use database::QueryLabel;
 use std::collections::HashMap;
 use std::env;
 use std::fs;
 use std::future::Future;
+use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::process::{self, Command};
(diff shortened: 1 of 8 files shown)


collector/src/compile/execute/mod.rs
```diff
@@ -564,17 +525,12 @@ enum DeserializeStatError {
...
 enum SelfProfileFiles {
-    Seven {
-        string_data: PathBuf,
-        string_index: PathBuf,
-        events: PathBuf,
-    },
-    Eight {
-        file: PathBuf,
-    },
+    Eight { file: PathBuf },
```
michaelwoerister, line 533: I assume this is referring to the file format version? Maybe this wrapper type is not needed anymore, since it seems to be mainly dealing with the 3-files-vs-1-file distinction.

Kobzol: Yes, this is the file format. I checked `analyzeme` and it seems to return an error when it encounters the 3 file format, so I think that we can delete this.

michaelwoerister:
I don't really know this code -- but looking through the diff the changes make sense to me.

michaelwoerister (approved):

PR #1782, opened 2023-12-23
deps cleanup

klensy:
Cleanups few unused\duped deps. Should be more, but have some issues with github connection, so only this few.


Kobzol (approved):
Looks good, thanks! If CI is green, then LGTM.

PR #1783, opened 2023-12-25
database: move bins to separate crates

klensy:
There no way to do bin only dependencies today, so move bins from database; eventually allows to build less unneeded crates via dependencies.

Played around with crate features; hope there no surprise with cargo feature merge.


Kobzol:
Hi, I wonder what's the motivation for this? Just reducing compile times for the collector/site binaries?

klensy:
Yes, size too; plus more clear separation of functionality/dependencies.

Kobzol:
It seems to remove only two dependencies from collector/site, and doesn't affect their binary sizes a lot (1 KiB for collector it seems), so I'm not sure if it's worth the churn. That being said, it probably doesn't hurt. Let's see what Mark thinks.

klensy:
No immediate effect, yes.

Mark-Simulacrum:
My sense is that this isn't worth the additional complexity.

klensy:
Okay.

PR #1785, opened 2023-12-26
Triage 2023 12 26

Kobzol:
Triage for 2023-12-26.

### DIFF
--- a/triage/2023-12-26.md
+++ b/triage/2023-12-26.md
@@ -1,0 +1,207 @@
+# 2023-12-26 Triage Log
+
+Some of the recent noise swings have been appearing again this week, but luckily
+less often than before. There were a few actual regressions, but most of them
+were localized to a single stress test. On the other hand, there were a few very
+nice wins across the board, especially for check and incremental builds, primarily
+thanks to [#118824](https://github.com/rust-lang/rust/pull/118824) and
+[#119265](https://github.com/rust-lang/rust/pull/119265).
+
+Triage done by **@kobzol**.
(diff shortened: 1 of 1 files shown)


triage/2023-12-26.md
```diff
@@ -0,0 +1,207 @@
...
+| (instructions:u)                   | mean  | range          | count |
+|:----------------------------------:|:-----:|:--------------:|:-----:|
+| Regressions ❌ <br /> (primary)    | 0.9%  | [0.9%, 0.9%]   | 1     |
+| Regressions ❌ <br /> (secondary)  | -     | -              | 0     |
+| Improvements ✅ <br /> (primary)   | -0.6% | [-2.1%, -0.2%] | 101   |
+| Improvements ✅ <br /> (secondary) | -1.2% | [-3.2%, -0.2%] | 65    |
+| All ❌✅ (primary)                 | -0.6% | [-2.1%, 0.9%]  | 102   |
+
+* Very nice check and incremental wins across the board.
+* The single regression is probably just noise returning back to steady state.
```
lqd, line 207: It looks like it has sustained for a bit, interestingly enough, rather than being a simple noise swing. But it's fine, it's strangely been fixed by now by https://github.com/rust-lang/rust/pull/119258, and the situation is better than the previous steady state.... 🤷

lqd (approved):

triage/2023-12-26.md
Kobzol: Oh, it seems that I interpreted these two exact PRs as start of noise - end of noise :laughing: In any case, it's not worth to investigate on this PR, which otherwise had so many wins.

lqd: Agreed that there's no need to look into it at all, it's almost noise like and is already fixed anyways.

PR #1791, opened 2023-12-29
Removes some unused crates, features, dedupes few crates

klensy:
Disables `tera` features as currently unused, drops few unused crates, features; see individual commits.


Kobzol (approved):
Thanks! LGTM if CI is green.

PR #1797, opened 2024-01-08
Use a unique directory for self-profile output

Kobzol:
It is not straightforward to thread all the distinguishing benchmark parameters to `rustc-fake`, so something like this should hopefully be future-proof and unique enough.

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -5,7 +5,7 @@ use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
-use std::time::{Duration, Instant};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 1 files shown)


collector/src/bin/rustc-fake.rs
```diff
@@ -37,6 +37,20 @@ fn run_with_determinism_env(mut cmd: Command) {
     );
 }
 
+// We want each rustc execution to have a separate self-profile directory,
+// to avoid overwriting the results. PID of this process and timestamp should
+// hopefully be unique enough.
```
Mark-Simulacrum, line 42: I think just the PID should be unique?

Mark-Simulacrum (approved):

collector/src/bin/rustc-fake.rs
Kobzol: Linux can recycle PIDs, so in theory another rustc-fake running later could have the same PID.

PR #1798, opened 2024-01-09
Eagerly load self-profile files

Kobzol:
After the recent self-profile refactorings, I thought that we might simplify the self-profile handling even more, by eagerly loading the self-profile file into memory, which avoids the hack with generating a unique self-profile directory.

This also simplifies code.


Kobzol:
I did more benchmarks and it turned out that the S3 upload (which is now more optimized) was the bigger bottleneck.

PR #1799, opened 2024-01-12
Remove trace log that logs touching a file

Kobzol:
I don't think that it provides a lot of value, but it sure does provide a lot of garbage output - easily hundreds of thousands of lines per benchmark, which makes it harder to search through the logs.


nnethercote (approved):

Mark-Simulacrum (approved):
I think it's been useful in the past when debugging rebuild failures (i.e. building too much), but I think that may also have been before we started touching as much as we do now. Seems good.

PR #1806, opened 2024-01-23
Switch fraction chart back to absolute values and improve its legend

Kobzol:
The fraction chart in benchmark detail on the compare page is designed to show two things:

1) What is the fraction of frontend/backend/linker within a single artifact collection
2) How has the absolute time of FE/BE/LNK changed between two artifact collections

However, it is quite challenging (perhaps impossible) to display both of these things in a single chart, without skewing the results.
Before https://github.com/rust-lang/rustc-perf/pull/1775, the chart showed 2), which made 1) skewed. After that PR, the chart shows 1), which doesn't really even allow comparing 2).

Here is an example of the skew that was happening before #1775:
image

In this case, the absolute time of the frontend (purple part) has stayed the same, but its fraction within the collection has changed a lot. But visually, I think that the first thing that my mind when I see it is "the fraction is the same", because the purple rectangle has the same width.

Based on reactions on this PR, it seems that the absolute view is actually OK, and the "skew" isn't problematic for people to undestand. So this PR reverts the display to the absolute value.

Apart from that, the PR also makes tiny stylistic changes in the chart and the legend.

### DIFF
--- a/site/frontend/src/pages/compare/compile/table/sections-chart.vue
+++ b/site/frontend/src/pages/compare/compile/table/sections-chart.vue
@@ -135,6 +135,7 @@ function deactivate() {
 <style scoped lang="scss">
 .wrapper {
   display: flex;
+  align-items: center;
 }
 .chart {
   display: flex;
--- a/site/frontend/src/pages/compare/compile/table/sections-chart.vue
+++ b/site/frontend/src/pages/compare/compile/table/sections-chart.vue
@@ -35,13 +35,24 @@ function calculate_width(value: number, totalDuration: number): string {
   return `${(fraction * 100).toFixed(2)}%`;
 }
 
+function getSectionByName(
+  sections: CompilationSections,
+  name: string
+): CompilationSection | null {
+  const values = sections.sections.filter((s) => s.name === name);
+  if (values.length === 0) {
+    return null;
+  }
+  return values[0];
+}
+
 function formatPercent(
   sections: CompilationSections,
   sectionName: string
 ): string {
-  const values = sections.sections.filter((s) => s.name === sectionName);
-  if (values.length === 0) return "??";
-  const value = values[0].value;
+  const section = getSectionByName(sections, sectionName);
+  if (section === null) return "??";
+  const value = section.value;
   const total = calculateTotalSectionsDuration(sections);
   const percent = (value / total) * 100;
   return `${percent.toFixed(2)}%`;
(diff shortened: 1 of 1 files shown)


nnethercote:
> But visually, I think that the first thing that my mind when I see it is "the fraction is the same", because the purple rectangle has the same width.

I don't feel that at all. I strongly prefer the absolute view and find the relative view quite misleading -- with the relative view I immediately think "why did the frontend get slower and the linker faster?" which is quite wrong. You can use the percentages in the text label to get the relative values.

I won't object to having both if you find the relative view useful, but I think absolute should be the default.

Kobzol:
I guess it's quite subjective, yeah. Before, we had the absolute view, and @lqd found it quite confusing and we converged on the relative view :laughing:

lqd:
My recollection is that it wasn’t absolute before, and there were no time values only relative percentages, and that the visualization changed total size sometimes on the before row, sometimes on the after row. 

So we were displaying relative percentages of a different total time, making the values incomparable. I also had mentioned a case where I wished to see absolute numbers. So I wasn’t able to compare how much of the process a section took nor the absolute time taken by a section. 

As I said before, I believe we probably only need the absolute time each of the section took (albeit noisy wall times), in a table.

Kobzol:
It's possible that the visualization was different before, I changed the code slightly in the meantime. This PR already added the absolute times to the hover labels, btw.

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
```diff
@@ -279,6 +279,20 @@ onMounted(() => {
...
+        {name: "Linker", value: 15},
+      ],
+    };
+    sectionsDetail.value.after = {
+      sections: [
+        {name: "Frontend", value: 20},
+        {name: "Backend", value: 30},
+        {name: "Linker", value: 50},
+      ],
+    };
```
lqd, line 295: Remember to remove this before merging.

lqd (approved):
Maybe the chart legend could also be displaying the expected units, as it's currently showing the relative values even when the chart is in absolute mode.

Kobzol:
> Maybe the chart legend could also be displaying the expected units

It started to overflow when I included more data :man_facepalming: I'll try to tinker with the design a little bit to fix that, the chart probably doesn't need to be so large.

Kobzol:
> and that the visualization changed total size sometimes on the before row, sometimes on the after row.

This also happens with the new "absolute mode". It needs to work like this, otherwise the width of the longer collection could be unbounded (AFAIK).

Kobzol:
I did some changes based on the feedback here. Let me know what you think.

chart

lqd:
Great!

site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
Kobzol: Thanks!

PR #1811, opened 2024-02-06
Update dependencies of the `hashmap` runtime crate

Kobzol:
So that it compiles on nightly.

Discussed [here](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Runtime.20benchmarks.20broken). The `css` runtime benchmark group is not fixed by this PR.

### DIFF
--- a/collector/runtime-benchmarks/hashmap/Cargo.lock
+++ b/collector/runtime-benchmarks/hashmap/Cargo.lock
@@ -4,68 +4,68 @@ version = 3
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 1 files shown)


nnethercote (approved):
This is ok, but it looks like you've done a full `cargo update`. Would just `cargo update -p ahash` work?

Kobzol:
I tried it, but it did not, there were some other dependencies that needed to be updated to get rid of the old version completle.y

PR #1812, opened 2024-02-11
Fix compilation of the css runtime benchmark

Kobzol:
`ahash` has released a minor version that fixes the recent `std_simd` problem.


nnethercote (approved):
Thanks!

Issue #1731, opened 2023-10-02
graphs are busted on mobile


leeoniya:
not sure is 3 columns is a sane choice. might need some extra media queries :)

![Screenshot_20231002-150458082](https://github.com/rust-lang/rustc-perf/assets/43234/a275b862-1a68-44c6-bc8b-730813d16361)

Kobzol:
Yeah the used graphing library is not really very "scalable", and its dimensions are pretty much set by taking the window width and dividing by 3 or 4 :) If you know your way around JS/CSS, we'd be glad for PRs. We try to make the status and compare pages somewhat readable on mobile, but other pages (especially this one) haven't really been in the spotlight.

leeoniya:
> Yeah the used graphing library is not really very "scalable"

i'm the author of said library 😅 . i enjoyed the quotes around "scalable", given the library it replaced (highcharts) choked on the amount of data you guys render :)

it's been a while since https://github.com/rust-lang/rustc-perf/issues/742. it's probably ripe for a version upgrade anyways. maybe i'll find some time.

Kobzol:
Haha, didn't realize that :D Yeah, in terms of performance scalabity, it's great actually! I was talking about responsiveness, of course :)

Maniktherana:
@leeoniya #1826 should address this

PR #1826, opened 2024-02-29
fix: add minimum width to graphs

Maniktherana:
Added a set width and height here as the graphs dont seem to resize beyond 380px on larger screens as it is.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1731

Results from my phone:
IMG_8112


Kobzol (approved):
Looks great, thank you! :)

PR #1828, opened 2024-02-29
Correct sqlite-to-postgres column mapping

### LINKED ISSUE #1827: Incorrect Column Mapping in sqlite-to-postgres
I am encountering an error when attempting to run `site` after migrating via `sqlite-to-postgres`.

### Error Message:
```console
thread 'tokio-runtime-worker' panicked at /workspace/rust-lang/rustc-perf-sqlite2pg/database/src/pool/postgres.rs:617:88:
called `Result::unwrap()` on an `Err` value: "cycles:u is not a codegen backend"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Cause:

It appears the issue is related to a mismatch in column ordering between the sqlite and postgresql.

eth3lbert:
The `backend` column was added later to the `pstat_series` table.
This PR reorders the column order to comply with the schema.

Closes #1827

### DIFF
--- a/database/src/bin/sqlite-to-postgres.rs
+++ b/database/src/bin/sqlite-to-postgres.rs
@@ -7,6 +7,7 @@ use bytes::{BufMut, Bytes, BytesMut};
 use chrono::{DateTime, TimeZone, Utc};
 use database::pool::{postgres, sqlite, ConnectionManager};
 use futures_util::sink::SinkExt;
+use hashbrown::HashMap;
 use serde::{Serialize, Serializer};
 use std::convert::{TryFrom, TryInto};
 use std::io::Write;
@@ -20,6 +21,9 @@ trait Table {
 
     /// Comma-separated list of table's attribute names in SQLite.
     fn sqlite_attributes() -> &'static str;
+
+    /// Comma-separated list of table's attribute names in Postgres.
+    fn postgres_attributes() -> &'static str;
 
     /// Name of `generated always as identity` attribute in Postgres,
     /// if applicable.
@@ -53,6 +57,10 @@ impl Table for Artifact {
     }
 
     fn sqlite_attributes() -> &'static str {
+        "id, name, date, type"
+    }
+
+    fn postgres_attributes() -> &'static str {
         "id, name, date, type"
     }
 
@@ -92,6 +100,10 @@ impl Table for ArtifactCollectionDuration {
         "aid, date_recorded, duration"
     }
 
+    fn postgres_attributes() -> &'static str {
+        "aid, date_recorded, duration"
+    }
+
     fn postgres_generated_id_attribute() -> Option<&'static str> {
         None
     }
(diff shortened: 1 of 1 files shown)


Kobzol:
That's interesting. The backend column was indeed added later, but after [this](https://github.com/rust-lang/rustc-perf/blob/ef64e686eedcdb8ea176d04745fe78cca5a4aff9/database/src/pool/sqlite.rs#L379) migration, backend should be before metric. So it probably depends on which version of the SQLite table you have when doing the export.

eth3lbert:
Interesting! What would you expect in this situation? Perhaps adding some columns mapping strategy?

eth3lbert:
Oh wait! In SQLite, the backend column is placed before the metric column. However, in PostgreSQL, it remains after the metric column.

Kobzol:
Oh, good find, I made this error when writing the last migration. It's a PITA to keep these two DB backends in sync.

eth3lbert:
Yep, and it seems somewhat inevitable. I think it would be reasonable to introduce some column mapping to avoid these kinds of issues in the future.

Kobzol:
Yeah, the conversion script should probably be able to support different column order in both backends.

eth3lbert:
Or, we could try passing columns to function`sqlite_attributes` and return a mapped and ordered columns back.
The column names are available through
```sql
select name from PRAGMA_TABLE_INFO('table_name') order by cid
```

Kobzol:
This is a bit outside of my current SQL-fu, but if it works, let's go for it :)

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -56,6 +60,10 @@ impl Table for Artifact {
         "id, name, date, type"
     }
 
+    fn postgres_attributes() -> &'static str {
+        "id, name, date, type"
```
eth3lbert, line 64: Alternatively, we could use a macro to get the field names of the row struct instead of hardcoding them.

Kobzol (commented):
This looks reasonable. I assume that the conversion now works for you, after the latest commit?

eth3lbert:
Yes, works as expected.

P.S. I just made some minor wording changes and added information about the panics.

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -193,6 +213,10 @@ impl Table for CollectorProgress {
         "aid, step, start, end"
     }
 
+    fn postgres_attributes() -> &'static str {
```
Kobzol, line 216: Hmm, except for this one table, it seems like all the attribute names are the same. When trying to port this table, the script would crash, is that right?

If that is true, then we basically don't need `postgres_attributes`, right? Since if the names don't match, the code panics, and if they do match, we don't need the new method.

eth3lbert: Yes, this is the only table that has a different column name from sqlite.  But it wouldn't panic, because it just matches the column names in postgres.

It is also true that if we could ensure all column names are identical between the two databases, then we could just use one from `sqlite_attributes`.  The `postgres_attributes`  is only useful when postgres has a different column name from sqlite that represents the same content.

Kobzol: Oh, sorry, I didn't realize that the `postgres_columns` logic fixes this. It's a bit complicated, but it works, so let's go with it.

Kobzol (approved):

PR #1831, opened 2024-03-02
To allow the `/perf/status_page` to run even if the current collection is empty

s7tya:
For now, because of some reasons we can't see `/status.html` page locally. This PR fixes that issue.

Changes I made:
- fix collector_progress table's column name of sqlite version: `end_time` to `end`
- give a default value the same as postgres version
- returns empty vec instead of panic if there's no ready commit

Here's the successfully working screenshot of `/status.html` using the latest dump we can get from `fetch-latest` binary.

### DIFF
--- a/database/src/pool/sqlite.rs
+++ b/database/src/pool/sqlite.rs
@@ -1096,7 +1096,7 @@ impl Connection for SqliteConnection {
                 "
                 select
                     step,
-                    end_time is not null,
+                    end is not null,
                     coalesce(end, strftime('%s', 'now')) - start,
                     (select end - start
                         from collector_progress as cp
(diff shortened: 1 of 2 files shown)


Kobzol (commented):
Thank you so much for fixing this, it's great! :)

Could you just please remove the first and last commits, since they basically reverse one another? You can do e.g. `git rebase -i origin/master`, and then replace `pick` with `d` (as in `drop`) before the first and the last commit.

(btw, this is a bit embarrassing, I might have tried to look at this in the past two years to find out that it "only" takes to rename the column and fix the assert :laughing: )

s7tya:
Thanks for your review. I just updated!

Kobzol (approved):
Thank you! :)

PR #1832, opened 2024-03-03
Fix status page layout

### LINKED ISSUE #1815: Fix layout of status page
The [status page](https://perf.rust-lang.org/status.html) shows a table with a timeline of benchmark runs. The table shows errors that have occurred during the run. Currently, if there is an error, and you open it, it will modify the layout of the table, and it will "jump" below the "Current collection" column, because of the usage of `flex-wrap: wrap`.

It would be great to tweak the CSS slightly, so that it doesn't jump around the page.

s7tya:
In this PR, I added default width for every children of `.wrapper`. Also, let it wrap and keep fit within the frame not to break the layout.

Fixes: #1815

### DIFF
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -460,6 +460,11 @@ loadStatus(loading);
   display: flex;
   flex-wrap: wrap;
   column-gap: 100px;
+
+  & > * {
+    max-width: 640px;
+    width: 100%;
+  }
 }
 .current {
   .benchmark {
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -428,6 +428,9 @@ loadStatus(loading);
 
 <style scoped lang="scss">
 .timeline {
+  max-width: 100%;
+  width: fit-content;
+
   table {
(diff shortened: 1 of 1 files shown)


Kobzol:
Thanks! Error jumping has disappeared, good job. I have a few comments:

- I wonder if we could also add `word-break: break-word;` to the error, to handle super long lines (this is already with `word-break`, without it the error just gets super wide):
image

- The current version of the page has different widths for the current collection status and for the table:
image

This makes sense, since the status is typically narrower. After your change, both are 640px, but that makes the status too wide, and the table too narrow:
image
(notice how the dates are on multiple lines unnecessarily)

It would be great if the table rows fit on a single line, at least on typical resolutions (e.g. 1200+px width). For example, we could e.g. make both columns be in a fixed width page (say, 1400px, if available), and then distribute the widths percentually, e.g. 0.4 for the status and 0.6 for the table, plus margins, or something like that. With this PR, the table is currently a bit messy.

s7tya:
Thank you. I changed layout using css grid and configure the column size ratio to 4:6 while the screen size is same/over 1440px. On smaller screens, let them keep their content size.

Kobzol (approved):
Looks great, thanks!

Issue #1821, opened 2024-02-23
Fix collector build error with nightly toolchain


Dicarbene:
While building rustc-perf/collector using toolchain :

```rust
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2024-02-22, rust version 1.78.0-nightly (3406ada96 2024-02-21)
```
I get error like this.

```rust
error[E0635]: unknown feature `stdsimd`
  --> /home/dicarbene/.cargo/registry/src/mirrors.ustc.edu.cn-61ef6e0cd06fb9b8/ahash-0.8.6/src/lib.rs:99:42
   |
99 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
   |
```
I find that stdsimd is deleted in the latest version
![image](https://github.com/rust-lang/rustc-perf/assets/53905542/fa165ad7-25f0-4128-8e17-6d042a5639ec)

[error[E0635]: unknown feature stdsimd](https://users.rust-lang.org/t/error-e0635-unknown-feature-stdsimd/106445/6)

I'm working on it to see if we could  bump up some lib version to fix this.

If anyone encounter this error, try fix it building with stable toolchain or run  `cargo update ahash` should help

Kobzol:
`cargo tree -i ahash` should be helpful to see what packages depend on the old ahash version. But in any case, ahash should now have minor releases with a fix for both 0.7 and 0.8, so just updating ahash should hopefully work.

Dicarbene:
> `cargo tree -i ahash` should be helpful to see what packages depend on the old ahash version. But in any case, ahash should now have minor releases with a fix for both 0.7 and 0.8, so just updating ahash should hopefully work.

Thanks for the help, this is the dependencies tree.
```shell
ahash v0.8.9
├── hashbrown v0.14.3
│   ├── database v0.1.0 (/home/dicarbene/rustc-perf/database)
│   │   ├── collector v0.1.0 (/home/dicarbene/rustc-perf/collector)
│   │   │   └── site v0.1.0 (/home/dicarbene/rustc-perf/site)
│   │   └── site v0.1.0 (/home/dicarbene/rustc-perf/site)
│   ├── hashlink v0.8.4
│   │   └── rusqlite v0.28.0
│   │       └── database v0.1.0 (/home/dicarbene/rustc-perf/database) (*)
│   ├── indexmap v2.1.0
│   │   ├── h2 v0.3.22
[...]

PR #1833, opened 2024-03-03
collector: reduce size

klensy:
First 2 commits fixes build with old `ahash` and new unused imports warning with fresh nightly; 3rd one disables `unicode` feature for `regex` crate as it currently unused in `collector` crate.

Size diff can be see with `cargo build -p collector`, otherwise regex's features will be merged and give no size difference.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1821


Kobzol (approved):
Thank you! :)

Kobzol:
The third commit has broken the `collector binary_stats` command :(

PR #1836, opened 2024-03-06
Support missing tables migration

s7tya:
In this PR, I added support for migrating `runtime_pstat`, `runtime_pstat_series`, `artifact_size` that were missing in current migrators.


Kobzol (approved):
Thank you very much!

TBH, I think that we should just delete the SQLite to Postgres migration program, I don't see any value in it, and it requires maintenance from time to time.

s7tya:
I'm agree with that. However, if we don't remove it, maybe we could add sqlite-to-postgres-to-sqlite. (just an idea though)

Issue #1837, opened 2024-03-06
"Perf builds for each rolled up PR" table gets confused by pipe `|` characters in description


Zalathar:
See https://github.com/rust-lang/rust/pull/121859#issuecomment-1973958656 for an example.

The bot tried to render the rolled-up perf builds in a table, but the line for https://github.com/rust-lang/rust/pull/121784 abruptly cuts off mid-description, because the issue description contains pipe characters (`||`).

Presumably they aren't being escaped by the bot, so the GitHub table renderer treats them as cell boundaries instead of visible pipe characters.

Zalathar:
I imagine some backslash-escaping for pipe characters is needed here:

https://github.com/rust-lang/rustc-perf/blob/9580e15048a6718bda3a07c32b060d7075b89855/site/src/github.rs#L65-L83

Kobzol:
Ah, good point, we should probably escape Markdown there.

Kobzol:
Do you want to try to fix this? :)

PR #1839, opened 2024-03-07
Escape pipe symbol in rolled up build PR descriptions

Kobzol:
I haven't found any crate for Markdown escaping, so I just escaped the pipe manually.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1837

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -75,7 +75,8 @@ pub async fn unroll_rollup(
                         format!("{}…", m.split_at(59).0)
                     }
                 })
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Hm, maybe we can do something like run through https://docs.rs/pulldown-cmark/latest/pulldown_cmark/struct.Parser.html and check we don't find anything? But I guess the risk is pretty low - we just end up with a broken table.

So this seems fine to me.

Mark-Simulacrum (approved):

PR #1846, opened 2024-03-11
Add unit test for `fill_libraries`

chengr4:
I was trying to add unit tests for `fill_libraries` of `ToolchainComponents`, but it seems like `fill_libraries` can't find the path created by `tempfile`, causing `components.lib_rustc` to be `None`. 

```rust
let dir = tempfile::tempdir().unwrap();  // create mock dir
let file_path = dir.path().join("librustc_driver.so"); // create mock file
let _file = File::create(&file_path).unwrap();
```

The three lines can create a temp file at `/var/folders/c_/cz61mkg50r9_z4lyndqjmnlw0000gn/T/.tmpR7l2bp/librustc_driver.so`

I hope to borrow someone's insights about this to find a solution. I have spent some hours testing it and could not understand why. 😓

## Current Result

```
thread 'toolchain::tests::test_fill_libraries' panicked at collector/src/toolchain.rs:616:9:
assertion `left == right` failed
 left: None
 right: Some("librustc_driver.so")
```

---

> BTW, I think it is better to figure out this with PR (find code base easier), If I should put this in the "issue", please let me know.

### DIFF
--- a/collector/src/toolchain.rs
+++ b/collector/src/toolchain.rs
@@ -598,3 +598,50 @@ fn get_lib_dir_from_rustc(rustc: &Path) -> anyhow::Result<PathBuf> {
 
     Ok(Path::new(sysroot_path.as_ref().trim()).join("lib"))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn fill_libraries() {
+        let mut components = ToolchainComponents::default();
+
+        // create mock dir and libraries
+        let temp_dir: tempfile::TempDir = tempfile::tempdir().unwrap();
+        let lib_rustc_path = create_temp_lib_path("librustc_driver.so", &temp_dir);
+        let lib_std_path = create_temp_lib_path("libstd.so", &temp_dir);
+        let lib_test_path = create_temp_lib_path("libtest.so", &temp_dir);
+        let lib_new_llvm_path =
+            create_temp_lib_path("libLLVM.so.18.1-rust-1.78.0-nightly", &temp_dir);
+
+        components.fill_libraries(temp_dir.path()).unwrap();
+
+        assert_eq!(components.lib_rustc, Some(lib_rustc_path));
+        assert_eq!(components.lib_std, Some(lib_std_path));
+        assert_eq!(components.lib_test, Some(lib_test_path));
+        assert_eq!(components.lib_llvm, Some(lib_new_llvm_path));
+    }
+
+    #[test]
+    fn fill_old_llvm_library() {
+        let mut components = ToolchainComponents::default();
+        let lib_old_llvm = "libLLVM-17-rust-1.76.0-stable.so";
+
+        // create mock dir and libraries
+        let temp_dir: tempfile::TempDir = tempfile::tempdir().unwrap();
+        let lib_old_llvm_path = create_temp_lib_path(lib_old_llvm, &temp_dir);
+
+        components.fill_libraries(temp_dir.path()).unwrap();
+
+        assert_eq!(components.lib_llvm, Some(lib_old_llvm_path));
+    }
+
+    fn create_temp_lib_path(lib_name: &str, temp_dir: &tempfile::TempDir) -> PathBuf {
+        let lib_path = temp_dir.path().join(lib_name);
+        // create mock file
+        File::create(&lib_path).unwrap();
+
+        lib_path
+    }
+}


Kobzol:
Hi! Note that `fill_libraries` takes a path to a directory as input, not a path to the library (the library path is what the function computes). So you should put `dir.path()` to the function. Also, you should drop the opened `_file`, to make sure that it will be actually created on disk by the time you call `fill_libraries`.

chengr4:
> Hi! Note that `fill_libraries` takes a path to a directory as input, not a path to the library (the library path is what the function computes). So you should put `dir.path()` to the function. Also, you should drop the opened `_file`, to make sure that it will be actually created on disk by the time you call `fill_libraries`.

Thank you so much !@Kobzol. Your magical "prompt" saves my time 🥹.

Firstly, could you please provide me with two specific filenames related to the old and new versions of `libLLVM.so`? I would like to add them to the test case. 

Additionally, I would like to ask for your assistance in confirming that my **"naming"** is OK.

And If you notice any other issues elsewhere, please let me know 🙏.

Kobzol:
> Firstly, could you please provide me with two specific filenames related to the old and new versions of libLLVM.so? I would like to add them to the test case.

Old LLVM can be e.g. `libLLVM-17-rust-1.76.0-stable.so` (you can check `$HOME/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib` to find an example LLVM library path).
New LLVM can be e.g. `libLLVM.so.18.1-rust-1.78.0-nightly`.

Naming is fine. I prefer not to prefix tests with `test_`.

chengr4:
@Kobzol 
I add the test case for LLVM in random way. 
Please check it out whether it is appropriate.

---

BTW,

my `$HOME/.rustup/toolchains/stable-aarch64-apple-darwin/lib` shows
```
librustc-stable_rt.asan.dylib          librustc_driver-1230441069d3e3cb.dylib
librustc-stable_rt.lsan.dylib          libstd-b0083070c892a1db.dylib
librustc-stable_rt.tsan.dylib          libtest-597563d0b411a8e9.dylib
```
it looks like aarch MAC is using an another extension. I have no idea whether this will be a problem.

Kobzol:
Well, since the collector doesn't support macOS, it is not a problem currently :D

collector/src/toolchain.rs
```diff
@@ -598,3 +598,58 @@ fn get_lib_dir_from_rustc(rustc: &Path) -> anyhow::Result<PathBuf> {
...
+        let mut components = ToolchainComponents::default();
+        let mut libraries_to_read = vec!["librustc_driver.so", "libstd.so", "libtest.so"];
+        let llvms = vec![
+            "libLLVM-17-rust-1.76.0-stable.so",    // old LLVM
+            "libLLVM.so.18.1-rust-1.78.0-nightly", // new LLVM
+        ];
+        let mut libraries_path_map = HashMap::new();
+
+        // create a random case for testing LLVM
+        let mut rng = rand::thread_rng();
```
Kobzol, line 620: This seems to be unnecessarily complicated and over-engineered :sweat_smile: And also having random/nondeterministic tests is not a good idea, when it's not needed.

Could you please simplify the test to this:
1) Create a directory with LLVM, rustc and libstd `.so` files
2) Call `fill_libraries` on the directory
3) Assert that the paths of the component attributes equals the created paths.

And then you can create two of these tests, one for the old and one for the new LLVM format.

collector/src/toolchain.rs
```diff
@@ -598,3 +598,57 @@ fn get_lib_dir_from_rustc(rustc: &Path) -> anyhow::Result<PathBuf> {
...
+    #[test]
+    fn fill_old_llvm_library() {
+        let mut components = ToolchainComponents::default();
+        let lib_old_llvm = "libLLVM-17-rust-1.76.0-stable.so";
+
+        // create mock dir and libraries
+        let temp_dir: tempfile::TempDir = tempfile::tempdir().unwrap();
+        let lib_old_llvm_path = create_temp_lib_path(lib_old_llvm, &temp_dir);
+
+        let _ = components.fill_libraries(temp_dir.path());
```
Kobzol, line 638:
```suggestion
        components.fill_libraries(temp_dir.path()).unwrap();
```
to avoid ignoring a potential error. (same in the other test).

collector/src/toolchain.rs
```diff
@@ -598,3 +598,57 @@ fn get_lib_dir_from_rustc(rustc: &Path) -> anyhow::Result<PathBuf> {
...
+
+        assert_eq!(components.lib_rustc, Some(lib_rustc_path));
+        assert_eq!(components.lib_std, Some(lib_std_path));
+        assert_eq!(components.lib_test, Some(lib_test_path));
+        assert_eq!(components.lib_llvm, Some(lib_new_llvm_path));
+
+        // close the temp dir
+        let _ = temp_dir.close();
```
Kobzol, line 626:
```suggestion
```
This is not needed, it will be destroyed automatically

collector/src/toolchain.rs
```diff
@@ -598,3 +598,57 @@ fn get_lib_dir_from_rustc(rustc: &Path) -> anyhow::Result<PathBuf> {
...
+    }
+
+    fn create_temp_lib_path(lib_name: &str, temp_dir: &tempfile::TempDir) -> PathBuf {
+        let lib_path = temp_dir.path().join(lib_name);
+        // create mock file
+        let lib_file = File::create(&lib_path).unwrap();
+        // close the file to release the resources.
+        drop(lib_file);
```
Kobzol, line 651:
```suggestion
```
Since this is now in a function, this is no longer needed, the file will be dropped (and closed) when the function call ends.

Kobzol (changes requested):
Thanks! Now it's much cleaner. A small cleanup and it's ready to go :)

Kobzol:
Thank you, now it looks good :) Before merging, it would be nice if you squashed the commits - 10 commits is quite a lot for two added tests. You can do it e.g. like this:
```bash
$ git fetch origin master
$ git checkout unit-test
$ git reset --soft $(git merge-base HEAD master)
$ git commit -m"..." # Write a commit message
$ git push --force-with-lease
```

chengr4:
made it!

Kobzol (approved):
Thanks!

PR #1847, opened 2024-03-12
Address reported issues for the detail chart

eth3lbert:
Fix reported issues https://github.com/rust-lang/rustc-perf/pull/1842#issuecomment-1991461734 and https://github.com/rust-lang/rustc-perf/pull/1842#issuecomment-1992444926

### DIFF
--- a/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
+++ b/site/frontend/src/pages/compare/compile/table/benchmark-detail.vue
@@ -258,7 +258,7 @@ function graphLink(
   testCase: CompileTestCase
 ): string {
   // Move to `30 days ago` to display history of the test case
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Looks great. It's surprising how complicated it is to think about all the various situations that can arise in this seemingly simple situation :sweat_smile:

eth3lbert:
Ah, I forgot to change the word `benchmarked` to `end` 😅

Kobzol:
The original motivation for the `benchmarked` name was that mostly you look at the page when you open a perf. result from a PR that was benchmarked very recently, so the word makes sense there. Of course, sometimes you also open it later, and then it doesn't make that much sense, since both the start and end commits are "benchmarked", so `end` probably makes more sense. But it's a very small thing.

PR #1854, opened 2024-03-14
Print correct port in startup message

### LINKED ISSUE #1853: Server doesn't print correct port when set to something other than default.
If environment variable `PORT` is set, the site still prints `View the results in a web browser at 'http://localhost:2346`.
If unintended, I can create a pr.

Zentrik:
I'm fairly new to rust, so I'm not sure if there is a difference between `{}` and `{:?}` for `u16`s but this seems to work.
Fixes #1853

### DIFF
--- a/site/src/main.rs
+++ b/site/src/main.rs
@@ -42,7 +42,9 @@ async fn main() {
                 "Loading complete; {} commits and {} artifacts",
                 commits, artifacts,
             );
-            eprintln!("View the results in a web browser at 'http://localhost:{}/compare.html'", port);
+            eprintln!(
+                "View the results in a web browser at 'http://localhost:{port}/compare.html'"
+            );
             // Spawn off a task to post the results of any commit results that we
             // are now aware of.
             site::github::post_finished(&res).await;


site/src/main.rs
```diff
@@ -38,17 +42,13 @@ async fn main() {
                 "Loading complete; {} commits and {} artifacts",
                 commits, artifacts,
             );
-            eprintln!("View the results in a web browser at 'http://localhost:2346/compare.html'");
+            eprintln!("View the results in a web browser at 'http://localhost:{}/compare.html'", port);
```
Kobzol, line 45:
```suggestion
            eprintln!("View the results in a web browser at 'http://localhost:{port}/compare.html'");
```

Kobzol: You can also use inline arguments, like this :) I wouldn't even mention it, but you'll need to reformat the code to pass CI anyway (by running `cargo fmt`), so you can make this change while you're at it :)

Kobzol (commented):
Thank you! :) Left one comment.

Kobzol (approved):
Thanks! :)

PR #1855, opened 2024-03-14
Fix out of bounds access in `quartiles`

Zentrik:
Admittedly, code isn't particularly nice but I don't see any obvious better way as `len=1,2` have to be special cased.

For an examples consider len=2, h1_end underflows and hw_begin = 2.


Zentrik:
Yes when I set `NUM_PREVIOUS_COMMITS` to 3, I assume `pcs` then had length `2` and so there was a panic on `let q1 = median(&pcs[..=h1_end]);`

Kobzol:
Could you please just squash the commits into a single one?

Zentrik:
I presume if `len=0` this will still crash actually, I could add some code to in that case default to the `DEFAULT_SIGNIFICANCE_THRESHOLD` if you want.

Kobzol (approved):
No, that's fine. Thanks!

PR #1860, opened 2024-03-15
Load 'perf/info' and 'perf/get' in Comparison page asynchronously

Zentrik:
Previously, 'perf/info' would load then a request would be made to 'perf/get' in `loadCompareData`. By moving `await` after `loadCompareData`, `perf/get` is requested then `perf/info` without a wait in between. I don't know if `await` is necessary but doesn't seem to hurt.

I don't know much js, but hopefully this is fine as it seems to work locally. It saves about 50-100ms for me.

### DIFF
--- a/site/frontend/src/pages/compare/page.vue
+++ b/site/frontend/src/pages/compare/page.vue
@@ -112,7 +112,6 @@ const runtimeSummary: Ref<SummaryGroup | null> = ref(null);
 
 const loading = ref(false);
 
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
This could be a bit dangerous if some of the TS setup code started using the `info` variable. But as long as the `await` is there, and only the UI template uses it, it should be fine.

Good idea, thanks!

PR #1862, opened 2024-03-16
Add brotli compression of static assets

Zentrik:
Let me know if the rust code isn't idiomatic. This achieves a 200kb reduction in compare.js size from ~292kb to ~77kb.

### DIFF
--- a/site/frontend/.parcelrc
+++ b/site/frontend/.parcelrc
@@ -1,0 +1,9 @@
+{
+    "extends": "@parcel/config-default",
+    "compressors": {
+        "*.{css,js}": [
+          "...",
+          "@parcel/compressor-brotli"
+        ]
+      }
+}
--- a/site/src/resources.rs
+++ b/site/src/resources.rs
@@ -17,6 +17,7 @@ struct StaticAssets;
 #[derive(RustEmbed)]
 #[folder = "frontend/dist"]
 #[include = "*.js"]
+#[include = "*.br"]
 #[include = "*.css"]
 struct StaticCompiledAssets;
 
@@ -25,6 +26,11 @@ struct StaticCompiledAssets;
 #[folder = "frontend/templates/"]
 #[include = "*.html"]
 struct TemplateAssets;
+
+pub enum Payload {
+    Compressed(Vec<u8>),
+    Uncompressed(Vec<u8>),
+}
 
 pub struct ResourceResolver {
     tera: RwLock<Tera>,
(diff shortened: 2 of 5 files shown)


site/frontend/.parcelrc
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "@parcel/config-default",
+    "compressors": {
+        "*.{html,css,js,svg,map}": [
```
Kobzol, line 4:
```suggestion
        "*.{css,js}": [
```
I wouldn't worry about compressing the `map` files, since we don't ship them to production anyway. We also don't have `.svg` nor `.html` files currently in the output of `parcel`.

site/src/resources.rs
```diff
@@ -8,6 +8,7 @@ use rust_embed::RustEmbed;
 #[derive(RustEmbed)]
 #[folder = "frontend/static/"]
 #[include = "*.js"]
+#[include = "*.br"]
```
Kobzol, line 11:
```suggestion
```
We don't have any brotli files in the `static` directory.

site/src/server.rs
```diff
@@ -594,7 +594,11 @@ lazy_static::lazy_static! {
 }
 
 /// Handle the case where the path is to a static file
-async fn handle_fs_path(req: &Request, path: &str) -> Option<http::Response<hyper::Body>> {
+async fn handle_fs_path(
+    req: &Request,
+    path: &str,
+    allow_compression: bool,
```
Kobzol, line 600:
```suggestion
    use_compression: bool,
```
I think that this more clearly states what we want to do. Same in `get_static_asset`.

site/src/resources.rs
```diff
@@ -39,10 +41,23 @@ impl ResourceResolver {
         })
     }
 
-    pub fn get_static_asset(&self, path: &str) -> Option<Vec<u8>> {
-        StaticCompiledAssets::get(path)
-            .or_else(|| StaticAssets::get(path))
-            .map(|file| file.data.to_vec())
+    pub fn get_static_asset(&self, path: &str, allow_compression: bool) -> (Option<Vec<u8>>, bool) {
```
Kobzol, line 44: Returning a tuple of the payload and a `bool` is ugly, and allows representing states that don't make sense (`(None, true)`). Could you please create an enum, e.g. `Payload`, which will have `Compressed` and `Uncompressed` variants, and then return `Option<Payload>`?

site/src/server.rs
```diff
@@ -627,17 +633,28 @@ async fn handle_fs_path(req: &Request, path: &str) -> Option<http::Response<hype
...
-    match p.extension().and_then(|x| x.to_str()) {
+    match path_extension.and_then(|x| x.to_str()) {
         Some("html") => response = response.header_typed(ContentType::html()),
         Some("png") => response = response.header_typed(ContentType::png()),
         Some("json") => response = response.header_typed(ContentType::json()),
         Some("svg") => response = response.header("Content-Type", "image/svg+xml"),
         Some("css") => response = response.header("Content-Type", "text/css"),
         Some("js") => response = response.header("Content-Type", "application/javascript"),
+        Some("br") => {
+            response = response.header(
```
Kobzol, line 653: Let's set this header near line 638, instead of changing the extension. When we compress the data, we should still keep e.g. the `Content-Type` encoding for the original extension.

Kobzol (changes requested):
This is a great idea! https://github.com/rust-lang/rustc-perf/pull/1201 enabled compression on-the-fly, but since our webserver is not very powerful, and on-the-fly compression provides much worse results with the current options (the `compare.js` script still has more than 200 KiB), and we already bake static assets into the binary, this seems like a great way to reduce payload size.

I left some comments about the implementation.

site/src/server.rs
Zentrik: I was concerned that the original requested path might be for a '.br' file. I can remove this if you don't think that would happen.

Kobzol: There's not really any reason to support that, I think :)

Kobzol (approved):
Awesome, and the code is now shorter too! Thanks a lot :)

PR #1863, opened 2024-03-16
Attemp to find earlier commit for the detail chart

### LINKED ISSUE #1861: Selecting earliest commit in graph throws an error
image
If I select the left most commit in this graph, the url is 'https://perf.rust-lang.org/compare.html?start=null&end=37b65339c8cf1d18e331f184dbf70bbddcb1e4f7&stat=instructions:u' which throws an error due to `start` being `null`.

This seems to affect all the graphs on the site.
Sorry if intended.

eth3lbert:
Resolve #1861

### DIFF
--- a/site/frontend/src/graph/render.ts
+++ b/site/frontend/src/graph/render.ts
@@ -356,8 +356,11 @@ function genPlotOpts({
       },
       tooltipPlugin({
         onclick(_u, _seriesIdx, dataIdx) {
+          // No earlier data to show
+          if (dataIdx == 0) return;
+
           let thisCommit = commits[dataIdx][1];
-          let prevCommit = (commits[dataIdx - 1] || [null, null])[1];
+          let prevCommit = commits[dataIdx - 1][1];
           window.open(
             `/compare.html?start=${prevCommit}&end=${thisCommit}&stat=${stat}`
           );


Kobzol (commented):
Thanks! Now I understand what you have meant. Yeah, so basically you changed the range of the data that we query, but you did by changing the query parameters on the frontend, rather than changing the backend logic.

I still think that it would be much easier to just display `commits[1:]` in the graph, it would help us avoid all this complexity. Now you're loading an additional day of data, and the computation of the indices is a quite messy code that is difficult to grasp. The frontend code was built on multiple such special cases over the years, and I'd like to avoid adding more, if possible :sweat_smile:

Actually, an even simpler option would be to just avoid opening the link if the previous commit is `null`, that should be a one-line change. I think that might even make more sense, since in the graph you don't see any previous commit for the leftmost point to compare to, so it's probably not very useful to even allow opening the compare page for the leftmost commit.

eth3lbert:
Fair enough. Since I'm not delving deeply into uplot, I also consider this a hacky solution that adds extra complexity 😅 
Feel free to just close the issue.

eth3lbert:
> Actually, an even simpler option would be to just avoid opening the link if the previous commit is null, that should be a one-line change.

I second this one. It makes more sense to me.

Kobzol:
I made the change to this PR, with you as a co-author. Thanks for looking into this!

Kobzol (approved):

PR #1864, opened 2024-03-16
Add "Rust codegen" as a new compilation section

Kobzol:
Inspired by discussions about my [recent blog post](https://kobzol.github.io/rust/rustc/2024/03/15/rustc-what-takes-so-long.html), I thought that it would be nice if we split the "Backend" section into the actual backend (e.g. LLVM) and the Rust part that monomorphizes stuff and generates LLVM IR from MIR. This PR does that:

image

### DIFF
--- a/site/src/self_profile.rs
+++ b/site/src/self_profile.rs
@@ -240,7 +240,6 @@ async fn download_and_analyze_self_profile(
 /// backend, linker) from the self-profile queries.
 fn compute_compilation_sections(profile: &ProfilingData) -> Vec<CompilationSection> {
     let mut first_event_start = None;
-    let mut frontend_end = None;
     let mut backend_start = None;
     let mut backend_end = None;
     let mut linker_duration = None;
@@ -250,10 +249,7 @@ fn compute_compilation_sections(profile: &ProfilingData) -> Vec<CompilationSecti
             first_event_start = event.payload.timestamp().map(|t| t.start());
         }
 
-        if event.label == "analysis" {
-            // End of "analysis" => end of frontend
-            frontend_end = event.payload.timestamp().map(|t| t.end());
-        } else if event.label == "codegen_crate" {
+        if event.label == "codegen_crate" {
             // Start of "codegen_crate" => start of backend
             backend_start = event.payload.timestamp().map(|t| t.start());
         } else if event.label == "finish_ongoing_codegen" {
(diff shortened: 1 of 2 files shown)


site/src/self_profile.rs
```diff
@@ -240,19 +240,20 @@ async fn download_and_analyze_self_profile(
...
         if first_event_start.is_none() {
             first_event_start = event.payload.timestamp().map(|t| t.start());
         }
 
-        if event.label == "analysis" {
-            // End of "analysis" => end of frontend
-            frontend_end = event.payload.timestamp().map(|t| t.end());
+        if event.label == "collect_and_partition_mono_items" {
+            // We mark this query as "Rust codegen", where monomorphization and MIR-to-backend
+            // conversion happens.
```
bjorn3, line 255: collect_and_partition_mono_items happens before the translation from MIR to llvm ir. It is the step which determines which instances to actually codegen and into which codegen unit. I believe it also implicitly loads the MIR for dependencies.

Kobzol: Aah, you're right, thanks. The MIR-to-IR translation happens in `codegen_module`, right? That explains the staircase shape:
image

Hmm, then this becomes quite tricky, as the MIR-to-IR computation is overlaid with LLVM, and we can no longer render it as a separate part.

I still think that the mono collector is a reasonably self-contained part and it could be useful to visualize it (with a more accurate description, of course). What do you think?

bjorn3: > The MIR-to-IR translation happens in codegen_module, right?

Yes

> Hmm, then this becomes quite tricky, as the MIR-to-IR computation is overlaid with LLVM, and we can no longer render it as a separate part.

If you make the graph represent cpu-clock (time spent on each individual thread added together) rather than task-clock or wall-time you can add all `codegen_module` queries together to form "Rust codegen" time and all `codegen_module_optimize` queries to form "backend" time.

Kobzol: `cpu-clock` is not supported by `self-profile`, and also if we added time of multiple threads together, it would no longer make sense to show this in the same chart as a fraction of the whole compilation time (it could be another chart, but that's probably overkill).

Kobzol:
Because of the overlap of "Rust codegen" with "LLVM codegen", it does not make sense to visualize it in the current chart. So I'll close this PR. I'll make a new one with just the last commit.

PR #1865, opened 2024-03-19
Improve message printed when the site starts

Kobzol:
The distinction between commits and artifacts is confusing (local compiler versions were actually under artifacts).

### DIFF
--- a/site/src/main.rs
+++ b/site/src/main.rs
@@ -38,10 +38,7 @@ async fn main() {
                 eprintln!("Loading complete but no data identified; exiting.");
                 std::process::exit(1);
             }
-            eprintln!(
-                "Loading complete; {} commits and {} artifacts",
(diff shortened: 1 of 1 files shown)


workingjubilee:
Seems good to me. It would be fine to say "N things, of which X are this and Y are that", I think, but it's not clear that's actually more informative.

Kobzol:
If you run the website locally, you will usually not have any commits in the DB anyway (unless you have downloaded a DB dump from the live server). And for the user of the website, the distinction isn't actually important, for the compare page both commits and artifacts are mostly equal. So I'd opt into using the simpler variant with a single number.

workingjubilee:
If the distinction is mostly useless even to those who might care about it, that makes sense, yeah!

lqd (approved):

PR #1867, opened 2024-03-19
Triage 2024 03 19

Kobzol:
[Rendered](https://github.com/Kobzol/rustc-perf/blob/9b4df43b82c5d0fd214d6fae1b8ba4f5e3fdfec1/triage/2024-03-19.md).

### DIFF
--- a/triage/2024-03-19.md
+++ b/triage/2024-03-19.md
@@ -1,0 +1,188 @@
+# 2024-03-19 Triage Log
+
+Even though the summary might not look like it, this was actually a relatively quiet week,
+with a few small regressions. The large regression that is also shown in the summary table was
+caused by extending the verification of incremental compilation results.
+However, this verification is not actually fully enabled by default, so these regressions are mostly
+only visible in our benchmarking suite, which enables the verification to achieve more deterministic
+benchmarking results. One small regression was also caused by enabling frame pointers for the Rust
+standard library, which should improve profiling of Rust programs.
+
+Triage done by **@kobzol**.
+Revision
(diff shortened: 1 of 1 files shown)


triage/2024-03-19.md
```diff
@@ -0,0 +1,189 @@
...
+|  Improvements ✅ <br /> (primary)  |   -   |       -        |   0   |
+| Improvements ✅ <br /> (secondary) | -1.3% | [-1.3%, -1.3%] |   1   |
+|         All ❌✅ (primary)          | 2.0%  |  [0.2%, 7.8%]  |  193  |
+
+* This PR adds additional verification to the compiler that checks if incremental compilation didn't
+  encounter problems.
+* This result looks quite bad, but that is mostly caused by the configuration of our benchmark
+  suite,
+  which fully enables this verification to make benchmarking more deterministic.
+* Normal builds of user code should only pay 1/32 of the cost here, so it should not be an issue.
```
lqd, line 73: I had the same misconception, but I believe is wrong. My understanding (which could be wrong) is: the checks didn’t change themselves, but new checks were added under the same feature gate, which rustc-perf enables. I don’t believe the checks are done outside of debug-assertions, and the "force incremental verification checks" opt-in, i.e. not present in the 1/32nd of all query calls.

Kobzol: Ah, I see, makes sense. So in other words, it shouldn't be a change for normal Rust code.

Kobzol: Ah, I see, makes sense. So in other words, it shouldn't be a change for normal Rust code.

Kobzol: Ah, I see, makes sense. So in other words, it shouldn't be a change for normal Rust code.

lqd: I believe so, from what I remember of the PR

Kobzol: Ok, reworded.

lqd (approved):

PR #1868, opened 2024-03-19
Remove runtime_benchmark table temporary

chengr4:
@Kobzol 

I think we may have one in the future, so I do this way. But if you prefer to remove it, I will just do so.

---

ref: https://rust-lang.zulipchat.com/#narrow/stream/421156-gsoc/topic/Idea.3A.20multiple.20collectors.20for.20Rust.20benchmark.20suite/near/427639989

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -39,7 +39,6 @@ Here is the diagram for compile-time benchmarks:
 ```
 
 For runtime benchmarks the schema very similar, but there are different table names:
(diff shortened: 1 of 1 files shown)


Kobzol:
We don't know if we will eventually add it, and even if yes, talking some non-existing in the documentation is confusing :) So please remove it altogether.

chengr4:
fixed it

Kobzol (approved):
Thank you!

PR #1870, opened 2024-03-20
fix typo

RoboSchmied:
fix typo: `endianess` -> `endianness`


Kobzol:
Hi, this PR doesn't change the code of `rustc-perf`, but code that is inside a compile-time benchmark that is just stored inline in this repository. As we have noted a few times in the past (e.g. https://github.com/rust-lang/rustc-perf/pull/1510), we don't accept PRs for this code, since it is supposed to be immutable, and the only changes we make to it is if it stops compiling.

So I'm going to go ahead and close this PR. Thank you nonetheless!

Issue #1869, opened 2024-03-20
"Benchmarks for artifact sizes" confuses me


chengr4:
According to @Kobzol 's [blog](https://kobzol.github.io/rust/rustc/2023/08/18/rustc-benchmark-suite.html), The main function of `helloworld-tiny` and `ripgrep-13.0.0-tiny` are to monitor the size of  Rust executable

> Refer to blog: There are also two secondary benchmarks (`helloworld-tiny` and `ripgrep-13.0.0-tiny`), which exist solely to monitor the size of a “minimal-sized” Rust executable, which is compiled with compiler flags that should favour small executable size.

However, on the [graphs](https://perf.rust-lang.org/index.html) page, the title states 'Benchmarks for artifact sizes', but the metric displayed is "CPU instructions", which confuses me.

<img width="1105" alt="Screenshot 2024-03-20 at 11 43 43 AM" src="https://github.com/rust-lang/rustc-perf/assets/54136390/e856ab62-bda5-4581-bcf9-e75b73c6839c">

I believe it should present the value of `size:xxxxxx`.

## Solutions I come up with

1. (Simple) Write a note below the title explaining that it only makes sense when the `kind` is `size:xxxxxx`.
2.  We should separate the APIs and let them have their own behavior

---

related to : #1817

Kobzol:
Hi, the whole graphs page is displayed with a single metric, so if you choose instructions, you'll see instruction counts for all the benchmarks.

While the binary size of these tiny benchmarks is indeed the main motivation for why we have added them, their compilation time ia actually not uninteresting! They use different compiler flags, and LTO, so it is also interesting to look at how does their compilation time change.

Kobzol:
Feel free to think of a better label. For example, `Benchmarks optimized for binary size` or something like that.

PR #1871, opened 2024-03-20
Make the title clear

chengr4:
Hi @Kobzol,

I asked ChatGTP by 

```
"benchmarks that is compiled with compiler flags that should favor small executable size" make a title for this.
```

It gave me `Compiler-Optimized Benchmarks for Small Executable Sizes`

So I decided to combine yours and his 😀.

What do you think? you are boss you call 😎

BTW, I also asked why using plural for "Sizes" here is its answer:

```
"Small Executable Sizes" to indicate that there are multiple sizes being referred to. 
Even though "Executable" is singular, 
"Sizes" is used to denote that there are various sizes or a range of sizes under consideration.

It's a bit idiomatic in English to use the plural form in such cases, 
even if technically referring to a singular concept, 
to encompass the diversity or range implied by the context.
```

---

Fixes: #1869

### DIFF
--- a/site/frontend/src/pages/graphs/page.vue
+++ b/site/frontend/src/pages/graphs/page.vue
@@ -149,7 +149,7 @@ loadGraphData(selector, loading);
       style="margin-top: 50px; border-top: 1px solid #ccc"
     >
       <div style="padding: 20px 0">
-        <strong>Benchmarks for artifact sizes</strong>
+        <strong>Benchmarks optimized for small binary size</strong>
       </div>
       <div id="size-charts"></div>
     </div>


site/frontend/src/pages/graphs/page.vue
```diff
@@ -149,7 +149,7 @@ loadGraphData(selector, loading);
       style="margin-top: 50px; border-top: 1px solid #ccc"
     >
       <div style="padding: 20px 0">
-        <strong>Benchmarks for artifact sizes</strong>
+        <strong>Benchmarks Optimized for Small Binary Sizes</strong>
```
Kobzol, line 152:
```suggestion
        <strong>Benchmarks optimized for small binary size</strong>
```

Kobzol (commented):
Thanks. I would prefer the variant that I suggested in the comment, otherwise OK.

site/frontend/src/pages/graphs/page.vue
chengr4: fixed it

Kobzol (approved):
Thanks :)

PR #1872, opened 2024-03-20
Fix #1829

### LINKED ISSUE #1829: Add a runtime benchmark for `async`
It would be nice to have some runtime benchmark for `async/await` usage, Futures etc., e.g. using `tokio`.

Daksh-10:
In #1829,
I added a runtime benchmark for async/await, it reads two files asynchronously for 1000 times and provides the duration for which the code works.


collector/runtime-benchmarks/asynctest/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+
+            for _ in 0..1000 {
+            rt.block_on(async_operation());
+            }
+
+            let end_time = Instant::now();
+
+            let duration = end_time - start_time;
+            move || {
+                duration
```
Kobzol, line 44: The actual benchmark is only what is in the closure returned fro the function. So here you are benchmarking only the line `duration` :) Ideally, you should prepare the runtime in the first closure, and then perform the `block_on` operation in the second closure. Check the other benchmarks to see how this works.

collector/runtime-benchmarks/asynctest/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    let mut total_characters = 0;
+    for line in reader.lines() {
+        let line = line?;
+        total_characters += line.len();
+    }
+
+    Ok(total_characters)
+}
+
+async fn async_operation() {
```
Kobzol, line 19: We already have a reasonably large text file with Sherlock Holmes text in the `runtime-benchmarks/data` directory. Check how do other benchmarks use it, and try to reuse that same file.

collector/runtime-benchmarks/asynctest/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    let f1 = read_the_textfile("./collector/runtime-benchmarks/Async2/src/poem.txt").await.expect("Can't read file");
+    let f2 = read_the_textfile("./collector/runtime-benchmarks/Async2/src/poem2.txt").await.expect("Can't read file");
+    let total_char_count =f1+f2;
```
Kobzol, line 22: This benchmark only contains two `await` points, and the rest of the code is synchronous. It would be great to try to do something that uses `await `much more. For example, iterate over the lines of the file and run some async code for each line, or something like that.

collector/runtime-benchmarks/asynctest/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "asynctest"
```
Kobzol, line 2: Eventually, this should be ideally called e.g. `async`, rather than `asynctest`.

Kobzol (commented):
Hi, thank you for working on this! I left some comments.

Daksh-10:
I will try to fix them, thank you for your suggestions :)

collector/runtime-benchmarks/asynctest/src/main.rs
Daksh-10: I tried using the decompress_file function, but it returns the output in vec<u8> form and using async on it didn't made much sense, so I used GzDecoder

collector/runtime-benchmarks/async/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+    (total_char,duration.as_millis())
+}
+
+fn main() {
+    run_benchmark_group(|group| {
+        group.register_benchmark("Async", || {
+            // This closure should prepare data that will be needed for the benchmark (if any),
+            // and then return a closure that will actually be benchmarked/profiled.
+            // Create a Tokio runtime
+            let rt = Runtime::new().unwrap();
```
Kobzol, line 39: Let's use a single-threaded `tokio` runtime, to try to reduce noise a bit.

collector/runtime-benchmarks/async/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+async fn line_char_count(line: String) -> usize {
+    let line_count = line.chars().count();
+    line_count
+}
+
+async fn async_operation() -> (usize, u128) {
+    let start_time = Instant::now();
+
+    let file = File::open("./collector/runtime-benchmarks/data/sherlock.txt.gz").expect("can't read a file");
```
Kobzol, line 22: Let's use relative paths and inline the file into the binary to avoid I/O during the benchmark ( see https://github.com/rust-lang/rustc-perf/blob/c1f870e4f1089ae12360e034657256382fe2ddcf/collector/runtime-benchmarks/compression/src/main.rs#L8).

collector/runtime-benchmarks/async/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+    let decoder = GzDecoder::new(file);
+    let reader2 = BufReader::new(decoder);
+    let mut total_char = 0;
+    total_char_count(reader2, &mut total_char).await;
+
+    let end_time = Instant::now();
+    let duration = end_time - start_time;
+    (total_char,duration.as_millis())
```
Kobzol, line 30: There's no need to measure the time here (the benchmark suite takes care of that), just return `total_char` to avoid it being optimized.

Kobzol (commented):
Thanks! I left a few more comments.

Could you please run the benchmark locally and post the instruction count/cycle results etc.?

Daksh-10:
Yes, I will send the ss for instructions/cycle next time.

Kobzol:
By the way, if you want to update to the newest version of master (and/or force push), you can do that like this:
```bash
$ git fetch origin master:master
$ git rebase master
$ git push --force
```

Daksh-10:
I did the changes you said, but the relative path was not working on my local machine ( if you say, I can change it to "../../data/sherlock.txt.gz" )

and here is the table 
image

Kobzol (commented):
Thanks, I left more comments.

I went through a profile of the benchmark (using `samply`), and it looks like it mostly stresses the decompression and memory movement. OTOH, there are 15 thousand lines in the file, so it should execute at least that number of `await`s, so if something went wrong there, we could perhaps observe it in the results.

The standard deviation of the executed instructions is remarkably low, which is great. I think that for a start, it looks good.

Daksh-10:
There is a problem, I changed my setup to a different computer and then setup the project locally, but when I run the runtime benchmarks it runs for every benchmark, but the one I added (the async one). It gives this error 
image

I think the issue is in the configuration of the group (i.e. I used 'cargo new async' to create the folder ) I added, but I don't know how it might be affecting the compilation. Can you help me with this, I gave it some time but wasn't able to find the solution.
Also, I obviously added it in the 'runtime-benchmark' and gave it the benchlib dependency.
image

Daksh-10:
> Can you compile the project manually, by going into `collector/runtime-benchmarks/async` and running `cargo build`?

I build it and now its running, should have tried this lol, but why is it running now?? 
Are the following 'compiling' different ...
image

... from the this compiling
image

Kobzol:
I don't know what was causing your previous error. Perhaps it was a missing Cargo.lock file (that you haven't committed yet in this PR)? When you executed `cargo build`, cargo generated it.

Daksh-10:
LOL, I didn't know that we can measure execution time of parts of code. I will think of more ways for async.

Kobzol:
Well, there's no builtin support for it, I just wrapped these two lines with `Instant::now()` etc. :laughing:

Daksh-10:
I thought of some more ways like, fetching info from public resources, sending thousands of requests to dummy servers, but these might have issues of internet speed or shutting down of resources in future. So, probably should I just increase the iterations for async word_count function?? Because it works.

Kobzol:
Yeah, let's just increase the work done within async for now.

Daksh-10:
Updated results
image

Kobzol:
I checked the [code](https://godbolt.org/z/bfW79aGjT), and I'm relatively sure that the `await` is optimized away completely, and the code doesn't even yield into the `tokio` runtime. That happens because the `await`-ed future (`line_char_count`) is executed eagerly, and if it does not block (return `Poll::Pending`), so it will basically execute synchronously.

The current version of the code thus doesn't really benchmark anything inside of the tokio runtime. We probably need a better benchmark, sigh. Benchmarking is hard :)

Daksh-10:
I did thought that `line_char_count` will take very very less time. Hence, executed eagerly (without switching to waiting and then to running state like in the Round Robin Algorithm, if I am thinking correctly) , therefore, executing synchronously. I think we need a function which takes a considerable about of time,so I will try to look up some string manipulation questions :) 

Also working on system level is tough, but interesting. I didn't understand what this [code](https://godbolt.org/z/bfW79aGjT) is, but  I got an overview.

Daksh-10:
image
Here, the stddev for instruction is high. According to me, that is a positive as number of instructions executed at any given time for async depends on  I/O operations resulting in oscillation of waiting and running states.

Kobzol:
Hi, sorry for taking so long, this PR fell through the cracks. Given our experience with very weird benchmark results in this PR, I think that it would be better to add an async benchmark adapted from some real code, rather than such a trivial microbenchmark. Our current runtime benchmark suite is not really prepared for such a quickly running code (this is also a problem for some other benchmarks in the suite), and from what we have seen so far, this benchmark was measuring a lot of things, but not the overhead of `async`.

Daksh-10:
What we can do is one choice from HTTP request to free APIs, I/O operations on file(read and write), some basic function executed asynchronously or all of them.
AsyncRuntimeBenchmark[https://github.com/pkolaczk/async-runtimes-benchmarks] , this repository has async benchmark for different languages, we can use this approach too :)

Issue #1876, opened 2024-03-22
Percent delta from first should be the default


the8472:
Currently both the "graphs" page and the "30 day history" embedded in individual benchmarks display the "Raw" metrics.
Most of the time changes are so small that those views are ~useless without lots of zooming. And even then one has to zoom in on one of several lines.
Defaulting to "Delta from First" would make those views more useful, especially the embedded views since those don't have the dropdown.

Kobzol:
Regarding the embedded graphs:
![image](https://github.com/rust-lang/rustc-perf/assets/4539057/cd1f0a08-4e3e-4c3d-b4e3-a1ded4024a8f)

The idea was to show both a high-level picture of how did the perf. change (left side), and also changes relative to the previous commit (to find out noise spikes).

If I understand this issue correctly, you would like to change the left graph to be "percent from first", to make it easier to gauge the  change from the start of the observed period?

For the graphs displayed above, it would look like this:
![image](https://github.com/rust-lang/rustc-perf/assets/4539057/87153b1b-58b8-4968-b513-13c86cd745ac)

Kobzol:
The right side is essentially a derivative of the raw graph, and the percent from first is essentially a zoomed in version of the raw graph. So this change makes sense to me.

the8472:
> If I understand this issue correctly, you would like to change the left graph to be "percent from first", to make it easier to gauge the change from the start of the observed period?

It's more about the zoom level. The absolute graphs start at a zero base which means the graph covers a huge value range which makes most changes tiny in comparison.

Better default zoom/axis offsets would help too, but since we already have percent-from-first in other places I think just reusing that would be simpler.

the8472:
Why though? Personally I never found those particularly useful. They're essentially flat.
Unless you're comparing full vs. incremental times or something like that I suppose.

Kobzol:
A few reasons:
- It allows us to see the absolute values of e.g. instruction counts, which I find useful.
- This page is probably more often viewed by non-experts (w.r.t. the benchmark detail), so I would show the absolute numbers, since it makes the graphs easier to interpret, in my opinion.
- The fact that the profiles are mostly flat is fine to me, if they are mostly flat, it means that nothing much has changed in the past N days. People tend to overlook axis labels in my experience, and if we showed this:
![image](https://github.com/rust-lang/rustc-perf/assets/4539057/3d176c34-4f7f-4584-825c-627300d96e16)
[...]

PR #1878, opened 2024-03-23
Use "percent from first" graph in benchmark detail

Kobzol:
This PR changes the absolute (left) graph in benchmark detail to use the "percent from first commit" value instead. This should make it easier to observe smaller changes (more discussion in the linked issue).

image

Fixes: https://github.com/rust-lang/rustc-perf/issues/1876


lqd (approved):
Sure, we can try it out and see if noise is easier to weed out this way.

Issue #1880, opened 2024-03-25
Parts of the codebase don't not have a license


RaphaelMarinier:
Parts of the codebase don't not have a license, for instance:
- https://github.com/rust-lang/rustc-perf/blob/master/database
- https://github.com/rust-lang/rustc-perf/tree/master/intern
are not covered by any explicit license.

This makes re-using this project harder, as the code with no license is not open-source (e.g. cannot be legally modified, redistributed, etc..).

An MIT LICENSE file at the root of the project would make everything clearer.

(I know that the bulk of the code which is under site/ and collector/ is already explicitly covered by the MIT license, which is good).

Kobzol:
This is more of an oversight than anything else, we consider the whole repo (apart from code under `collector/compile-benchmarks`) to be MIT-licensed. Let's try to fix it in https://github.com/rust-lang/rustc-perf/pull/1881.

Kobzol:
There is now a top-level MIT license: https://github.com/rust-lang/rustc-perf/pull/1881.

PR #1881, opened 2024-03-25
Add top-level license file

Kobzol:
Fixes: https://github.com/rust-lang/rustc-perf/issues/1880


Kobzol:
@Mark-Simulacrum Could you take a look please? Thanks!

krasimirgg:
+1, now that rustc-perf is a submodule of rust-lang/rust, having no top-level license file is an obstacle to integrate and re-use.

LICENSE.md
```diff
@@ -19,3 +19,5 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+Note that this license doesn't apply to code under the collector/compile-benchmarks directory.
```
Mark-Simulacrum, line 23: I would put this at the top perhaps or we can do a setup similar to rust-lang/rust with the COPYRIGHT file which lists out exceptions.

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
```
Mark-Simulacrum, line 1: I don't think we can just move this file. There's a number of directories without an explicit LICENSE file right now, my sense is that our best bet would be to do a checkbox issue to get approvals and then land this PR, similar to those done for many other rust-lang repos (e.g., https://github.com/rust-lang/rfcs/issues/2076).

LICENSE.md
Mark-Simulacrum: While we're doing that I think if we can just get permission to do MIT OR Apache 2.0 that would also be good (for all directories) rather than just MIT.

Kobzol:
> +1, now that rustc-perf is a submodule of rust-lang/rust, having no top-level license file is an obstacle to integrate and re-use.

@krasimirgg Could you please clarify if this blocks you from using `rust-lang/rust` source distributions, or just integrating with the `opt-dist` PGO/BOLT component?

LICENSE.md
Kobzol: Done in https://github.com/rust-lang/rustc-perf/issues/1933.

LICENSE.md
Kobzol: Moved it to the top.

Mark-Simulacrum (approved):

Issue #1817, opened 2024-02-13
Add runtime benchmark historical charts to the graphs page


Kobzol:
The [graphs page](https://perf.rust-lang.org/index.html) currently only shows compile benchmarks, it would be nice to add also runtime benchmarks to it.

It can be done in multiple ways, e.g. by adding a completely separate page, by adding a new tab, or just by introducing a new category within the same page (there is already a separate category for binary sizes at the bottom). I would prefer either the tab or the category.

chengr4:
I will try to make this

chengr4:
@Kobzol Merry Christmas!

Basically, I plan to proceed with the design outlined in [this reply](https://github.com/rust-lang/rustc-perf/pull/1882#issuecomment-2020682402), but before I begin, I’d like to confirm it with you to ensure we're aligned. 

You said, "maybe with first just adding a single tab for compile-time benchmarks, and not modifying anything else."

I assume that I am a user, seeing only a single `compile-time` tab at the top of the graphs page without any additional functionality might feel a bit odd. Personally, I still lean towards implementing both the `compile-time` and `runtime` tabs at once. When switching to the `runtime` tab, we could simply display 'to be continued' for now. 

An additional benefit of this approach is that the tab-switching functionality can be completed at the same time, avoiding the need to figure out how to implement tab switching later on.

<img width="1830" alt="Image" src="https://github.com/user-attachments/assets/8e9c00e4-1d28-4019-864c-3d6a5f1ae364" />

This is just my personal opinion, and I will directly follow your decision.

Kobzol:
I should have more time for reviews now, and the build system issue should be hopefully resolved now, so feel free to implement both the tab system and the runtime graphs in the same PR :)

PR #1882, opened 2024-03-26
`Tab` component abstraction

chengr4:
For #1817

### DIFF
--- a/site/frontend/src/components/tab.vue
+++ b/site/frontend/src/components/tab.vue
@@ -9,7 +9,7 @@ const props = defineProps<{
 <template>
   <div class="tab" :title="props.extraInfo" :class="{selected: props.selected}">
     <div class="title">{{ props.title }}</div>
-    <div class="summary table-wrapper">
+    <div class="summary">
       <slot name="summary"></slot>
     </div>
   </div>
@@ -59,17 +59,6 @@ const props = defineProps<{
 .selected {
   border-style: solid;
   border-color: black;
-}
-
-.table-wrapper {
-  table {
-    width: 100%;
-    table-layout: auto;
-  }
-
-  th {
-    font-weight: normal;
-  }
 }
 
 .summary {
--- a/site/frontend/src/components/tab.vue
+++ b/site/frontend/src/components/tab.vue
@@ -1,0 +1,82 @@
+<script setup lang="ts">
+const props = defineProps<{
+  extraInfo: string;
+  title: string;
+  selected: boolean;
+}>();
+</script>
+
+<template>
+  <div class="tab" :title="props.extraInfo" :class="{selected: props.selected}">
+    <div class="title">{{ props.title }}</div>
+    <div class="summary table-wrapper">
+      <slot name="summary"></slot>
+    </div>
+  </div>
+</template>
+
+<style scoped lang="scss">
+.tab {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  min-width: 200px;
+  min-height: 60px;
+  padding: 5px;
+  text-align: center;
+  border: 2px dotted #cccccc;
+  border-radius: 5px;
+  cursor: pointer;
+
+  &:hover {
+    @extend .selected;
+  }
+
+  @media (min-width: 600px) {
+    &:not(:first-child) {
+      margin-left: 30px;
+    }
+
+    &:not(:last-child) {
+      :after {
+        content: "";
+        position: absolute;
+        right: -17px;
+        border-right: 1px solid black;
+        top: 20%;
+        bottom: 20%;
+      }
+    }
+  }
+}
+
+.title {
+  font-weight: bold;
+  font-size: 1.1em;
+  margin-bottom: 5px;
+}
+
+.selected {
+  border-style: solid;
+  border-color: black;
+}
+
+.table-wrapper {
+  table {
+    width: 100%;
+    table-layout: auto;
+  }
+
+  th {
+    font-weight: normal;
+  }
+}
+
+.summary {
+  display: flex;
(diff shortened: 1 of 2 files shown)


Kobzol:
Hi, thanks for the PR. I think that we can do this in parts, i.e. in the first PR, just introduce tabs, and add more stuff in later PRs.

I'm not sure how should the tabs work though. Compile-time vs benchmarks optimized for size are completely separate "axes", and it doesn't really make sense to treat them as being on the same level.

I think that we could do sort of two level grouping. On the top level, there could be two tabs (compile-time vs runtime benchmarks), and then inside each tab, there could be several subsets of the data, and some of them could be collapsed. Something like this:
image

Again, I would start by just adding the tabs, maybe with first just adding a single tab for compile-time benchmarks, and not modifying anything else. It would also be nice to reuse the tab component from the compare page, if it could be done without complicating its code too much.

chengr4:
How about also add the runtime tab, and indicate "to be continued" explicitly 🤔?

Kobzol:
I don't think there's a lot of value in that, but if you want, feel free to add it. The point that I was trying to make is to do small PRs that are easy to review, while still keeping the code working in the meantime :)

chengr4:
@Kobzol 
I basically just abstracted `Tab` component.
I will pause here until you agree to proceed. When you have time, please chech it out. Thank you.

I personally find abstracting `Tabs` to be quite challenging. However, if you believe it's necessary, please instruct me. Alternatively, my next step could be to add tabs on the graph page.

site/frontend/src/components/tab.vue
```diff
@@ -0,0 +1,71 @@
+<script setup lang="ts">
+const props = defineProps<{
+  extraInfo: string;
```
Kobzol, line 3: Maybe `tooltip` is more understandable than `extraInfo`.

Kobzol:
I agree that it *could* be unwieldy to generalize the tab component, but I like the current state of the PR - CSS styling and switching of tabs can be kept in the `Tab` component, and just the inner content is replaced with the slot. That seems straightforward enough. And since we will probably use tabs on more places, I think that it's nice to encapsulate it like this, to avoid repeating the styling and selection logic of the tabs.

So :+1: from me.

chengr4:
> And since we will probably use tabs on more places, I think that it's nice to encapsulate it like this, to avoid repeating the styling and selection logic of the tabs.

Challenge accepted 🫠🫠

chengr4:
@Kobzol 
I feel before I keep going, I should share the current state with you.

Basically, the `Tabs` was abstracted. I created an array for creating `TabComponent`  and passed it into `Tabs`. 
Moreover, all calculation and `SummaryTable` were lifted to `compare/page.vue`. I wish I am on the right path.

I was thinking a better way to handle those calculation rather than put them all in the `compare/page.vue`. I also found out we got another `SummaryTable`. Maybe it is another refactoring issue.

 If you have any opinion please let me know 🙏.

Kobzol (commented):
Oh, you also made the tab switcher generic :sweat_smile: I liked that you made the single tab to be a separate component, to avoid duplicating its DOM and CSS. I pretty much wanted to stay at that, and merge the PR once it was cleaned up.

Having the "tab switcher" generic is even better, but it should be really a generic component. Currently, you hardcode the `Tab` type in the tab switcher, and also `compileTimeSummary`, `runtimeSummary` etc. These types shouldn't be known to a generic tab switcher component. The tab switcher should just receive N tabs (Vue components), handle click events on each tab, and then somehow mark the selected tab. I think that a reasonable API for that would be to use something like React render props (maybe there's some equivalent in Vue?), so that the switcher uses a function to render each tab, and passes the function an index that states which tab is selected. Something like this:
```vue
<TabSwitcher tabCount="3" :renderTab="(index: number) => renderTabWithIndex(index)" />
```

The above interface example uses integer indices, to make the tab switcher generic. Maybe there is some TypeScript type-level magic that would allow passing in an iterable type (generic enum?) that would allow selecting the individual tabs using a proper type, rather than just indices, but it could be quite ugly.

If you want to, you can try to see if you can get something done with TypeScript so that we can pass arbitrary types to the tab switcher, but if it doesn't work out, I think that just using indices would be good enough.

chengr4:
Lol
I misunderstood you earlier. Actually, I also preferred the state before I abstracted `TabSwitcher`.

For challenging of generalizing `TabSwitcher`, we can leave it for next time.
I will revert commits and clean up.

But really thank you for the advice and code review

chengr4:
@Kobzol
I reverted commits, please take a look to see if there are any other modifications needed,
when you are available. 

If everything's fine, we can merge this PR. I'll open a new one to create tabs for the graph page.

site/frontend/src/pages/compare/tabs.vue
```diff
@@ -95,36 +98,33 @@ const activeTab: Ref<Tab> = ref(props.initialTab);
...
-    </div>
-    <div
-      class="tab"
-      title="Runtime benchmarks: measure how long does it take to execute (i.e. how fast are) programs compiled by the compared rustc."
-      :class="{selected: activeTab === Tab.Runtime}"
+      </template>
+    </TabComponent>
+    <TabComponent
+      :tooltip="'Runtime benchmarks: measure how long does it take to execute (i.e. how fast are) programs compiled by the compared rustc.'"
+      :title="'Runtime'"
```
Kobzol, line 113:
```suggestion
      title="Runtime"
```
If you don't use `:` before the name of the attribute, it will be evaluated as a string, so you won't need the extra apostrophes. (same for the other `:title` attributes).

Kobzol (approved):
One final nit, otherwise LGTM.

PR #1895, opened 2024-04-20
Add command to compare binary stats of two arbitrary binaries

Kobzol:
The `collector binary_stats` command is quite useful to compare the binary statistics of sections and symbols of two ELF artifacts. Before, it was only possible to compare the binaries produced by some compile benchmark in `rustc-perf`. After this PR, you can use this command as a general introspection tool to compare two arbitrary binaries on your disk.

This should be helpful e.g. to compare the binary size of two Rust toolchain components.


collector/README.md
```diff
@@ -516,23 +516,28 @@ Codegen diff is currently only implemented for runtime benchmarks.
...
+  ./target/release/collector binary_stats compile `<rustc>` --include <benchmark name> --rustc2 <rustc2>
+  ```
+  or between two codegen backends:
+  ```bash
+  ./target/release/collector binary_stats compile `<rustc>` --include <benchmark name> --rustc2 <rustc>
+      --backend <Llvm|Cranelift> --backend2 <Llvm|Cranelift>
+  ```
+- Arbitrary binary artifacts on disk:
+  ```bash
+  ./target/release/collector binary_stats local `<artifact-path>` [<artifact-to-compare-to>]
```
lqd, line 539: Should that mention `--artifact2`?

lqd (commented):
This looks fine to me, but I've never used clap or seen that part of the collector before so I'm not sure if I'm an adequate reviewer.

collector/README.md
Kobzol: It's actually a positional argument, so no flag needs to be passed.

Kobzol:
Ok, thanks for looking anyway!

Mark-Simulacrum (approved):
Seems OK.

collector/README.md
lqd: Then the code mentions `--artifact2` but means "the second positional argument"?

Kobzol: Aahh, fuck, I left the old comment there. Thanks.

collector/README.md
```diff
@@ -516,23 +516,28 @@ Codegen diff is currently only implemented for runtime benchmarks.
...
```
chengr4, line 539: The "artifact" I recognize is the version of `rustc`. I wonder if using "binary artifacts" to refer to other things would cause confusion? What about just telling the user that it is a binary executable or library directly

Kobzol: Well, naming is hard :) Feel free to send a PR that clarifies it to binary/library, if you want.

PR #1907, opened 2024-05-14
Show mean of primary/secondary results in GitHub summary comment tooltip

Kobzol:
The performance summary tables in GH comments (e.g. [this one](https://github.com/rust-lang/rust/pull/124256#issuecomment-2109264134)) show all metrics other than instruction count in a collapsed manner. I think that this sometimes leads to missing interesting information in these tables, if no one clicks on them.

This PR adds a short summary of the performance result to the `Results` reveal button, so that it is clear how large is the effect even without clicking on the button. It should look like this:

image

### DIFF
--- a/site/src/github/comparison_summary.rs
+++ b/site/src/github/comparison_summary.rs
@@ -324,8 +324,25 @@ fn write_metric_summary(
                 write_summary_table(&primary, &secondary, false, message);
             }
             DefaultMetricVisibility::Hidden => {
+                let format_summary =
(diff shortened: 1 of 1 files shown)


nnethercote (approved):
`tooltip` seems like a strange name. To me that means something that shows up when you hover over a GUI element. Maybe call it `summary`? Otherwise, LGTM.

Kobzol:
I also didn't like the name, but couldn't think of a better one. Summary is already kind of overloaded in the code, but it is a good name nevertheless. Changed the PR to use it.

PR #1910, opened 2024-05-22
Update & remove unused npm packages

s7tya:
I updated the packages to reduce npm warnings and removed some unused ones like "vite" and "@types/highcharts".
I'm a bit concerned about Highcharts because we can't test the dashboard page locally. One option might be to revert the last commit and stick with Highcharts 6.x.


s7tya:
Oops, sorry I broke highcharts. Will fix it

Kobzol (approved):
Looks good, thank you! :)

PR #1914, opened 2024-05-29
refactor: hard-code v1 API response

weihanglo:
Another step toward making rustc-perf self-contained.
See <https://github.com/rust-lang/rust/pull/125465>.

It makes no different between depending on an arbitrary Git commit
of rust-lang/team and inlining the responded JSON structure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,7 +1109,6 @@ checksum = "d530e1a18b1cb4c484e6e34556a0d948706958449fca0cab753d649f2bce3d1f"
 dependencies = [
  "equivalent",
  "hashbrown",
- "serde",
 ]
 
 [[package]]
@@ -2088,15 +2087,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "rust_team_data"
-version = "1.0.0"
-source = "git+https://github.com/rust-lang/team#89bb40d3dca1bd9a6c6dbc1b8cb2ddc1fcc932ad"
-dependencies = [
- "indexmap",
- "serde",
-]
-
-[[package]]
 name = "rustc-demangle"
 version = "0.1.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Kobzol:
Hmm. I don't like this at all :laughing: Especially since it's only for a git dependency that is being used by `site`, which isn't required for `rust-lang/rust` at all.

Isn't it possible to vendor a git dependency in Cargo?

weihanglo:
> Isn't it possible to vendor a git dependency in Cargo?

It is indeed supported. The caveats of doing that are:

* It currently depends on both the team API and the Git repository, rather than just the team API. If the commit is gone (unlikely though), it then cannot be built. If there is a breaking change the team API, we're gonna change both. Depending on the Git repository doesn't really save us from breaking changes.
* The Git source needs to be added to the exception source list in rust-lang/rust tidy check (See <https://github.com/rust-lang/rust/pull/125465#issuecomment-2130745904>).

> it's only for a git dependency that is being used by site, which isn't even required for rust-lang/rust.

I am sorry this requires <https://github.com/rust-lang/cargo/issues/12307> 😞.

---

I am fine with either way. We could alternatively add an exception in rust-lang/rust and work on rust-lang/cargo#12307. Feel free to merge or close :)

Kobzol:
> It currently depends on both the team API and the Git repository, rather than just the team API. If the commit is gone (unlikely though), it then cannot be built. If there is a breaking change the team API, we're gonna change both. Depending on the Git repository doesn't really save us from breaking changes.

To be clear, we don't really have issues with updating the team API in this repo, so I'm not too worried about the approach that we use with the git dependency. I'm just interested in getting vendoring working without having to hardcode stuff.

In `rust-lang/rust`, none of the team API code is used, therefore my idea was that if we just vendor the git reference at the time of packaging the source tarball, it should just work, right? We just need it to compile (in any way), it won't be actually used during benchmarking.

Kobzol:
Ok, thanks! That being said, not all git dependencies are really needed. It would be nice to resolve the messy `measureme` situation where we depend on three different versions.

PR #1916, opened 2024-05-31
chore: update to `analyzeme@12.0.0` and `rust_team_data@a5260e7`

weihanglo:
Another step toward making rustc-perf self-contained for <https://github.com/rust-lang/rust/pull/125465>.

This update two dependencies:

* `analyzeme@12.0.0`. Previously we depend on Git repository directly. In this version the v7 profdata file format support is dropped.
* `rust_team_data@a5260e7`. This version includes the license issue fixed (<https://github.com/rust-lang/team/pull/1462>). Since rust_team_data is only used by the website, it should be fine just depending on Git. See <https://github.com/rust-lang/rustc-perf/pull/1914>.

r? @Kobzol

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2090,7 +2090,7 @@ dependencies = [
 [[package]]
 name = "rust_team_data"
 version = "1.0.0"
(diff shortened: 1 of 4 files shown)


Kobzol (approved):
Nice cleanup, thanks!

PR #1919, opened 2024-06-08
Change the filter from includes() to a regex match

### LINKED ISSUE #1900: A way to use regex to filter benchmarks would be nice
Presumably we could have a toggle by the search bar to switch to regex mode.

s7tya:
closes #1900

Changed to use a regex match instead of `includes()`.


site/frontend/src/pages/compare/compile/common.ts
```diff
@@ -160,7 +160,7 @@ export function computeCompileComparisonsWithNonRelevant(
   function shouldShowTestCase(comparison: TestCaseComparison<CompileTestCase>) {
     const name = `${comparison.testCase.benchmark} ${comparison.testCase.profile} ${comparison.testCase.scenario} ${comparison.testCase.backend}`;
     const nameFilter = filter.name && filter.name.trim();
-    const nameFiltered = !nameFilter || name.includes(nameFilter);
+    const nameFiltered = !nameFilter || name.match(new RegExp(nameFilter));
```
Kobzol, line 163: When the input is not a valid regex, we should somehow catch the error, and treat the input as a normal string:
image

Because when this happens, it sadly breaks the page.

Also, it would be nice to treat the types a bit more explicitly here, i.e. compare that the result of match `!== null`, rather than implicitly coercing an array to a bool.

site/frontend/src/pages/compare/runtime/common.ts
```diff
@@ -25,7 +25,8 @@ export function computeRuntimeComparisonsWithNonRelevant(
   function shouldShowTestCase(comparison: TestCaseComparison<RuntimeTestCase>) {
     const name = comparison.testCase.benchmark;
     const nameFilter = filter.name && filter.name.trim();
-    return !nameFilter || name.includes(nameFilter);
+    const nameFiltered = !nameFilter || name.match(new RegExp(nameFilter));
```
Kobzol, line 28: This logic will be slightly more complicated (see the other comment), so it would be nice to unify the logic in some shared function, something like `filterBenchmarkName(benchmarkName, filterName)` and put it e.g. in `site/src/pages/compare/shared.ts`.

Kobzol (commented):
Thanks for the PR! Left a few comments.

s7tya:
I noticed that currently it's not really filtering the "benchmark name" because the variable `name` is set to `${comparison.testCase.benchmark} ${comparison.testCase.profile} ${comparison.testCase.scenario} ${comparison.testCase.backend}`.

It should just filter the names? I thought it's little bit weird to filter invisible string by regex.

site/frontend/src/pages/compare/shared.ts
```diff
@@ -78,3 +78,18 @@ export function getBoolOrDefault(
   }
   return defaultValue;
 }
+
+export function isBenchmarkNotFilteredByName(
```
Kobzol, line 82: This name is a bit.. unintuitive :laughing: What about something like `benchmarkNameMatchesFilter`?

site/frontend/src/pages/compare/shared.ts
```diff
@@ -78,3 +78,18 @@ export function getBoolOrDefault(
...
+  benchmarkName: string,
+  filterName: string
+): boolean {
+  if (!filterName) return true;
```
Kobzol, line 86: We should trim `filterName` at the start of the function, otherwise a space will be parsed as a regex, but it won't match anything.

Kobzol (commented):
> It should just filter the names instead? I thought it's little bit weird to filter invisible(i mean on the UI) string by regex.

That's a good point! It's a historical leftover, but I think that with the current filters that we have, and the regex, we should just use the name alone.

Left two nits, once you take a look at them, we can ship this!

site/frontend/src/pages/compare/shared.ts
s7tya: I added this early return because `filterName` is somehow nullable. Although it is not specified as such in the type, I encountered an error where `filterName.trim()` was undefined because `filterName` itself was undefined.

Kobzol: Then that's probably a bug somewhere in the code that calls this function. Could you try to investigate where does this `undefined` come from?

s7tya: ahh i noticed that the project is `strict: false`, so I'll change filterName from string to string | null.

Kobzol: Ah, I didn't notice that `null` is actually a valid value for the filter. Well, we could probably remove that and just use an empty string to mark a missing value, but let's keep it like it is, modify the function parameter to `string | null`, and handle the null case.

site/frontend/src/pages/compare/shared.ts
```diff
@@ -78,3 +78,18 @@ export function getBoolOrDefault(
   }
   return defaultValue;
 }
+
+export function benchmarkNameMatchesFilter(
+  benchmarkName: string,
+  filterName: string | null
+): boolean {
+  if (!filterName) return true;
```
Kobzol, line 86: Now this should be changed to something like this:
```typescript
if (filterName === null) return true;
const trimmedFilterName = filterName.trim();
if (trimmedFilterName === "") return true;
```
So that a name filter containing only whitespace will match all benchmarks.

s7tya: Of course I can add it explicitly, it's already doing same thing:
(empty regex can match everything)

CleanShot 2024-06-09 at 21 34 19@2x

Kobzol (approved):
Thanks! :)

PR #1922, opened 2024-06-10
Add runtime detail graphs api

### LINKED ISSUE #1818: Add runtime benchmark historical chart to runtime benchmark detail
The runtime benchmark table on the [compage page](https://perf.rust-lang.org/compare.html) currently doesn't show any detail nor historical charts, which makes it difficult to analyze the results. It would be nice to add a similar benchmark detail as is present for compile time benchmarks to it:
image

Apart from the charts, the detail could also contain a command for profiling the runtime benchmark localy using the `rustc-perf` suite.

s7tya:
closes #1818


s7tya:
Thank you for the review! Fixed them.

Kobzol (approved):
Awesome :tada: Thank you!

PR #1929, opened 2024-06-23
Fix styling of benchmark detail graphs

Zentrik:
Not sure if there was a way to fix this without duplicating the styles, but this works (see https://tealquaternion.camdvr.org/compare.html for an example)
This broke in https://github.com/rust-lang/rustc-perf/pull/1922.

### DIFF
--- a/site/frontend/src/pages/compare/benchmark-detail.scss
+++ b/site/frontend/src/pages/compare/benchmark-detail.scss
@@ -1,0 +1,39 @@
+.columns {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 15px;
+  margin: 10px 0;
+
+  .grow {
+    flex-grow: 1;
+  }
+
+  &.graphs {
+    flex-wrap: nowrap;
+  }
+}
+
+.graphs {
+  margin-top: 15px;
+}
+
+.rows {
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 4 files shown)


Kobzol:
Oh, this is annoying, by moving the code to a new component, the CSS from the parent component was lost. It is interesting that the CSS still applies to the root element of the graph detail (`.columns` was still applied), but it does not apply to its children (`.rows` did not apply), probably that's just how Vue works.

Thanks for the fix! I'm not very fond of the duplicated styles though :( Could you please take `.columns, .graphs, .rows and .title` and put it e.g. into `src/pages/compare/benchmark-detail.scss`, and then import it from `benchmark-detail.vue`, the compile graph component and the runtime graph component (https://stackoverflow.com/a/60576240/1107768)? I tried it locally and it seems to work. Something like [this](https://github.com/rust-lang/rustc-perf/compare/master...Kobzol:rustc-perf:shared-css-test).

We should probably eventually deduplicate the compile and runtime graph components, as they are very similar, but that should be left for a future PR.

Kobzol (approved):
Thanks! Could you please squash the commits? I'll merge it after. Let me know if you want help with the squashing.

Kobzol:
Sorry, I either missed or didn't get the notification about your latest commit :facepalm: Thank you!

PR #1931, opened 2024-06-26
Improve symbol normalization in `binary_stats`

Kobzol:
A few improvements and tests added when I was investigating https://github.com/rust-lang/rust/pull/126965#issuecomment-2191101137.

### DIFF
--- a/collector/src/artifact_stats.rs
+++ b/collector/src/artifact_stats.rs
@@ -102,16 +102,44 @@ impl ArtifactStats {
     }
 }
 
-/// Tries to match hashes produces by rustc in mangled symbol names.
-static RUSTC_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
-
 /// Demangle the symbol and remove rustc mangling hashes.
+///
+/// Normalizes the following things, in the following order:
+/// - Demangles the symbol.
+/// - Removes `.cold` and `.warm` from the end of the symbol, to merge cold and hot parts of a function
+/// into the same symbol.
+/// - Removes rustc hashes from the symbol, e.g. `foo::[abcdef]` -> `foo::[]` or
+/// `foo::abcd` -> `foo`.
+/// - Removes suffixes after a dot from the symbol, e.g. `anon.abcdef.123` -> `anon` or
+/// `foo.llvm.123` -> `foo`.
+///
+/// These modifications should remove things added by LLVM in the LTO/PGO phase.
+/// See more information here: https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html#vendor-specific-suffix
 fn normalize_symbol_name(symbol: &str) -> String {
-    let regex =
-        RUSTC_HASH_REGEX.get_or_init(|| Regex::new(r"(::)?\b[a-z0-9]{15,17}\b(\.\d+)?").unwrap());
+    /// Tries to match hashes in brackets produced by rustc in mangled symbol names.
+    static RUSTC_BRACKET_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
+    /// Tries to match hashes without brackets after :: produced by rustc in mangled symbol names.
+    static RUSTC_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
+    /// Tries to match suffixes after a dot.
+    static DOT_SUFFIX_REGEX: OnceLock<Regex> = OnceLock::new();
 
-    let symbol = rustc_demangle::demangle(symbol).to_string();
-    regex.replace_all(&symbol, "").to_string()
+    let bracket_hash_regex =
+        RUSTC_BRACKET_HASH_REGEX.get_or_init(|| Regex::new(r"\[[a-z0-9]{13,17}\]").unwrap());
+    let hash_regex = RUSTC_HASH_REGEX.get_or_init(|| Regex::new(r"::[a-z0-9]{15,17}$").unwrap());
+    let dot_suffix_regex = DOT_SUFFIX_REGEX.get_or_init(|| Regex::new(r"\.[a-z0-9]+\b").unwrap());
+
+    let mut symbol = rustc_demangle::demangle(symbol).to_string();
+
+    if let Some(stripped) = symbol.strip_suffix(".cold") {
+        symbol = stripped.to_string();
+    }
+    if let Some(stripped) = symbol.strip_suffix(".warm") {
+        symbol = stripped.to_string();
+    }
+    let symbol = bracket_hash_regex.replace_all(&symbol, "");
+    let symbol = hash_regex.replace_all(&symbol, "");
+    let symbol = dot_suffix_regex.replace_all(&symbol, "");
+    symbol.to_string()
 }
 
 /// Simple heuristic that tries to normalize section names.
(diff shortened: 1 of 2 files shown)


collector/src/artifact_stats.rs
```diff
@@ -102,16 +102,41 @@ impl ArtifactStats {
...
-static RUSTC_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
-
 /// Demangle the symbol and remove rustc mangling hashes.
+///
+/// Normalizes the following things, in the following order:
+/// - Demangles the symbol.
+/// - Removes `.cold` and `.warm` from the end of the symbol, to merge cold and hot parts of a function
+/// into the same symbol.
+/// - Removes rustc hashes from the symbol, e.g. `foo::[abcdef]` -> `foo::[]` or
+/// `foo::abcd` -> `foo`.
```
lqd, line 112: I'm not sure I understand this  `foo::abcd` -> `foo` transform. What does `normalize_short_hash` look like without it?

lqd (commented):
(Oooops I didn't see this, sorry.)

I'm not sure we can be fully accurate and without conflating symbols to be perfect in all cases. Depending on the tasks, sometimes in the past I've had to compare using mangled symbols because the monomorphizations were important, e.g. when using const generics.

Here we're best effort so these are improvements to me.

collector/src/artifact_stats.rs
```diff
@@ -102,16 +102,41 @@ impl ArtifactStats {
...
+/// - Removes suffixes after a dot from the symbol, e.g. `anon.abcdef.123` -> `anon` or
+/// `foo.llvm.123` -> `foo`.
 fn normalize_symbol_name(symbol: &str) -> String {
-    let regex =
-        RUSTC_HASH_REGEX.get_or_init(|| Regex::new(r#"(::)?\b[a-z0-9]{15,17}\b(\.\d+)?"#).unwrap());
+    /// Tries to match hashes in brackets produces by rustc in mangled symbol names.
+    static RUSTC_BRACKET_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
```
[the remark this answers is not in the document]
Kobzol: Definitely, but we currently have a MSRV of 1.75.0, I would wait a few months before bumping it to 1.80.0.

collector/src/artifact_stats.rs
```diff
@@ -102,16 +102,41 @@ impl ArtifactStats {
...
 /// Demangle the symbol and remove rustc mangling hashes.
+///
+/// Normalizes the following things, in the following order:
+/// - Demangles the symbol.
+/// - Removes `.cold` and `.warm` from the end of the symbol, to merge cold and hot parts of a function
+/// into the same symbol.
+/// - Removes rustc hashes from the symbol, e.g. `foo::[abcdef]` -> `foo::[]` or
+/// `foo::abcd` -> `foo`.
+/// - Removes suffixes after a dot from the symbol, e.g. `anon.abcdef.123` -> `anon` or
+/// `foo.llvm.123` -> `foo`.
```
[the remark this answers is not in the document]
Kobzol: Oh, this has something to do with LTO? I had no idea :D It's not just LTO though, right? E.g. `.cold` and `.hot` is probably just function splitting, maybe even based on PGO/BOLT profiles.

collector/src/artifact_stats.rs
Kobzol: Yeah, this is tricky, because `abcd` can be a normal name, or a hash. The regex tries to remove hashes, but keep normal names. So `std::fs::File` is obviously ok, but `std::fs::File::abcde231312fa` is not. But what is and isn't a hash isn't really trivial to figure out.

collector/src/artifact_stats.rs
```diff
@@ -229,3 +254,60 @@ pub fn compile_and_get_stats(
...
+        assert_eq!(
+            normalize_symbol_name(
+                "_RNvNtCsifRNxopDi_20rustc_builtin_macros6format16make_format_args"
+            ),
+            "rustc_builtin_macros::format::make_format_args".to_string()
+        );
+    }
+
+    #[test]
+    fn normalize_hash_without_brackets() {
```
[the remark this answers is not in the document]
Kobzol: I tried to refactor and comment the tests to make them more clear. Let me know what you think.

collector/src/artifact_stats.rs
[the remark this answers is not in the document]
lqd: Yeah I'd assume most of them are coming from profile-driven transforms, but there's been historically some function splitting passes that can also use static analysis to mark regions. They can rely on assertions and intrinsics, diverging paths, manual `#[cold]` attributes, and language features that make use of all of the above (eg I think our drop paths were marked cold and non-inlinable, things like that.).

collector/src/artifact_stats.rs
lqd: Then maybe we should punt on this specific normalization step for later, if it's hard to avoid such false positives.

collector/src/artifact_stats.rs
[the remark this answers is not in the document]
lqd: (I had also added a couple examples in the v0 mangling rfc https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html#vendor-specific-suffix)

collector/src/artifact_stats.rs
Kobzol: I wouldn't say that it has caused issues so far. Right now the regex is `::[a-z0-9]{15,17}$`. Having a such a long identifier (without any underscore or hyphen) at the end of the demangled name is probably very uncommon.

collector/src/artifact_stats.rs
```diff
@@ -229,3 +254,80 @@ pub fn compile_and_get_stats(
...
+        check(
+            "_RNvNtCsifRNxopDi_20rustc_builtin_macros6format16make_format_args",
+            "rustc_builtin_macros[e293f6447c7da]::format::make_format_args",
+            "rustc_builtin_macros::format::make_format_args",
+        );
+    }
+
+    #[test]
+    fn normalize_hash_without_brackets() {
+        // Check that hashes withouto brackets are removed.
```
lqd, line 316:
```suggestion
        // Check that hashes without brackets are removed.
```

collector/src/artifact_stats.rs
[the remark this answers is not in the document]
lqd: Looks great.

collector/src/artifact_stats.rs
lqd: I don't know if they always appear in specific constructs, e.g. for closures or something. If that's the case maybe we should normalize the whole and not the parts. 

Writing this also makes me realize that eg ".cold" and ".warm" are handled differently than the numeric suffix in ".llvm.2132123131". But really all these are the same should likely be normalized on the mangled symbols as vendor specific suffixes after a dot.

Kobzol: I remember that there were some weird things with the dot, and just removing everything after a dot didn't work I think, but I don't remember the specifics :( Indeed it is weird that the numeric suffix is removed just by demangling, but not the `.cold` annotations.

lqd: Agreed. It could be a legacy/v0 thing, or a missing feature in the demangler.

collector/src/artifact_stats.rs
Kobzol: Thanks, fixed.

collector/src/artifact_stats.rs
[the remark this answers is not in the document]
Kobzol: Added a comment about LTO/PGO.

collector/src/artifact_stats.rs
```diff
@@ -102,16 +102,44 @@ impl ArtifactStats {
...
+/// `foo.llvm.123` -> `foo`.
+///
+/// These modifications should remove things added by LLVM in the LTO/PGO phase.
+/// See more information here: https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html#vendor-specific-suffix
 fn normalize_symbol_name(symbol: &str) -> String {
-    let regex =
-        RUSTC_HASH_REGEX.get_or_init(|| Regex::new(r"(::)?\b[a-z0-9]{15,17}\b(\.\d+)?").unwrap());
+    /// Tries to match hashes in brackets produces by rustc in mangled symbol names.
+    static RUSTC_BRACKET_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
+    /// Tries to match hashes without brackets after :: produces by rustc in mangled symbol names.
```
lqd, line 121: I guess

```suggestion
    /// Tries to match hashes in brackets produced by rustc in mangled symbol names.
    static RUSTC_BRACKET_HASH_REGEX: OnceLock<Regex> = OnceLock::new();
    /// Tries to match hashes without brackets after :: produced by rustc in mangled symbol names.
```

or "that rustc produces"?

lqd (approved):
one last typo fix.

and if we don't want to do anything yet about the pre-demangling dot because of the other issue, then we should be good

collector/src/artifact_stats.rs
Kobzol: It was supposed to be "produced". Thanks!

PR #1935, opened 2024-07-03
Add more visible warning when a benchmark fails to compile in a PR

Kobzol:
This should make the warning more visible, to avoid situations where we overlook it (e.g. https://github.com/rust-lang/rust/pull/125507). This is how it should look like:
image


nnethercote (approved):
I'll approve because it's a clear improvement, but the "no action needed" heading isn't true when there is a compile error...

Kobzol:
Good point, I also modified the header.

nnethercote (approved):

PR #1936, opened 2024-07-03
Update `deeply-nested-multi` to increase type length limit

Kobzol:
Needed because of changes done in https://github.com/rust-lang/rust/pull/125507. See [this conversation](https://github.com/rust-lang/rust/pull/125507#issuecomment-2206794614).


lqd (approved):

lqd:
I think between this change and the PR causing the regressions, the graph has now become super hard to read, having to show such a wide range. For example https://perf.rust-lang.org/index.html?start=2024-05-03&end=83217ae97e94eb43ce31396b49a938608f7f9c06&benchmark=deeply-nested-multi&profile=opt&scenario=incr-unchanged&stat=instructions%3Au&kind=raw

Kobzol:
It only affects this one benchmark, and it will resolve itself in ~30 days. Do you think that some changes to graph handling are warranted or we can just wait?

PR #1940, opened 2024-07-06
use "bootstrap" instead of "rustbuild"

onur-ozkan:
Let's stick with the single name "bootstrap" to refer to the bootstrap project to avoid confusion.

cc https://github.com/rust-lang/rust/pull/127434


Kobzol:
Hi, this code is under the compile benchmarks directory. We don't modify or hotfix that code, it's essentially read only, unless we need to make it compile :) If you need to make a change to cargo, it should be done upstream.

onur-ozkan:
> Hi, this code is under the compile benchmarks directory. We don't modify or hotfix that code, it's essentially read only, unless we need to make it compile :) If you need to make a change to cargo, it should be done upstream.

Hi, sure, https://github.com/rust-lang/cargo/pull/14207.

PR #1941, opened 2024-07-08
fix alert display when build fails

klensy:
cc #1935

See https://github.com/rust-lang/rust/pull/127473#issuecomment-2213091171, second alert signs right shifted, while instead they should be aligned with previous alert signs.

second \n before `alert_row` needed or markdown will render this as appended to last list item.


Kobzol (approved):
Markdown is tricky :) Thank you!

PR #1944, opened 2024-07-11
insert only summaries with specified profile when specified

### LINKED ISSUE #1943: Empty summary graphs are shown when selecting profile on graphs page
While looking at #1938, I notice that there's empty summary graphs when we select specific profile on graphs page. It's not high priority as I think there's not many people facing this issue, it would be nice to be fixed.

https://perf.rust-lang.org/index.html?profile=opt
image

s7tya:
fix #1943

I added a profile param to the `create_summary()`. This lets us filter the summary based on the specified profile. If no profile is specified, it will behave as before. If a profile is specified, it will only include that profile in the summary.

### DIFF
--- a/site/src/request_handlers/graph.rs
+++ b/site/src/request_handlers/graph.rs
@@ -235,7 +235,13 @@ async fn create_graphs(
         .collect();
 
     if request.benchmark.is_none() {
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thanks! :)

PR #1945, opened 2024-07-11
Add runtime chart to dashboard

### LINKED ISSUE #1816: Add runtime benchmark historical chart to the dashboard
The dashboard currently contains data for `check`/`debug`/`opt`/`doc` builds. It would be nice to also add a historical chart for our runtime benchmarks to it (probably with the wall-time metric, same as for the compile benchmarks).

s7tya:
fix #1816 

There isn't a perf-config.json for the runtime benchmarks, so I averaged all the runtime benchmarks. However, it would be somewhat useful to add one.

### DIFF
--- a/site/frontend/src/pages/dashboard.ts
+++ b/site/frontend/src/pages/dashboard.ts
@@ -3,7 +3,7 @@ import {DASHBOARD_DATA_URL} from "../urls";
 
 import {getJson} from "../utils/requests";
 
-interface DashboardCases {
+interface DashboardCompileBenchmarkCases {
   clean_averages: [number];
   base_incr_averages: [number];
   clean_incr_averages: [number];
@@ -13,10 +13,11 @@ interface DashboardCases {
 interface DashboardResponse {
   Ok: {
     versions: [string];
-    check: DashboardCases;
-    debug: DashboardCases;
-    opt: DashboardCases;
-    doc: DashboardCases;
+    check: DashboardCompileBenchmarkCases;
+    debug: DashboardCompileBenchmarkCases;
+    opt: DashboardCompileBenchmarkCases;
+    doc: DashboardCompileBenchmarkCases;
+    runtime: [number];
   };
 }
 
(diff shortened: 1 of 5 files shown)


site/frontend/src/pages/dashboard.ts
```diff
@@ -78,12 +79,44 @@ function render(
...
+  Highcharts.chart({
+    chart: {
+      renderTo: element,
+      zooming: {
+        type: "xy",
+      },
+      type: "line",
+    },
+    title: {
+      text: `Average time for a run`,
```
Kobzol, line 92:
```suggestion
      text: `Average time for a runtime benchmark`,
```

site/frontend/templates/pages/dashboard.html
```diff
@@ -1,21 +1,36 @@
...
-The dashboard shows performance results for all stable Rust releases going back to
-<code>1.28.0</code>, along with the latest <code>beta</code> release. The displayed
-duration is an arithmetic mean amongst all
-<a href="https://github.com/rust-lang/rustc-perf/tree/master/collector/compile-benchmarks#stable">stable</a>
-benchmarks.
+    The dashboard shows performance results for all stable Rust releases going back to
+    <code>1.28.0</code>, along with the latest <code>beta</code> release. The displayed
+    duration is an arithmetic mean amongst all
+    <a href="https://github.com/rust-lang/rustc-perf/tree/master/collector/compile-benchmarks#stable">stable</a>
+    benchmarks.
```
Kobzol, line 22: We could mention the runtime benchmarks too. Something like "The dashboard also shows the average duration of runtime benchmarks, which measure the performance of Rust programs compiled by a given version of the Rust compiler.".

Kobzol (commented):
Thanks! Left two comments.

Kobzol (approved):
Thanks! Let's see how this looks on the live site.

Having the dashboard for runtime benchmarks might be a bit weird, because we don't have any stable runtime benchmarks, but visualizing something is better than not seeing anything, I suppose. If the data will become too weird, we can always hide it.

PR #1946, opened 2024-07-12
bump deps, resolving cargo audit issues

klensy:
see individual commits

`chrono` api is complex, hope this works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,11 +273,13 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.79"
+version = "1.0.99"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -66,6 +66,12 @@ dependencies = [
  "serde",
 ]
 
+[[package]]
+name = "android-tzdata"
```
Kobzol, line 70: This doesn't seem very useful for us :) But it's enabled by default by `chrono`. We'd need to manually enable features to get rid of it.

Kobzol (approved):
Thanks!

Cargo.lock
klensy: It's from clock feature from `chrono` https://github.com/chronotope/chrono/blob/v0.4.38/Cargo.toml#L26, which is selected by `postgres-types` https://github.com/sfackler/rust-postgres/blob/647a9256b80ef961b8065678f41745ca040202b5/postgres-types/Cargo.toml#L39-L41

PR #1947, opened 2024-07-13
remove interpolated points of dashboard's runtime bench and fix graph unit

s7tya:
I fixed the graph unit from seconds to nanoseconds and removed interpolated points from when runtime benchmarks hadn't been implemented yet.

### DIFF
--- a/site/frontend/src/pages/dashboard.ts
+++ b/site/frontend/src/pages/dashboard.ts
@@ -80,6 +80,13 @@ function render(
 }
 
 function renderRuntime(element: string, data: [number], versions: [string]) {
+  // Remove null and convert nanoseconds to miliseconds
+  // The null values, which indicate that the runtime data is missing, are only present at the beginning of the array.
+  const formattedData = data
+    .filter((data) => data != null)
+    .map((data) => data / 1_000_000);
+  const nullCount = data.length - formattedData.length;
+
   Highcharts.chart({
     chart: {
       renderTo: element,
@@ -92,11 +99,11 @@ function renderRuntime(element: string, data: [number], versions: [string]) {
       text: `Average time for a runtime benchmark`,
     },
     yAxis: {
-      title: {text: "Seconds"},
+      title: {text: "Miliseconds"},
       min: 0,
     },
     xAxis: {
-      categories: versions,
+      categories: versions.slice(nullCount),
       title: {text: "Version"},
     },
     series: [
@@ -104,7 +111,7 @@ function renderRuntime(element: string, data: [number], versions: [string]) {
         showInLegend: false,
         type: "line",
         animation: false,
-        data,
+        data: formattedData,
       },
     ],
   });
--- a/site/src/request_handlers/dashboard.rs
+++ b/site/src/request_handlers/dashboard.rs
@@ -137,13 +137,26 @@ pub async fn handle_dashboard(ctxt: Arc<SiteCtxt>) -> ServerResult<dashboard::Re
     let responses = ctxt
         .statistic_series(runtime_benchmark_query.clone(), aids.clone())
         .await?;
+
+    // The flag is used to ignore only the initial values where the runtime benchmark was not implemented.
+    let mut ignore_runtime_benchmark = true;
     let points = db::average(
         responses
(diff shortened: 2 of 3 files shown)


site/frontend/src/pages/dashboard.ts
```diff
@@ -92,7 +92,7 @@ function renderRuntime(element: string, data: [number], versions: [string]) {
       text: `Average time for a runtime benchmark`,
     },
     yAxis: {
-      title: {text: "Seconds"},
+      title: {text: "Nanoseconds"},
```
Kobzol, line 95: I would actually prefer to just show a larger unit (milliseconds?), and then multiply the values, rather than showing these super big numbers.

site/src/request_handlers/dashboard.rs
```diff
@@ -153,6 +155,8 @@ pub async fn handle_dashboard(ctxt: Arc<SiteCtxt>) -> ServerResult<dashboard::Re
                 ArtifactId::Commit(c) => format!("master: {}", &c.sha.to_string()[0..8]),
                 ArtifactId::Tag(aid) => aid.clone(),
             })
+            .collect::<Vec<_>>()
```
Kobzol, line 158: Correct me if I'm wrong, but this would also affect the compile time benchmarks, and we definitely don't want to do that, because that would essentially skip compile time versions for which we do not have runtime data.

I think it would be better to return e.g. something like `Vec<Option<f64>>` for the runtime data, and then let the frontend skip the compiler versions for which there is no data available.

Kobzol (commented):
Thank you for the PR, I left some comments. Are you able to test this locally, btw? (you can do that by manually modifying the results.db file to insert results that will be shown in the dashboard, or modify the querying code so that it also returns non-stable artifacts). Without testing, it will be quite challenging to get this right.

s7tya:
Thanks for the heads-up. I've been testing them locally, but only the data related to the runtime graphs were showing up, so I didn't catch my mistake. I just modified a few artifact names to "1.xx.x" and changed the type to "release." It only worked for the runtime data.

Kobzol:
Hmm, if the `versions` array contained some compiler versions, then your changes were probably fine, but the question is whether there was some compile benchmark data in your local DB. In any case, it would be great to simulate the situation where you have e.g. 5 compile-time artifacts, but only e.g. 3 runtime artifacts, to see if the missing runtime data is being ignored.

s7tya:
(I just updated the and the change may fix the problem, i'm still investigating why my local artifacts doesn't have compile benchmark)

site/frontend/src/pages/dashboard.ts
```diff
@@ -80,6 +80,12 @@ function render(
 }
 
 function renderRuntime(element: string, data: [number], versions: [string]) {
+  // Remove null and convert nanoseconds to miliseconds
+  const formattedData = data
+    .filter((data) => data != null)
+    .map((data) => data / 1_000_000);
+  const nullCount = data.length - formattedData.length;
```
Kobzol, line 87: This assumes that nulls be only at the beginning, but that should be fine in our case. It would be nice to add a comment to explain this invariant here though.

site/src/request_handlers/dashboard.rs
```diff
@@ -137,13 +137,25 @@ pub async fn handle_dashboard(ctxt: Arc<SiteCtxt>) -> ServerResult<dashboard::Re
     let responses = ctxt
         .statistic_series(runtime_benchmark_query.clone(), aids.clone())
         .await?;
+
+    let mut ignore_runtime_benchmark = true;
```
Kobzol, line 141: This bool is needed if there is some interpolated data *after* the first runtime benchmark result?

As above, please add a comment here that explains why do we need this handling (because historically, we don't have runtime benchmark data for old artifacts).

Kobzol (commented):
Thank you! Left two comments that ask you to add two comments :laughing:

Kobzol (approved):
Thanks! Given that this is non-trivial to test locally, let's see how it will look like on the live site.

s7tya:
Thanks!

PR #1948, opened 2024-07-14
drop protobuf and derive_more from deps

klensy:
First commit: `prometheus` currently uses plain text, so disable `protobuf`; this reduces `site` bin by ~ 1mb;
Second: drops `derive_more`

Bumped stable rust version from 1.70 to 1.73 for CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.70.0
+          toolchain: 1.73.0
           override: true
           components: rustfmt, clippy
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,17 +567,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derive_more"
-version = "0.99.17"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.70.0
+          toolchain: 1.79.0
```
Kobzol, line 26: This one is pretty fresh, it would be nice if `rustc-perf` compiled at least with a few older versions of the compiler.

klensy:
Bumped stable version for CI from 1.70 to current 1.79, is it ok?

.github/workflows/ci.yml
klensy: To which version should i drop?

Kobzol: Ideally the oldest one that still compiles :) Looks like `1.73.0` should be enough.

klensy: Well, locally i can build as old as 1.78, because 1.77 will give
`
link: extra operand 'rustc-perf\\target\\debug\\build\\windows_x86_64_msvc-f220d21a5bc470a5\\build_script_build-f220d21a5bc470a5.build_script_build.d44e75c843d0cc87-cgu.0.rcgu.o'
`

klensy: Bumped to 1.73

Kobzol (approved):
Thanks!

PR #1949, opened 2024-07-17
Add perf triage for 2024-07-16

rylev:


triage/2024-07-16.md
```diff
@@ -0,0 +1,107 @@
+# 2024-07-16 Triage Log
+
+Fairly quite week with the only pure regressions being small and coming from correctness fixes. The biggest single change came from turning off the `-Zenforce-type-length-limit` check which had positive impacts across many different benchmarks since the compiler is doing strictly less work.
```
Kobzol, line 3:
```suggestion
Fairly quiet week with the only pure regressions being small and coming from correctness fixes. The biggest single change came from turning off the `-Zenforce-type-length-limit` check which had positive impacts across many different benchmarks since the compiler is doing strictly less work.
```

Kobzol: Also maybe we could mention that the check has been really enabled like two weeks ago, so this won't be seen as an actual improvement to most people (?).

Kobzol (commented):
Left one nit, otherwise LGTM.

triage/2024-07-16.md
rylev: TWiR has already been released so I'm not sure the change in wording will really make a difference.

Kobzol (approved):

Issue #1939, opened 2024-07-05
Add REUSE annotations for benchmarks in `collector/compile-benchmarks`


pietroalbini:
The code in `collector/compile-benchmarks` is now a submodule of `rust-lang/rust`, and included in source tarballs we distribute. We thus should annotate its licensing information so that it shows up in [the upcoming REUSE-powered licensing information](https://github.com/rust-lang/rust/issues/99414).

Once `rust-lang/rust` [upgrades to REUSE 4.x](https://github.com/rust-lang/rust/issues/127361) (which adds the new `REUSE.toml`), it should be possible to add a `REUSE.toml` in any sub-directory of `collector/compile-benchmark`, with a content like this:

```toml
version = 1

[[annotations]]
path = ["**"]
SPDX-FileCopyrightText = "Authorship / copyright line"
SPDX-License-Identifier = "SPDX-EXPRESSION"
```

Kobzol:
Now that REUSE 4.0 is used in rustc, we can proceed with this. Would you prefer to put a `REUSE.toml` file into each compile-time benchmark directory, or have a top-level one in `collector/compile-benchmarks` with entries for all the benchmarks? I'd probably slightly prefer the latter, but no strong opinion.

pietroalbini:
Both work, I don't have a strong opinion either.

PR #1951, opened 2024-07-23
Use Reuse 4 to manage licenses

Kobzol:
This PR introduces `REUSE.toml` files for managing licenses and copyright in this repository, including all the compile-time benchmarks under the `collector/compile-benchmarks` directory. More context can be found in https://github.com/rust-lang/rustc-perf/issues/1939.

For the existing compile-time benchmarks, I used this "algorithm":
1) If it was an artificial stress test created by rustc developers, I used the copyright `The Rust Project Developers (see https://thanks.rust-lang.org)` and the MIT license. Note that we couldn't use `MIT OR Apache-2.0`, because the `compile-benchmarks` directory was only licensed under MIT previously.
2) If it was a third-party crate, I took its licenses (they were present for all the crates, luckily). As for copyright, if there was an explicit copyright in the README, I used it (this happened I think only for one crate). Otherwise, I tried to look for "XYZ developers" or something like that in the `authors` field of the crate's `Cargo.toml`. If I haven't found that, I just used `<crate-name> contributors`.

Fixes: https://github.com/rust-lang/rustc-perf/issues/1939


Kobzol:
I'm not sure if it's fine to land https://github.com/rust-lang/rustc-perf/pull/1951/commits/6a2cba87d97540be354712ba25d38cdd4957ee67, I don't know if e.g. GitHub recognizes `REUSE.toml`.

collector/compile-benchmarks/README.md
```diff
@@ -206,6 +206,13 @@ Rust code being written today.
       --id Test --profiles=Check --scenarios=IncrPatched
       --include=$NEW_BENCHMARK`
   - Add the new entry to `collector/compile-benchmarks/README.md`.
+  - Add a new licensing entry to `collector/compile-benchmarks/REUSE.toml` (see existing entries
+  for inspiration).
+    - If the benchmark is artificial, use the `MIT` license and set Rust developers as the
+    copyright owners (see e.g. `await-call-tree` as an example).
```
fbstj, line 212: I didn't know if you'd want to ask for new benchmarks to be licensed the "normal" for rustc `MIT or Apache-2.0`?

Kobzol: That's a very good point! And we can do that now, thanks to granular REUSE licenses. Thanks for the suggestion, updated the instructions.

fbstj: does the `await-call-tree` example still make sense?

Kobzol: Yes, because the "Rust Project" developers is quite long, I didn't want to include the whole thing here. It should be copy-pasted from an existing example.

fbstj: yes but that example is only under the `MIT` license so I didn't know if it made more sense to point to a `MIT or Apache-2.0` licensed example (though I also understand pointing at the first rust-project example in the file for just the text)

Kobzol: Yeah, now that example is only relevant because of the copyright, not the license. But I think that should be fine.

collector/compile-benchmarks/tt-muncher/REUSE.toml
```diff
@@ -0,0 +1,18 @@
+version = 1
```
pietroalbini, line 1: I think it's fine to inline this into the parent `REUSE.toml`.

pietroalbini (approved):
Looks good to me! Left one optional nit.

collector/compile-benchmarks/tt-muncher/REUSE.toml
Kobzol: Done! Thanks.

Kobzol:
Thanks! Let's see what happens on the repo's GH homepage once the old licenses are removed.

PR #1952, opened 2024-07-24
Update the RDS root CA list

MarcoIeni:
Fix issue https://github.com/rust-lang/infra-team/issues/105
Similar to https://github.com/rust-lang/triagebot/pull/1772

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -22,10 +22,10 @@ impl Postgres {
     }
 }
 
-const CERT_URL: &str = "https://s3.amazonaws.com/rds-downloads/rds-ca-2019-root.pem";
+const CERT_URL: &str = "https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem";
 
 lazy_static::lazy_static! {
-    static ref CERTIFICATE_PEM: Vec<u8> = {
+    static ref CERTIFICATE_PEMS: Vec<u8> = {
         let client = reqwest::blocking::Client::new();
         let resp = client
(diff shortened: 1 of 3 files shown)


database/src/pool/postgres.rs
```diff
@@ -1349,3 +1358,15 @@ fn parse_artifact_id(ty: &str, sha: &str, date: Option<DateTime<Utc>>) -> Artifa
...
+
+#[cfg(test)]
+mod tests {
+    use super::make_certificates;
+
+    // Makes sure we successfully parse the RDS certificates and load them into native-tls compatible
+    // format.
+    #[test]
+    fn can_make_certificates() {
+        make_certificates();
```
Kobzol, line 1370: I know that you took this from triagebot, but can we also just do a small assert that the resulting certificate chain is not empty? Thanks!

MarcoIeni: Sure! Done 👍

Kobzol:
Thanks! Let's try to deploy this and see if it works. Worst case we can just revert it.

Kobzol (approved):

MarcoIeni:
Thanks, let me know once you confirm it's working 👍

Kobzol:
The web interface seems to work fine, data from the DB is appearing. Let's see if collector can also store data there (will get results in ~1.5 hour).

Kobzol:
Actually, it also stores data in the meantime, and the data has been updated at https://perf.rust-lang.org/status.html. So it works.

MarcoIeni:
Great, thanks!

PR #1953, opened 2024-07-25
fix issue number in test`s package name

klensy:


Kobzol (approved):
Renaming the package shouldn't affect the DB or benchmarks. Thanks!

Issue #1734, opened 2023-10-12
Cli tool to compare commits locally


Nadrieril:
Hi! I tend to work on performance-sensitive parts of the compiler so I check performance locally a lot. In the current setup this requires me to run a docker script and check a website page. It would make my life much easier if I could run say `rustc-perf compare <commita> <commitb> instructions:u` and get out the little table of relevant benchmark diffs. Would that be easy to implement?

Kobzol:
Hi, you don't really have to run Docker, you can either just build the website locally (using `npm run watch && cargo build -p site`) or you can download a precompiled binary with the website (https://github.com/rust-lang/rustc-perf/releases).

That being said, a basic CLI output wouldn't be a bad idea, we already have it for runtime benchmarks.

It should be relatively easy to implement, either by modifying the behavior of `bench_local`, and adding output of the measured results, or by introducing a separate new command for this. The `collector` code was heavily refactored and running benchmarks from it in a new command should hopefully be relatively easy.

s7tya:
I think we can close this because the `bench_cmp` command is implemented and it's ported into the bootstrap. Of course we should improve the functionality though.

Kobzol:
Agreed :+1:

PR #1955, opened 2024-07-29
Create `bench_local_diff` command

s7tya:
I've implemented a small example command for comparing two artifacts. Currently, it only displays the range, mean, and count across all series. This is the first step towards #1734. I'm not sure if the way I'm collecting data is correct or not.

(Notice that the count column is different from the one shown on the website. It counts every point in the database, not just the benchmarks.)

```
$ cargo run --bin fetch_latest results.db
$ cargo +nightly run --release --bin collector bench_local_diff --a_id 28f9060f96b97141dab0e267d633b408d5ec20c1 --b_id d40d6149dad
661aaf4e42bac19925ed97448cdc9

+--------+-------+------------------+--------+
| name   | count | range            | mean   |
+--------+-------+------------------+--------+
| ❌     | 3048  | [-0.50%, -0.00%] | -1.30% |
+--------+-------+------------------+--------+
| ✅     | 3598  | [+0.00%, +0.73%] | +1.43% |
+--------+-------+------------------+--------+
| ✅, ❌ | 7930  | [-0.50%, +0.73%] | +0.15% |
+--------+-------+------------------+--------+
```


collector/src/bin/collector.rs
```diff
@@ -628,6 +630,18 @@ enum Commands {
         #[command(flatten)]
         db: DbOption,
     },
+
+    /// Displays diff between two local bench result.
+    BenchLocalDiff {
```
Kobzol, line 635: Local should not be needed, as we're just comparing whatever is in the database. I would probably call this `BenchCmp`, `collector bench_cmp` sounds intuitive, as we want to compare two artifacts. The same name is also used for the compare page on the web.

collector/src/bin/collector.rs
```diff
@@ -628,6 +630,18 @@ enum Commands {
...
+        #[command(flatten)]
+        db: DbOption,
+
+        #[arg(long)]
+        a_id: String,
```
Kobzol, line 640: We want to have some notion of a base and a modified artifact, so I would call these `base` and `modified`.
Ideally, it the arguments should be long, but also positional, so that it is possible to do `bench_cmp a b` or `bench_cmp --base a -- modified b`.

collector/src/bin/collector.rs
```diff
@@ -1187,6 +1201,105 @@ Make sure to modify `{dir}/perf-config.json` if the category/artifact don't matc
             println!("Data of artifact {name} were removed");
             Ok(0)
         }
+        Commands::BenchLocalDiff { db, a_id, b_id } => {
+            let pool = Pool::open(&db.db);
```
Kobzol, line 1205: Please move this already to e.g. `src/compare.rs`. This logic will probably grow much bigger than this in time :) We can keep the DB connection here, and send the DB connection and the two artifact names to a function that will do the comparison.

Kobzol (commented):
Thank you for working on this! :) I would like to avoid one big PR that will implement some super-duper UI at once, so let's first land the basic CLI interface, and just output a single useful piece of information (which compiler is faster on instruction counts on average, and by how much). Then you can improve the UI incrementally piece by piece.

Kobzol:
To clarify: even though I suggested just to output a simple text line for simplicity, you can keep the table that you have right now, if you want. Although I suggest that soon it will be replaced by a more complex UI anyway.

collector/src/lib.rs
```diff
@@ -376,3 +378,23 @@ impl CollectorCtx {
...
+    builder
+        .worker_threads(1)
+        .max_blocking_threads(1)
+        .enable_time()
+        .enable_io();
+    builder.build().expect("built runtime")
+}
+
+#[derive(Debug, clap::Args)]
+pub struct DbOption {
```
Kobzol, line 394: Sorry, I probably wasn't clear. This is a CLI option, this should stay in `bin/collector`. What I meant was to create a function in `compare.rs` that would essentially look like this:
```rust
async fn compare_artifacts(conn: Box<dyn Connection>, base: String, modified: String)
```

s7tya:
I've fixed the requested changes.

collector/src/bin/collector.rs
```diff
@@ -628,6 +629,18 @@ enum Commands {
         #[command(flatten)]
         db: DbOption,
     },
+
+    /// Displays diff between two local bench result.
+    BenchCmp {
+        #[command(flatten)]
+        db: DbOption,
+
+        #[arg(long)]
```
Kobzol, line 638: Could you please add doc comments to `base` and `modified`? So that it shows in the CLI help.

collector/src/bin/collector.rs
```diff
@@ -628,6 +629,18 @@ enum Commands {
...
+        base: String,
+
+        #[arg(long)]
```
Kobzol, line 641: It looks like `clap` doesn't support an argument to be both positional and long at once :( I probably mistook it with `typer`/`click` from Python. So let's remove the `long` `--` form, it would be annoying, the default usage should be `bench_cmp a b`.

collector/src/compare.rs
```diff
@@ -0,0 +1,110 @@
+use database::{Connection, Lookup};
+use itertools::Itertools;
+use tabled::{Table, Tabled};
+
+const DEFAULT_SIGNIFICANCE_THRESHOLD: f64 = 0.002;
+
+/// Compare 2 artifacts and print the result.
+pub async fn compare_artifacts(mut conn: Box<dyn Connection>, base: String, modified: String) {
+    let index = database::Index::load(&mut *conn).await;
+    let sids = index
```
Kobzol, line 10: Could you please filter only the results for instruction counts? To avoid mixing the metrics, as we have discussed. Thanks!

collector/src/compare.rs
```diff
@@ -0,0 +1,110 @@
...
+        .compile_statistic_descriptions()
+        .map(|(_, sid)| sid)
+        .collect::<Vec<_>>();
+
+    let base_id_number = conn
+        .artifact_by_name(&base)
+        .await
+        .expect("Cannot find specified artifact")
+        .lookup(&index)
+        .unwrap();
```
Kobzol, line 20: Let's return `anyhow::Result<()>` from the function and propagate errors, to avoid the `unwrap`s everywhere.

collector/src/compare.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    let pstats = conn
+        .get_pstats(&sids, &[Some(base_id_number), Some(modified_id_number)])
+        .await;
+    let tuple_pstats = pstats
+        .into_iter()
+        .map(|row| row.into_iter().collect_tuple::<(_, _)>().unwrap())
```
Kobzol, line 33: Could you just implement this by hand? The code uses `unwrap` anyway, so it would crash if there weren't exactly two elements. Introducing a new dependency just for this feels like overkill.

collector/src/compare.rs
```diff
@@ -0,0 +1,110 @@
...
+    impl From<&Vec<f64>> for Regression {
+        fn from(value: &Vec<f64>) -> Self {
+            let min = *value.iter().min_by(|a, b| a.total_cmp(b)).unwrap();
+            let max = *value.iter().max_by(|a, b| a.total_cmp(b)).unwrap();
+            let count = value.len();
+
+            Regression {
+                range: (min, max),
+                count,
+                mean: (value.iter().sum::<f64>() / count as f64),
```
Kobzol, line 62: Let's check the case where `count` is zero.

collector/src/compare.rs
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+    }
+
+    let change = tuple_pstats
+        .iter()
+        .filter_map(|&(a, b)| match (a, b) {
+            (Some(a), Some(b)) => {
+                if a == 0.0 {
```
Kobzol, line 71: Comparing floats with zero is.. tricky :laughing: But I suppose in this case we can just let it be.

Kobzol (commented):
It's really cool to be able to just run a CLI command and instantly see the difference :) This is gonna be very useful. Left a few more comments.

s7tya:
I marked this PR as a draft because we need to replace raw get_pstats fn to the selector which we brought to the db crate in the recent PR.

collector/src/bin/collector.rs
```diff
@@ -628,6 +629,18 @@ enum Commands {
         #[command(flatten)]
         db: DbOption,
     },
+
+    /// Displays diff between two local bench result.
```
Kobzol, line 633:
```suggestion
    /// Displays diff between two local bench results.
```

site/src/comparison.rs
```diff
@@ -6,6 +6,7 @@ use crate::api;
 use crate::github;
 use crate::load::SiteCtxt;
 
+use collector::compare::DEFAULT_SIGNIFICANCE_THRESHOLD;
```
Kobzol, line 9: Let's keep this in the site. It's unclear how to handle statistical computation in the comparison CLI. Because there we won't have almost any historical records, and probably the user wants to see all data.

So let's just keep the statistical stuff in `site` for now, and just show *all* the data in the collector CLI.

Kobzol (approved):
Awesome. Tried it locally and it works great. Thank you!

PR #1957, opened 2024-08-01
Do not return empty commit list when a parent commit is missing

Kobzol:
This can happen during force pushes to rust-lang/rust. If a parent commit is missing, just run the benchmark, the result won't have any parent available.


Kobzol:
Asked more people for a review, whoever gets to this first to unblock the queue :)

lqd (approved):
Even if there won’t be a parent we’ll still run both toolchains, right?

Kobzol:
If there isn't a parent, the benchmark should run, but the compare page for that run will probably error out (hopefully nothing more sinister will happen).

PR #1958, opened 2024-08-03
Move `Selector` into database crate

s7tya:
To implement #1734, it would be nice to move `Selector` from the site crate to the database crate.
This PR will move Selector itself and related things into the database crate.

I left things like SiteCtxt that is only used in the site crate as it is.


database/src/lib.rs
```diff
@@ -245,7 +245,7 @@ impl fmt::Display for Profile {
 ///
 /// These are usually reported to users in a "flipped" way. For example,
 /// `Cache::Empty` means we're doing a "full" build. We present this to users as "full".
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, serde::Serialize)]
```
Kobzol, line 252: I think that the `serde::Serialize` part is also unused. You can remove it in another PR if you remember.

site/src/db.rs
```diff
@@ -1,61 +1,2 @@
-use std::fmt;
-
 pub use crate::average::average;
```
Kobzol, line 1: Let's delete the `db` file, it's useless now.

database/src/lib.rs
```diff
@@ -9,6 +9,7 @@ use std::ops::{Add, Sub};
 use std::sync::Arc;
 use std::time::Duration;
 
+pub mod comparison;
```
Kobzol, line 12: Let's call it `metric`, since it only contains the metric enum right now :)

.github/workflows/ci.yml
```diff
@@ -109,14 +109,12 @@ jobs:
...
-        PROFILES: [
-          "Check,Debug,Doc",
-          "Opt",
-        ]
+        BENCH_INCLUDE_EXCLUDE_OPTS:
+          [
+            "--include cargo-0.60.0,stm32f4-0.14.0,webrender-2022",
+            "--exclude cargo-0.60.0,stm32f4-0.14.0,webrender-2022",
+          ]
+        PROFILES: ["Check,Debug,Doc", "Opt"]
```
Kobzol, line 117: `["Check,Debug,Doc","Opt"]` :)

Kobzol (commented):
Thanks! LGTM, apart from a few small nits. Would you like to also remove `async_trait` in this PR?

.github/workflows/ci.yml
s7tya: My editor automatically formats them, so I just remove the diff! Thanks.

Kobzol (approved):
Looks good, thanks! This should help us use queries from collector nicely.

PR #1959, opened 2024-08-05
Use inline script instead actions-rs

s7tya:
The repository for [`actions-rs/toolchain`](https://github.com/actions-rs/toolchain) has been archived and is no longer maintained. Instead, this PR uses an inline script like:

```yaml
      - name: Install stable toolchain
        run: |
          rustup install $RUST_TOOLCHAIN_VERSION --profile minimal
          rustup default $RUST_TOOLCHAIN_VERSION
          rustup component add --toolchain $RUST_TOOLCHAIN_VERSION rustfmt clippy
        env:
          RUST_TOOLCHAIN_VERSION: stable
```

Additionally, I've updated the versions for `actions/checkout`, `actions/setup-node`, and `Swatinem/rust-cache`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,12 @@ jobs:
           fetch-depth: 1
 
       - name: Install toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: 1.75.0
-          override: true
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
Thank you! :)

PR #1961, opened 2024-08-09
Update dependencies

Kobzol:
There are five packages in the workspace of this repository, and they have a lot of dependencies. Now that `rustc-perf` can be used as a part of bootstrap, it would be nice if it was a bit leaner and faster to compile.

This PR removes several unneeded dependencies and updates some other dependencies to reduce the total number of crates compiled. For the `collector` binary, this PR gets its dependency count down from 300 to 294 (I think that more improvements are possible by doing more major dependency updates, but this PR already does a lot, so let's keep that for another PR).


lqd (approved):

Kobzol:
Ok, let's see what happens.

PR #1963, opened 2024-08-14
Improve compare cmd UI

s7tya:
I've added the ability to select artifacts and metrics interactively. Now, an interactive prompt will start if the artifacts or metric isn't explicitly specified. I'm curious to know how many people are using a non-instructions:u metric. If the number is significant, we could consider defaulting to instructions:u and disabling the prompt to streamline the user experience.

### DIFF
--- a/collector/Cargo.toml
+++ b/collector/Cargo.toml
@@ -42,6 +42,7 @@ tabled = { version = "0.14.0", features = ["ansi-str"] }
 humansize = "2.1.3"
 regex = "1.7.1"
 analyzeme = "12.0.0"
+inquire = "0.7.5"
 
 benchlib = { path = "benchlib" }
 
--- a/collector/src/compare.rs
+++ b/collector/src/compare.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{str::FromStr, sync::Arc};
 
 use database::{
     metric::Metric,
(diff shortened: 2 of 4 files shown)


collector/Cargo.toml
```diff
@@ -44,6 +44,7 @@ regex = "1.7.1"
 analyzeme = "12.0.0"
 
 benchlib = { path = "benchlib" }
```
Kobzol, line 46: Could you please add this above this? I tried to visually separate local and remote dependencies (database and intern are moved as well in https://github.com/rust-lang/rustc-perf/pull/1961).

collector/src/compare.rs
```diff
@@ -10,11 +10,29 @@ use tabled::{Table, Tabled};
...
-    base: String,
-    modified: String,
+    base: Option<String>,
+    modified: Option<String>,
 ) -> anyhow::Result<()> {
     let index = database::Index::load(&mut *conn).await;
 
+    let aids = index.artifacts().collect::<Vec<_>>();
+    if aids.len() < 2 {
+        return Err(anyhow::anyhow!("There's no artifacts to compare."));
```
Kobzol, line 63: This can be misleading if there's only a single artifact. Let's say "There are not enough artifacts to compare, at least two are needded".

collector/src/compare.rs
```diff
@@ -10,11 +10,29 @@ use tabled::{Table, Tabled};
...
+            inquire::Select::new(
+                &format!("Choose {} artifact to compare:", name),
+                aids.clone(),
+            )
+            .prompt()?,
+        )
+    };
+
+    let base = base.unwrap_or(select_artifact_id("base")?.to_string());
+    let modified = modified.unwrap_or(select_artifact_id("modified")?.to_string());
```
Kobzol: It would be nice to remove the selected `base` artifact from the list offered in the select, so that you don't accidentally select the same artifact twice.

collector/src/compare.rs
```diff
@@ -7,14 +7,57 @@ use database::{
 };
 use tabled::{Table, Tabled};
 
+static ALL_METRICS: &[Metric] = &[
+    Metric::AssemblyFileSize,
```
Kobzol, line 11: If we're not defaulting instructions, then I'd put them at the top of this list at least. The most useful metrics are instruction counts, cycles, walltime, max RSS and binary size.

Kobzol (commented):
Pretty nice, thanks! I think that we should definitely default into instructions. However, with `inquire`, it's probably quite non-trivial to "go back" and display the metric selector once you have already shown the result. That's something that leads into a more interactive dialog and ratatui. So for now we can keep the metric dialog.

s7tya:
Thank you for the review!

collector/src/compare.rs
```diff
@@ -7,16 +7,87 @@ use database::{
...
+            )
+            .prompt()?,
+        )
+    };
+
+    let base = match base {
+        Some(v) => v,
+        None => select_artifact_id("base", &aids)?.to_string(),
+    };
+    if let Some(idx) = aids.iter().position(|id| id == &base) {
```
Kobzol, line 82: You can also use `aids.retain(|id| id != &base)`, seems a bit more straightforward to understand.

collector/src/compare.rs
```diff
@@ -7,16 +7,87 @@ use database::{
...
+        aids.remove(idx);
+    }
+    let modified = match modified {
```
Kobzol, line 85: I have realized now that the common case locally will be that you have exactly two artifacts. If after filtering out `base` you have only one artifact left, we don't need to display the dialog and we can just select the only other remaining artifact.

Kobzol (commented):
Got one more idea, left a comment.

collector/src/compare.rs
s7tya: Wow that's beautiful rusty code, I'll use it! Thanks.

Kobzol (approved):
Thanks!

PR #1964, opened 2024-08-14
apply showRawData option for markdown export

s7tya:
On Zulip, I saw [some people have expressed interest in adding a showRawData option for Markdown export](https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/Accumulating.20multiple.20reports.20into.20one/near/459515623), so this PR propose to add it.
I also created a `toMarkdownRow` function to reduce the chances of generating invalid Markdown, as the current implementation is missing the final `|` at the end of each row.


Kobzol (approved):
Looks great, thank you!

PR #1967, opened 2024-08-22
dynamically use perfetto

s7tya:
Run `npm run install-perfetto` before running website.
Then, access to the link inside of the compare page detail. ("query trace" button)

### DIFF
--- a/site/frontend/src/perfetto.ts
+++ b/site/frontend/src/perfetto.ts
@@ -39,3 +39,9 @@ function openTrace(arrayBuffer: ArrayBuffer, title: string) {
 
   window.addEventListener("message", onMessageHandler);
 }
+
+export async function checkIsEmbeddedPerfettoEnabled(): Promise<boolean> {
+  const result = await fetch(`/perfetto/index.html`);
+
+  return result.status === 200;
+}
--- a/site/.gitignore
+++ b/site/.gitignore
@@ -3,3 +3,4 @@
 frontend/node_modules
 frontend/dist/**/
 frontend/.parcel-cache
+frontend/static/perfetto
--- a/site/frontend/download_perfetto.sh
+++ b/site/frontend/download_perfetto.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+commit_hash="c74251226a8caa0b43377902ee06d2570faa0c15"
+git clone https://github.com/google/perfetto.git perfetto_tmp
+
+cd perfetto_tmp
+git checkout $commit_hash || exit 1
+
+function allow_url_in_csp() {
+    url=$1
+    sed -i.bak "1,/^[[:space:]]*'blob:'/s/^\([[:space:]]*\)'blob:'/\1'$(echo "$url" | sed 's/\//\\\//g')',\n\1'blob:'/" ui/src/frontend/index.ts
+}
+
+allow_url_in_csp 'https://perf.rust-lang.org'
+allow_url_in_csp 'http://localhost:2346'
+
+patch -u ui/src/frontend/trace_url_handler.ts < ../perfetto_trace_url_handler.patch
+
+tools/install-build-deps --ui
+ui/build
+
+cd ..
+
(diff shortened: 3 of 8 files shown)


site/frontend/src/perfetto.ts
```diff
@@ -39,3 +39,9 @@ function openTrace(arrayBuffer: ArrayBuffer, title: string) {
 
   window.addEventListener("message", onMessageHandler);
 }
+
+export const checkIsEmbeddedPerfettoEnabled = async (): Promise<boolean> => {
```
Kobzol, line 43: This is a bit head-scratching, could you just make an async function please?
```typescript
export async function checkIsEmbeddedPerfettoEnabled(): Promise<boolean> {
   const status = await fetch(`/perfetto/index.html`);
   return status === 200;
} 
```

s7tya: I could do that, but in modern web frontend development, I think const arrow functions are preferred for several reasons. Which do you prefer?

Kobzol: I mean, I'm not primarily a web developer, so I don't follow all the trends :laughing: Arrow functions are surely useful for inline callbacks and they can be useful for object methods because they have a more reasonable behavior w.r.t. `this`. But this seems like a perfectly normal function, so I don't see any reason why it should be an arrow function.

Most importantly, everywhere else in the code we use normal functions for such helper code, so it would be inconsistent :laughing:

s7tya: Sorry for the late. I was bit swamped last week. Alright, I'll rewrite it.

s7tya:
@LalitMaganti
Hello, sorry to bother you.

I’m currently working on embedding Perfetto's UI into perf.rust-lang.org, and it’s almost working perfectly. However, the page keeps reloading every few seconds. The only modification I made was patching the CSP URLs, so I’m unsure why this issue is happening.

Would you happen to have any idea what might be causing this?

LalitMaganti:
Never heard someone run into something like that before, that seems very strange.

I'd suggest using the SHA c4577456a1be9d41853015cd194850f972383db7 corresponding to the previous-stable version of the UI. Unlikely that will fix the issue but it will reduce one variable.

Does Devtools give any hint (log messages which might indicate the originator of the reload)? I don't think there's any code in the UI which would try and reload (modulo version updates and plugin reloads) but there may be some adverse interaction with how you're embedding the UI.

s7tya:
Thank you for the reply! It occurs only if this flag is true.
 
https://github.com/google/perfetto/blob/56b5e6fdd77f3d86ad60d402197d91bda911b1f4/ui/src/frontend/trace_url_handler.ts#L218

 I'll also paste my dev tools console. Though I've added some debug info.

CleanShot 2024-09-15 at 19 18 58

s7tya:
Also I'll switch to the stable commit you gave. Thanks!

LalitMaganti:
Can we just patch that line to return false? That's a bit of an edge case in the fact we use localhost to post files to the UI from the command line. Likely we could fix it upstream but would need some investigation and I don't want to block you on that. 

We can open a Perfetto bug upstream to track the root cause.

s7tya:
Alright. I'll file an issue on upstream and patch that line for now.

s7tya:
It should work fine now. I'll migrate the details page to Vue.js later and update the "Open in Perfetto" link on that page so it functions the same way as the one on the compare details page.

Kobzol (approved):
We might as well just use a forked version of perfetto to avoid this monkey patching :laughing: But let's keep it as it is, if we need to make more changes in the future, we can switch to a fork.

Kobzol (approved):
Thank you! :)

For onlookers: this PR adds the ability to inline the Perfetto trace visualizer on the website, but it is opt-in, Perfetto needs to be downloaded offline first. It takes too long to build to do it on CI, but I will try to build it manually (once) on the server.

Issue #1896, opened 2024-04-20
Store self-profiles in local execution


Kobzol:
Currently, the self-profiles from executing compile benchmarks can be uploaded to S3, which is used in the production deployment. However, it would be nice if we could also store them locally, primarily for examining them and testing GUI features that depend on self-profiles. To do that, it would be great to consolidate the logic of uploading and downloading self-profiles in both `collector` and `site` in some shared logic/trait, and then implement two versions - one that uploads/downloads from S3 (to be used in production), and one that just stores the self-profile data in some directory on disk.

Kobzol:
This was implemented in https://github.com/rust-lang/rustc-perf/pull/2375.

PR #1968, opened 2024-08-23
use brotli to compress self profile response

s7tya:
alternative to #1966
This PR is using brotli to compress and consider Accept-Encoding header according to previous PR's discussion.

### DIFF
--- a/site/src/request_handlers/self_profile.rs
+++ b/site/src/request_handlers/self_profile.rs
@@ -3,6 +3,7 @@ use std::io::Read;
 use std::sync::Arc;
 use std::time::Instant;
 
+use brotli::enc::BrotliEncoderParams;
 use bytes::Buf;
 use database::selector;
 use database::ArtifactId;
@@ -14,11 +15,13 @@ use crate::api::self_profile::ArtifactSizeDelta;
 use crate::api::{self_profile, self_profile_processed, self_profile_raw, ServerResult};
 use crate::load::SiteCtxt;
 use crate::self_profile::{get_or_download_self_profile, get_self_profile_raw_data};
+use crate::server::maybe_compressed_response;
 use crate::server::{Response, ResponseHeaders};
 
 pub async fn handle_self_profile_processed_download(
     body: self_profile_processed::Request,
     ctxt: &SiteCtxt,
+    allow_compression: bool,
 ) -> http::Response<hyper::Body> {
     log::info!("handle_self_profile_processed_download({:?})", body);
     let mut params = body.params.clone();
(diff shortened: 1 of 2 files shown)


s7tya:
It will take about 22s to load both directory giving gzip into the perfetto, and using brotli & content-encoding: brotli by my eyes & stopwatch.
However, it would be nice to use brotli than gzip if it's available, so I propose this than the previous one.

s7tya:
(Current production RLO takes about 200s to load. so this makes x10 faster. )

Kobzol:
Because we don't have https://github.com/rust-lang/rustc-perf/issues/1896 yet, it is not easy to fully test (and benchmark) this locally. So I guess we can try it on the live server and see how fast it is in practice.

site/src/server.rs
```diff
@@ -440,8 +440,18 @@ async fn serve_req(server: Server, req: Request) -> Result<Response, ServerError
         }
         "/perf/processed-self-profile" => {
             let ctxt: Arc<SiteCtxt> = server.ctxt.read().as_ref().unwrap().clone();
+            let support_brotli = req
```
Kobzol, line 443: There is already `allow_compression` on line 342, let's reuse it. This one has a bug; it should check `ACCEPT_ENCODING`, not `CONTENT_ENCODING` :) Splitting it by commas is nicer than just the `contains` check in `allow_compression`, but I think that in that case we should check if the trimmed string starts with `br`, because there can be some [compression options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#syntax).

site/src/self_profile.rs
```diff
@@ -37,9 +37,11 @@ pub fn generate(
         ProcessorType::Crox => {
             let opt = serde_json::from_str(&serde_json::to_string(&params).unwrap())
                 .context("crox opts")?;
+            let data = crox::generate(self_profile_data, opt).context("crox")?;
```
Kobzol, line 40: Let's revert this change to avoid an unnecessary git diff.

site/src/request_handlers/self_profile.rs
```diff
@@ -132,6 +136,19 @@ pub async fn handle_self_profile_processed_download(
         })
         .status(StatusCode::OK);
 
+    if output.filename.ends_with("json") && support_brotli {
```
Kobzol, line 139: You could reuse `maybe_compressed_response` from the `server.rs` file. We just to check if the compression of the profiles is still good enough (i.e. that the resulting file is small enough) with the parameters configured in `server.rs`.

Kobzol:
Looks good! :) Was `5` the best trade-off for the compression options that you found? :)

s7tya:
I just took benchmark and changed to 4.

```
207.22 MB → 19.55 MB (size: 9.43%, quality: 2, time: 713.505083ms)
207.22 MB → 13.85 MB (size: 6.69%, quality: 3, time: 744.224166ms)
207.22 MB → 11.42 MB (size: 5.51%, quality: 4, time: 738.653667ms)
207.22 MB → 12.88 MB (size: 6.22%, quality: 5, time: 1.467692333s)
```

Kobzol (approved):
Thanks for doing the benchmark! The comment in `server.rs` said that `4` might be too slow, maybe on the production server it will be a bit worse, but I'm fine with starting with `4`, it looks like it has a good trade-off locally! It is surprising that `5` produces a larger output than `4` though.

Thanks for working on this :heart:

PR #1971, opened 2024-09-02
Make loading of PEM certificates async

Kobzol:
Running the blocking client inside async context was causing issues with Tokio (when the blocking client was dropped, it was sometimes causing panics in tokio), and it was a hack anyway. This also allows us to remove a feature from `reqwest` and also remove the `lazy_static` dependency.


database/src/pool/postgres.rs
```diff
@@ -75,11 +65,28 @@ async fn make_client(db_url: &str) -> anyhow::Result<tokio_postgres::Client> {
         Ok(db_client)
     }
 }
-fn make_certificates() -> Vec<Certificate> {
+async fn make_certificates() -> Vec<Certificate> {
     use x509_cert::der::pem::LineEnding;
     use x509_cert::der::EncodePem;
 
-    let certs = x509_cert::Certificate::load_pem_chain(&CERTIFICATE_PEMS[..]).unwrap();
+    static CERTIFICATE_PEMS: LazyLock<Mutex<Option<Vec<u8>>>> = LazyLock::new(|| Mutex::new(None));
```
Mark-Simulacrum, line 72: Do we actually need the LazyLock? https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.const_new

Mark-Simulacrum (approved):

database/src/pool/postgres.rs
Kobzol: Interesting. Didn't find this originally, because this requires the `parking_lot` feature. But enabling it does not seem to increase our dep tree, so why not.

PR #1973, opened 2024-09-05
disable bootstrap stage0 target checks

lqd:
Should fix current bootstrap measurement failure

r? @kobzol


Kobzol (approved):
Thank you! :)

PR #1975, opened 2024-09-08
collector bench_cmp: multiply normalized change by 100

### LINKED ISSUE #1974: collector bench_cmp subcommand incorrectly computes percentages
The collector's bench_cmp subcommand, invoked via `./target/release/collector bench_cmp ...`, fails to multiply the normalized change in compilation stats (e.g., compilation time) by 100 to obtain a _percent_ change. As a result, the percent changes reported by `bench_cmp` are much, much smaller than they really are.

The bug is [on this line](https://github.com/rust-lang/rustc-perf/blob/89bc0604d126e3e92a8254c20215ce21ffc6faa3/collector/src/compare.rs#L168), which computes `(b - a) / a` without multiplying by 100.

nmosier:
Fix #1974.

### DIFF
--- a/collector/src/compare.rs
+++ b/collector/src/compare.rs
@@ -165,7 +165,7 @@ pub async fn compare_artifacts(
                 if a == 0.0 {
                     None
                 } else {
-                    Some((b - a) / a)
+                    Some((b - a) / a * 100.0)
                 }
             }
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thank you! :)

PR #1984, opened 2024-10-04
Switch self profile to use HW counters instead of walltime (attempt 2)

Kobzol:
This is a second attempt to land https://github.com/rust-lang/rustc-perf/pull/1647, which was later reverted in https://github.com/rust-lang/rustc-perf/pull/1700.

I tried to manually SSH into the collector server and run the whole benchmark suite with `--self-profile` using icounts. It went through. I'd like to try this on the live server to see what happens. If it fails again, we'll know more to see where we should investigate the failure.

### DIFF
--- a/site/src/request_handlers/self_profile.rs
+++ b/site/src/request_handlers/self_profile.rs
@@ -460,7 +460,7 @@ pub async fn handle_self_profile(
         .benchmark(selector::Selector::One(bench_name.to_string()))
         .profile(selector::Selector::One(profile.parse().unwrap()))
         .scenario(selector::Selector::One(scenario))
-        .metric(selector::Selector::One(Metric::CpuClock));
+        .metric(selector::Selector::One(Metric::InstructionsUser));
 
     // Helper for finding an `ArtifactId` based on a commit sha
     let find_aid = |commit: &str| {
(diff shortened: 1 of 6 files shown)


site/src/request_handlers/self_profile.rs
```diff
@@ -526,7 +526,7 @@ pub async fn handle_self_profile(
         .benchmark(selector::Selector::One(bench_name.to_string()))
         .profile(selector::Selector::One(profile.parse().unwrap()))
         .scenario(selector::Selector::One(scenario))
-        .metric(selector::Selector::One(Metric::CpuClock));
+        .metric(selector::Selector::One(Metric::InstructionsUser));
```
Mark-Simulacrum, line 529: This will break all historical queries for self-profile data, right? Do we want to perhaps query for both or some other mechanism to not do that?

Kobzol: An alternative could be to first refactor self profile management (https://github.com/rust-lang/rustc-perf/issues/1896), so that we can e.g. version it, and then have some logic to detect if the archive has the old (walltime) or new (icount) version, and perform the query base on that.

But I'm not sure if people really look at old self-profile results at all, usually we're only interested in them for a brief moment after/before a PR is merged.

PR #1987, opened 2024-10-08
Avoid showing gray rectangle when width is negative

### LINKED ISSUE #1983: Odd gray region to right of graph after zooming in
To reproduce go to https://perf.rust-lang.org/compare.html and zoom in on any graph. (I've tested on Firefox and Edge on Linux)
Image

I couldn't figure out where in the code this was appearing from.

s7tya:
fixes #1983

This happens because the boundary indicating 'out of the specified range' is displayed even when the calculated width is negative.


Zentrik (approved):
LGTM

Kobzol (approved):
Thank you! :)

PR #1988, opened 2024-10-08
Expand when only one graph is selected

### LINKED ISSUE #1938: Improve rendering of history graphs
When you open the [graph history](https://perf.rust-lang.org/index.html) page with a single graph selected (common when coming from the compage page), it would be nice to show the graph over the full page, rather than just in a small box.

Also, some improvements could be made to improve the readability of the graph when it contains a wide range of values (https://github.com/rust-lang/rustc-perf/pull/1936#issuecomment-2208390792).

s7tya:
closes #1938

Though it's bit tricky that I'm using some elements' width and height to calculate graphs size, it might be somewhat useful. 

CleanShot 2024-10-09 at 01 01 21

### DIFF
--- a/site/frontend/src/graph/render.ts
+++ b/site/frontend/src/graph/render.ts
@@ -400,6 +400,8 @@ function normalizeData(data: CompileGraphData) {
 export type GraphRenderOpts = {
   // Width of the graph
   width: number;
+  // Height of the graph
+  height: number;
   // Render a title above the graph
   renderTitle?: boolean;
   // Function that can be used to hook into the rendering process
@@ -416,7 +418,7 @@ export function renderPlots(
 ) {
   const renderTitle = opts.renderTitle ?? true;
   const hooks = opts.hooks ?? {};
-  const width = opts.width;
+  const {width, height} = opts;
 
   normalizeData(data);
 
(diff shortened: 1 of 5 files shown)


Kobzol:
Thank you! I tried to make the calculations a bit simpler, and tested it with 1/4 columns on desktop/mobile. Let me know what you think!

s7tya:
Thanks! I think it looks great on mobile screens, but on larger screens, isn't the height getting fixed at 300px?

CleanShot 2024-10-10 at 08 09 24@2x

Kobzol:
It should not be (`const height = countGraphs === 1 ? window.innerHeight - top - 100 : 300;`). Did you reload the page after going out of the mobile view? It needs to be refreshed.

It looks like this on my side:
image

Kobzol:
Maybe `innerHeight` behaves different on your browser/OS? :thinking: Also, do you have the devtools closed? If they are opened, the inner height is much smaller if the devtools are docked at the bottom.

s7tya:
Suddenly, it's fixed, but I have no idea why. Anyway, if it works well on your end, let's merge it and see how it goes. I might have failed to clear my cache or something like that.

Kobzol (approved):

PR #1991, opened 2024-10-10
Improve the TUI for the benchmark suite

orhun:
See https://github.com/mainmatter/eurorust-2024-impl-Room/issues/9

cap

> [!WARNING]  
> This is a PoC - expect hackathon code quality!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -314,6 +314,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "cassowary"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Kobzol:
Thank you very much! I think that this is already in a pretty good shape (would be nice to eventually split the large function into a bunch of smaller once, but that's a detail).

In terms of the dependencies, the dependency count goes up from `307` to `329`. However, if we get rid of `inquire`, which shouldn't be needed since we now got `ratatui`, it goes down to `321`. And if we later also get rid of `tabled`, it goes back down to `309` :)

If you're up for it, it would be nice to get rid of `inquire`. Instead of the metric selection dialog, we could have a "select box" of metrics in the main TUI, with instructions being the default. Instead of the artifact selection dialog, we could have another ratatui screen that displays a table of artifacts and lets the user choose the ones they want to compare.

PR #1994, opened 2024-10-11
Use a manual parser for `@rust-timer` commands

Kobzol:
Instead of a regex. This allows the user to pass the arguments in an arbitrary order.

### DIFF
--- a/site/Cargo.toml
+++ b/site/Cargo.toml
@@ -57,3 +57,6 @@ jemalloc-ctl = "0.5"
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 toml = "0.7"
+
+[dev-dependencies]
+insta = "1.40.0"
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -5,15 +5,13 @@ use crate::github::{
 };
 use crate::load::SiteCtxt;
 
-use std::sync::Arc;
-
+use hashbrown::HashMap;
 use regex::Regex;
+use std::sync::Arc;
 
 lazy_static::lazy_static! {
     static ref BODY_TIMER_BUILD: Regex =
         Regex::new(r"(?:\W|^)@rust-timer\s+build\s+(\w+)(?:\W|$)(?:include=(\S+))?\s*(?:exclude=(\S+))?\s*(?:runs=(\d+))?").unwrap();
-    static ref BODY_TIMER_QUEUE: Regex =
-        Regex::new(r"(?:\W|^)@rust-timer\s+queue(?:\W|$)(?:include=(\S+))?\s*(?:exclude=(\S+))?\s*(?:runs=(\d+))?").unwrap();
 }
 
 pub async fn handle_github(
@@ -118,26 +116,25 @@ async fn handle_rust_timer(
         return Ok(github::Response);
     }
 
-    if let Some(captures) = BODY_TIMER_QUEUE.captures(&comment.body) {
-        let include = captures.get(1).map(|v| v.as_str());
-        let exclude = captures.get(2).map(|v| v.as_str());
-        let runs = captures.get(3).and_then(|v| v.as_str().parse::<i32>().ok());
-        {
-            let conn = ctxt.conn().await;
-            conn.queue_pr(issue.number, include, exclude, runs).await;
-        }
-        main_client
-            .post_comment(
-                issue.number,
+    if let Some(queue) = parse_queue_command(&comment.body) {
+        let msg = match queue {
+            Ok(cmd) => {
+                let conn = ctxt.conn().await;
+                conn.queue_pr(issue.number, cmd.include, cmd.exclude, cmd.runs)
+                    .await;
                 format!(
                     "Awaiting bors try build completion.
 
 @rustbot label: +S-waiting-on-perf
 
 {COMMENT_MARK_TEMPORARY}"
-                ),
-            )
-            .await;
+                )
+            }
+            Err(error) => {
+                format!("Error occurred while parsing comment: {error}")
+            }
+        };
+        main_client.post_comment(issue.number, msg).await;
         return Ok(github::Response);
     }
 
@@ -161,6 +158,68 @@ async fn handle_rust_timer(
     .await?;
 
     Ok(github::Response)
+}
+
+/// Parses the first occurrence of a `@rust-timer queue <...>` command
+/// in the input string.
+fn parse_queue_command(body: &str) -> Option<Result<QueueCommand, String>> {
+    let prefix = "@rust-timer";
+    let bot_line = body.lines().find_map(|line| {
+        let line = line.trim();
+        line.find(prefix)
+            .map(|index| line[index + prefix.len()..].trim())
+    })?;
+
+    let args = bot_line.strip_prefix("queue").map(|l| l.trim())?;
+    let mut args = match parse_command_arguments(args) {
+        Ok(args) => args,
+        Err(error) => return Some(Err(error)),
+    };
+    let mut cmd = QueueCommand {
+        include: args.remove("include"),
+        exclude: args.remove("exclude"),
+        runs: None,
+    };
+    if let Some(runs) = args.remove("runs") {
+        let Ok(runs) = runs.parse::<u32>() else {
+            return Some(Err(format!("Cannot parse runs {runs} as a number")));
+        };
+        cmd.runs = Some(runs as i32);
+    }
+
+    if let Some((key, _)) = args.into_iter().next() {
+        return Some(Err(format!("Unknown command argument {key}")));
+    }
+
+    Some(Ok(cmd))
+}
+
+/// Parses command arguments from a single line of text.
+/// Expects that arguments are separated by whitespace, and each argument
+/// has the format `<key>=<value>`.
+fn parse_command_arguments(args: &str) -> Result<HashMap<&str, &str>, String> {
+    let mut argmap = HashMap::new();
+    for arg in args.split_whitespace() {
+        let Some((key, value)) = arg.split_once('=') else {
+            return Err(format!(
+                "Invalid command argument `{arg}` (there may be no spaces around the `=` character)"
+            ));
+        };
+        let key = key.trim();
+        let value = value.trim();
+        if argmap.insert(key, value).is_some() {
+            return Err(format!("Duplicate command argument `{key}`"));
+        }
+    }
+
+    Ok(argmap)
+}
+
+#[derive(Debug)]
+struct QueueCommand<'a> {
+    include: Option<&'a str>,
+    exclude: Option<&'a str>,
+    runs: Option<i32>,
 }
 
 /// Run the `@rust-timer build` regex over the comment message extracting the commit and the other captures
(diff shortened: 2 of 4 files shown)


site/src/request_handlers/github.rs
```diff
@@ -5,15 +5,13 @@ use crate::github::{
 };
 use crate::load::SiteCtxt;
 
-use std::sync::Arc;
-
+use hashbrown::HashMap;
```
lqd, line 8: why hashbrown and not std?

site/Cargo.toml
```diff
@@ -57,3 +57,6 @@ jemalloc-ctl = "0.5"
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 toml = "0.7"
+
+[dev-dependencies]
+insta = "1.40.0"
```
lqd, line 62: do we want/need insta? Managing the snapshots files can be cumbersome, and it's yet another tool to use and learn. It's fine if we want to make good use of it in the future, and not a fancy `assert_eq!(format!("{obj:?}"), "Struct { field: 'bla'} ")`.

site/src/request_handlers/github.rs
```diff
@@ -163,6 +160,68 @@ async fn handle_rust_timer(
     Ok(github::Response)
 }
 
+/// Parses the first occurrence of a `@rust-timer queue <...>` command
+/// in the input string.
+fn parse_queue_command(body: &str) -> Option<Result<QueueCommand, String>> {
+    let prefix = "@rust-timer";
+    let bot_line = body.lines().find_map(|line| {
+        let line = line.trim();
```
lqd, line 168: you probably don't need to trim for a find

site/src/request_handlers/github.rs
```diff
@@ -163,6 +160,68 @@ async fn handle_rust_timer(
...
+        line.find(prefix)
+            .map(|index| line[index + prefix.len()..].trim())
+    })?;
+
+    let args = bot_line.strip_prefix("queue").map(|l| l.trim())?;
```
lqd, line 173: GH never reformats comments, right? That is, it will not introduce unexpected `\n`s that would separate the command from some of its arguments?

site/src/request_handlers/github.rs
```diff
@@ -163,6 +160,68 @@ async fn handle_rust_timer(
...
+    if let Some(runs) = args.remove("runs") {
+        let Ok(runs) = runs.parse::<u32>() else {
+            return Some(Err(format!("Cannot parse runs {runs} as a number")));
+        };
+        cmd.runs = Some(runs as i32);
+    }
+
+    if let Some((key, _)) = args.into_iter().next() {
+        return Some(Err(format!("Unknown command argument {key}")));
+    }
```
lqd, line 192: I guess you could just check that `args` is empty here, and dump all the remaining keys as unknown arguments in the error.

site/src/request_handlers/github.rs
```diff
@@ -118,26 +116,25 @@ async fn handle_rust_timer(
...
-                ),
-            )
-            .await;
+                )
+            }
+            Err(error) => {
+                format!("Error occurred while parsing comment: {error}")
+            }
+        };
+        main_client.post_comment(issue.number, msg).await;
```
lqd, line 137: do we need to take special care (like sanitization) in posting the error message to GH as its text will be partially controlled by users' input commands?

lqd (approved):
Left a few comments.

Here's a free test shuffling the arg order.

```rust
#[test]
fn queue_command_parameter_order() {
    insta::assert_compact_debug_snapshot!(parse_queue_command("@rust-timer queue runs=3 exclude=c,a include=b"),
        @r###"Some(Ok(QueueCommand { include: Some("b"), exclude: Some("c,a"), runs: Some(3) }))"###);
}
```

lqd:
When we do basically this for the `build` command it should also fix the bug I mentioned with the regex handling that ignores some of the commands in the comment. I believe I saw it happen during triage where people can want to `build` multiple shas to analyze a rollup, and "it didn't work".

site/src/request_handlers/github.rs
Kobzol: No real reason, it was autoimported. But I don't see any disadvantage in using `hashbrown`, since we already depend on it, it doesn't really matter. We don't need DOS protection here xD

site/Cargo.toml
Kobzol: Since I'm refactoring this, I want to do it properly. Note that I'm not using snapshot files, I'm using the inline snapshots, which makes this much easier to grok, I think.

I'm already anticipating future changes (like your PR to add the `backends`) parameter. It's no fun updating tens of tests anytime you change the structure of the thing that you parse. Snapshot testing makes this much easier.

I will document in readme what to do to update the snapshots though.

site/src/request_handlers/github.rs
Kobzol: Uhh, that would be quite surprising I think. GH actually has a bunch of formats of the comments, like text, HTML, MD etc., but I think (hope) that here we work with the raw text. We were already requiring `=` to be right after `include` etc. before, so I think that this should be fine.

site/Cargo.toml
lqd: Insta adds snapshot files locally when a test fails, the pending snap file. I saw that when adding the test I shared. It didn’t remove it when the test passed.

site/src/request_handlers/github.rs
Kobzol: That's an interesting idea. I don't think so, it should have the same rules as when you create the comment manually. I guess that the user could make the bot print "Unknown command argument SEND MONEY TO THE FOLLOWING BITCOIN ADDRESS TO SPONSOR THE RUST FOUNDATION", but I don't think that they can XSS or something like that (that would be really bad hole in GH's security).

Kobzol:
I wanted to also refactor the `build` command parsing, but it's a bit tricky, because it has a positional argument, and I need to move on to other stuff atm :( So let's merge this to make it slightly easier to add the `backends` parameter. We don't need to support `backends` in `build` anyway, I haven't ever someone use `include/exclude` with `build` anyway. Well, I haven't ever seen anyone use it with `queue` either, but at least we have a use-case for that now :laughing:

site/Cargo.toml
Kobzol: Ah, these things. Right, I guess we can add them to `.gitignore`, but they should be removed after you use `cargo insta review` I think (the snapshots shouldn't really be modified manually).

lqd: I didn't have `insta` installed to do a `cargo insta review` anyways. We're testing a struct with 3 fields, we don't really need all this test infra imho.

lqd:
Yeah, I've only used include a small number of times in the past. I was mostly thinking of fixing the multiple queue commands bug rather than making my life easier to parse `backends`, it wasn't that annoying to expand the regexes.

Kobzol:
Decided to go all the way and reimplemented also the `build` command :) Let me know what do you think.

site/Cargo.toml
Kobzol: There are already 23 tests in the file, I ain't gonna rewrite all the asserts to make sure we parse exactly what we think we do, every time we change the parsing code :laughing: `insta` was already super helpful for me when developing this PR.

site/src/request_handlers/github.rs
```diff
@@ -138,13 +137,30 @@ async fn handle_rust_timer(
         return Ok(github::Response);
     }
 
-    for captures in build_captures(&comment.body).map(|(_, captures)| captures) {
-        let include = captures.get(2).map(|v| v.as_str());
-        let exclude = captures.get(3).map(|v| v.as_str());
-        let runs = captures.get(4).and_then(|v| v.as_str().parse::<i32>().ok());
```
lqd, line 147: this really does make me wonder why our runs are i32s...

site/src/request_handlers/github.rs
```diff
@@ -138,13 +137,30 @@ async fn handle_rust_timer(
...
-        {
-            let conn = ctxt.conn().await;
-            conn.queue_pr(issue.number, include, exclude, runs).await;
+    let build_cmds: Vec<_> = parse_build_commands(&comment.body).collect();
```
lqd, line 140: probably don't need to collect here

(nor really need the vec in the first place since it's iterated twice anyways, but it's fine ^^)

site/src/request_handlers/github.rs
```diff
@@ -153,47 +169,82 @@ async fn handle_rust_timer(
...
+
+        let sha = sha.trim_start_matches("https://github.com/rust-lang/rust/commit/");
+        let args = iter.next().unwrap_or("");
+        let args = parse_command_arguments(args)?;
+        let params = parse_benchmark_parameters(args)?;
+        Ok(BuildCommand { sha, params })
+    })
+}
+
+fn get_command_lines<'a: 'b, 'b>(
```
lqd, line 211: I don't think you need 'b here?

site/src/request_handlers/github.rs
```diff
@@ -153,47 +169,82 @@ async fn handle_rust_timer(
...
+    command: &'b str,
+) -> impl Iterator<Item = &'a str> + 'b {
+    let prefix = "@rust-timer";
+    body.lines()
+        .filter_map(move |line| {
+            line.find(prefix)
+                .map(|index| line[index + prefix.len()..].trim())
+        })
+        .filter_map(move |line| line.strip_prefix(command))
+        .map(move |l| l.trim_start())
```
lqd, line 222: probably don't need to move either

lqd (approved):
I haven't tried super hard to find edge cases in the parsing because the params are never used but this LGTM.

site/src/request_handlers/github.rs
Kobzol: I'm pretty sure it's a remnant of PostgreSQL - it requires you to use signed integers, or at least the Rust binding that we use.

site/src/request_handlers/github.rs
Kobzol: It's not like I like to add random lifetime parameters everywhere xD It wasn't compiling without it. But apparently it was just me being tired on the bus. This seems to work:
```rust
fn get_command_lines<'a>(body: &'a str, command: &'a str) -> impl Iterator<Item = &'a str> {
    let prefix = "@rust-timer";
    body.lines()
        .filter_map(move |line| {
            line.find(prefix)
                .map(|index| line[index + prefix.len()..].trim())
        })
        .filter_map(move |line| line.strip_prefix(command))
        .map(|l| l.trim_start())
}
```
Thanks!

site/src/request_handlers/github.rs
Kobzol: Thanks, the code was a bit different before, but then I removed the second usage of the Vec.

Kobzol:
Well, there is some regression, the new parser does not really support trailing text, so this fails now:
```
@rust-timer build sha1234 (PR1234)
@rust-timer build sha1234 (PR1235)
```
I'm not sure if we want to support that though, it's not trivial to figure out what is a part of the command and what isn't.

lqd:
Yeah I noticed the test change there, but I'm not sure this is used in practice?

Kobzol:
Haven't really seen it IIRC.

lqd:
If there's any issue (esp since we can't really test the GH interactions), we'll debug it in production like everyone does 🙃

PR #1995, opened 2024-10-12
Mark the local benchmarks as 'master'

s7tya:
I created https://github.com/rust-lang/rustc-perf/pull/1921 to solve https://github.com/rust-lang/rustc-perf/issues/1920, but it didn't align with our goal. So, we'll just mark the local benchmarks as 'master.'

Fixes: https://github.com/rust-lang/rustc-perf/issues/1920


Kobzol (approved):
Works great, thanks a lot! Finally this long-standing papercut has been resolved. Hopefully marking local commits as master won't break anything :laughing:

PR #2000, opened 2024-10-23
Remove artifact from PR queue when purging it

Kobzol:
This should make the `purge_artifact` command more universal, and really purge the given artifact from all places in DB.


lqd (commented):
I don't know where the check the schema but assuming there's no FK or whateverer, this look good to me

Kobzol:
> Are we denying errors only on windows CI?

No, but we run beta on windows CI, that's the difference :)

lqd (approved):
Sorry, my previous comment was supposed to be an approval...

PR #2001, opened 2024-10-23
Fix CI warnings

Kobzol:
Fallout from the new `elided_named_lifetimes` lint.


nnethercote (approved):

lqd:
I don't believe we need to do that? We can update `async-trait` e.g. by cherry-picking https://github.com/rust-lang/rustc-perf/pull/1992/commits/57cabb6c36ddb75d15fa3a117d4e692c7559325c from that PR.

PR #2006, opened 2024-11-07
fix runtime cmp table header

### LINKED ISSUE #2004: Mismatched headers on the runtime tab
The table headers are missing a column corresponding to the arrow icon, causing the headers and data to be shifted by 1 column relative to each other.

Image

s7tya:
fix #2004


Kobzol (approved):
Thank you! :)

PR #2009, opened 2024-11-14
Merge `-Zhir-stats` into `-Zinput-stats`

samestep:
See https://github.com/rust-lang/rust/pull/133023#issuecomment-2477370136.


lqd (approved):
Thanks!

This is just an example of a rustc feature in the `eprintln` profiler documentation; we don't really need to wait for the rustc PR to land.

samestep:
@lqd Makes sense, thanks!

PR #2010, opened 2024-11-17
Add UI for comparing backend results

Kobzol:
This PR adds a new comparison mode to the compile result compare page. This mode allows users to compare the results between the LLVM and Cranelift backends if you are comparing the same commit against each other.

llvm-vs-cranelift

This was IMO the easiest way how to introduce this comparison mode. It would require non-trivial changes to get this working on the backend, and while this is arguably a bit of a hack, I think that it should work fine for the use-case that we need it for. In the future, this could be easily generalized to multiple backends, but for now it was not worth the hassle.

How to test:
1. Get results
```bash
$ rustup component add rustc-codegen-cranelift-preview --toolchain nightly
$ cargo run --bin collector -- bench_local `rustup +nightly which rustc` --include serde --profiles Debug --backends Llvm,Cranelift
```
2. Start the website (if you only have the data from the above run in the local DB, it should open the single commit compared against itself)
3. Click on "Self-compare backend"

### DIFF
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -245,13 +245,12 @@ export function transformDataForBackendComparison(
     }
   }
 
-  const transformed = [];
-  benchmarkMap.forEach((entry) => {
+  return Array.from(benchmarkMap, ([_, entry]) => {
     const comparison: CompileBenchmarkComparison = {
       benchmark: entry.benchmark,
       profile: entry.profile,
       scenario: entry.scenario,
-      // Treat the backend as LLVM
+      // Treat LLVM as the baseline
       backend: "llvm",
       comparison: {
         statistics: [entry.llvm, entry.cranelift],
@@ -260,8 +259,6 @@ export function transformDataForBackendComparison(
         significance_threshold: 1.0,
       },
     };
-    transformed.push(comparison);
+    return comparison;
   });
-
-  return transformed;
 }
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -206,5 +206,5 @@ export function createCompileBenchmarkMap(
 }
 
 export function testCaseKey(testCase: CompileTestCase): string {
-  return `${testCase.benchmark};${testCase.profile};${testCase.scenario};${testCase.category}`;
+  return `${testCase.benchmark};${testCase.profile};${testCase.scenario};${testCase.backend};${testCase.category}`;
 }
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -27,6 +27,7 @@ export type CompileBenchmarkFilter = {
     binary: boolean;
     library: boolean;
   };
+  selfCompareBackend: boolean;
 } & BenchmarkFilter;
 
 export const defaultCompileFilter: CompileBenchmarkFilter = {
@@ -57,6 +58,7 @@ export const defaultCompileFilter: CompileBenchmarkFilter = {
     binary: true,
     library: true,
   },
+  selfCompareBackend: false,
 };
 
 export type Profile = "check" | "debug" | "opt" | "doc";
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -208,3 +208,58 @@ export function createCompileBenchmarkMap(
 export function testCaseKey(testCase: CompileTestCase): string {
   return `${testCase.benchmark};${testCase.profile};${testCase.scenario};${testCase.backend};${testCase.category}`;
 }
+
+// Transform compile comparisons to compare LLVM vs Cranelift, instead of
+// before/after. Assumes that the data comes from the same commit.
+export function transformDataForBackendComparison(
+  comparisons: CompileBenchmarkComparison[]
+): CompileBenchmarkComparison[] {
+  const benchmarkMap: Map<
+    string,
+    {
+      llvm: number | null;
+      cranelift: number | null;
+      benchmark: string;
+      profile: Profile;
+      scenario: string;
+    }
+  > = new Map();
+  for (const comparison of comparisons) {
+    const key = `${comparison.benchmark};${comparison.profile};${comparison.scenario}`;
+    if (!benchmarkMap.has(key)) {
+      benchmarkMap.set(key, {
+        llvm: null,
+        cranelift: null,
+        benchmark: comparison.benchmark,
+        profile: comparison.profile,
+        scenario: comparison.scenario,
+      });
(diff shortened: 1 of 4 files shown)


site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -230,12 +274,22 @@ const filteredSummary = computed(() => computeSummary(comparisons.value));
   />
   <OverallSummary :summary="filteredSummary" />
   <Aggregations :cases="comparisons" />
+  <div class="warning" v-if="selfCompareBackend">
+    Comparing Cranelift against LLVM baseline!
```
lqd, line 278: Maybe something like

```suggestion
    Note: comparing results of the baseline LLVM backend to the cranelift backend.
```

site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -222,6 +258,14 @@ const filteredSummary = computed(() => computeSummary(comparisons.value));
     :selected-metric="selector.stat"
     :metrics="benchmarkInfo.compile_metrics"
   />
+  <div v-if="comparesIdenticalCommits">
+    Self-compare backend:
```
lqd, line 262: Maybe something like

```suggestion
    Compare codegen backends for this commit:
```

maybe with the given commit in the label, like "for commit $sha"

site/frontend/src/pages/compare/compile/common.ts
```diff
@@ -206,5 +208,60 @@ export function createCompileBenchmarkMap(
...
+    }
+  }
+
+  const transformed = [];
+  benchmarkMap.forEach((entry) => {
+    const comparison: CompileBenchmarkComparison = {
+      benchmark: entry.benchmark,
+      profile: entry.profile,
+      scenario: entry.scenario,
+      // Treat the backend as LLVM
```
lqd, line 254:
```suggestion
      // Treat LLVM as the baseline
```

site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -197,15 +215,33 @@ const urlParams = getUrlParams();
 const quickLinksKey = ref(0);
 const filter = ref(loadFilterFromUrl(urlParams, defaultCompileFilter));
 
+// Should we use the backend as the source of before/after data?
+const selfCompareBackend = computed(
+  () => comparesIdenticalCommits.value && filter.value.selfCompareBackend
```
lqd, line 220: should we also check that we have profiling results for the 2 backends?

site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -197,15 +215,33 @@ const urlParams = getUrlParams();
...
   exportToMarkdown(comparisons.value, filter.value.showRawData);
 }
 
+// Are we currently comparing the same commit against each other?
+const comparesIdenticalCommits = computed(() => {
+  return props.data.a.commit === props.data.b.commit;
+});
 const benchmarkMap = createCompileBenchmarkMap(props.data);
+
+// Artificially restructure the data to create a comparison between backends
```
lqd, line 233:
```suggestion
// If requested, artificially restructure the data to create a comparison between backends
```

lqd (commented):
Exciting! 

Here are a few comments/questions

site/frontend/src/pages/compare/compile/common.ts
```diff
@@ -206,5 +208,60 @@ export function createCompileBenchmarkMap(
...
+      benchmarkMap.set(key, {
+        llvm: null,
+        cranelift: null,
+        benchmark: comparison.benchmark,
+        profile: comparison.profile,
+        scenario: comparison.scenario,
+      });
+    }
+    const record = benchmarkMap.get(key);
+    if (comparison.backend === "llvm") {
```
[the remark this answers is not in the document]
Kobzol: We don't, the name comes from [here](https://github.com/rust-lang/rustc-perf/blob/1cb96c189becd5a95e456c0a965ff2d031af1a97/database/src/lib.rs#L355).

site/frontend/src/pages/compare/compile/compile-page.vue
Kobzol: Good point, added a check.

Kobzol:
Thanks for the comments, went through them and resolved them.

site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -222,6 +264,17 @@ const filteredSummary = computed(() => computeSummary(comparisons.value));
     :selected-metric="selector.stat"
     :metrics="benchmarkInfo.compile_metrics"
   />
+  <div
+    v-if="comparesIdenticalCommits"
```
lqd, line 268: I think this should look at the other property, like

```suggestion
    v-if="selfCompareBackend"
```

Otherwise, we'll show this piece of UI even if there isn't cranelift data for a commit: which in practice will always happen on master + perf.rlo.

lqd:
There are a few cases where we have before/after assumptions on the compare page, and I don't know if we need to do anything here now: 
- the links to the detailed query results; the 3 or so links we have there are about commits (it could be interesting to have "before" link be for the llvm backend, and "after" for cranelift) but I don't remember if the page supports selecting the backend, whether we store and load measureme files per backend in the first place, and so on. 
- the frontend/backend/linker sections graph 

These are all useful in understanding where the gains/losses are but it's likely a bit of work, and maybe it's fine not to do anything yet.

Probably not a big deal but when using this we'll also need to remember that we need to use the "after" commit in the compare UI when doing actual single-commit comparisons on rustc-perf. Otherwise there won't be results for both backends on the baseline commit, until we profile cranelift on each merge.

There's also an interesting case that may be more common for the foreseeable future: are broken benchmarks only shown when the 2 commits are different? I say "common" because cg_clif currently doesn't support LTO, so your recent try build shows broken benchmarks [here](https://perf.rust-lang.org/compare.html?start=1e0df74445a185fc97b7e806cd6c636eaded1326&end=51ff489d677fc94da0b0fc98a794cc7e5a06493a&stat=instructions:u) but not on the single-commit comparison. (At some point some of the exa benchmarks were showing as -100% for me, but I can't reproduce this anymore: after some reloads, the failed benchmark rows seem to be gone, and it likely was some transient caching issue.)

site/frontend/src/pages/compare/compile/compile-page.vue
Kobzol: Using this property would make it recursive :sweat_smile: I introduced another one.

lqd:
> At some point some of the exa benchmarks were showing as -100%

These shouldn't have data (and incidentally, toggling "Display raw data" filters these rows out), and I'm not sure exactly when it happens, but for historical purposes this is how it looks when it does happen: 
image.

lqd (approved):
🚀

PR #2014, opened 2024-11-27
Fix touch() failing on Windows

pali6:
It seems like 79faa3e5554c9bd515dbddcf4b9f53c1ecba53f4 introduced an issue on Windows. In order to modify file modification time there it needs to be opened in write mode, otherwise it fails with an "Access is denied" OS error. This PR fixes that.


Kobzol (approved):
Ha, we had the same issue in bootstrap (the rustc build system), but I forgot about it here. Thanks!

PR #2015, opened 2024-11-27
Restore leading spaces on empty context lines in patch files

pali6:
Usually `git apply` does not care about these. However, if the repository is checked out with CRLF line endings, an empty line in the patch context without a leading space results in a "corrupt patch" error.  In other cases this change should have no effect.


Kobzol (approved):
Thanks! :)

PR #2023, opened 2024-12-19
Use `rust-lang` instead of `rust-lang-ci` for all commits

Kobzol:
Needed for moving off `rust-lang-ci`.


Mark-Simulacrum (approved):

Kobzol:
Since it seems like CI is running fine (:crossed_fingers:) on `rust-lang/rust` now, let's merge this. If there are some problems and we need to switch back to the old repository, we can always just revert also this PR again. And even if it breaks, it's not *that* important, just unrolled builds should fail on rollups, and maybe some commit links would be broken.

PR #2031, opened 2025-01-13
Make it easy to add diffing of output of arbitrary profilers (demonstrate for dep-graph)

oli-obk:
I did some cleanups beforehand which, while not strictly necessary, were distracting me 😆

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -1158,18 +1158,15 @@ fn main_result() -> anyhow::Result<i32> {
                         scenarios,
                         &mut errors,
                     );
-                    match diffs.len().cmp(&1) {
-                        Ordering::Equal => {
-                            let short = out_dir.join("cgann-diff-latest");
-                            std::fs::copy(&diffs[0], &short).expect("copy to short path");
-                            eprintln!("Original diff at: {}", diffs[0].to_string_lossy());
-                            eprintln!("Short path: {}", short.to_string_lossy());
-                        }
-                        _ => {
-                            eprintln!("Diffs:");
-                            for diff in diffs {
-                                eprintln!("{}", diff.to_string_lossy());
-                            }
+                    if diffs.len() == 1 {
+                        let short = out_dir.join("cgann-diff-latest");
+                        std::fs::copy(&diffs[0], &short).expect("copy to short path");
+                        eprintln!("Original diff at: {}", diffs[0].to_string_lossy());
+                        eprintln!("Short path: {}", short.to_string_lossy());
+                    } else {
+                        eprintln!("Diffs:");
+                        for diff in diffs {
+                            eprintln!("{}", diff.to_string_lossy());
                         }
                     }
                 }
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -132,14 +132,8 @@ fn check_measureme_installed() -> Result<(), String> {
     }
 }
 
-fn check_installed(name: &str) -> anyhow::Result<()> {
-    if !is_installed(name) {
-        anyhow::bail!("`{}` is not installed but must be", name);
-    }
-    Ok(())
-}
-
-fn generate_cachegrind_diffs(
+#[allow(clippy::too_many_arguments)]
+fn generate_diffs(
     id1: &str,
     id2: &str,
     out_dir: &Path,
@@ -147,6 +141,7 @@ fn generate_cachegrind_diffs(
     profiles: &[Profile],
     scenarios: &[Scenario],
     errors: &mut BenchmarkErrors,
+    profiler: &Profiler,
 ) -> Vec<PathBuf> {
     let mut annotated_diffs = Vec::new();
     for benchmark in benchmarks {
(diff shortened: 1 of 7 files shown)


collector/src/toolchain.rs
```diff
@@ -372,13 +372,13 @@ pub fn get_local_toolchain(
     // specified. This follows the similar pattern used by rustup's binaries
     // (e.g., `rustc +stage1`).
     let (rustc, id) = if let Some(toolchain) = rustc.strip_prefix('+') {
-        let output = Command::new("rustup")
+        let status = Command::new("rustup")
```
Kobzol, line 375: Hmm, this will now randomly print the path to the toolchain to stdout, right? By default, `Command` inherits stdout and stderr, so if you use `status`, it will just print its output. We could add
```rust
.stdout(Stdio::null())
.stderr(Stdio::null())
```
although I personally just use `output` if I want to "hide it" and I don't expect it to be large.

collector/src/utils/diff.rs
```diff
@@ -0,0 +1,25 @@
...
+    process::{Command, Stdio},
+};
+
+use anyhow::Context as _;
+
+use super::is_installed;
+
+/// Compares two text files using `diff` and writes the result to `path`.
+pub fn run_diff(left: &Path, right: &Path, out_file: &Path) -> anyhow::Result<()> {
+    if !is_installed("diff") {
```
Kobzol, line 13: `check_installed` could be used here, since you now made it reusable :)

collector/src/bin/collector.rs
```diff
@@ -1145,29 +1145,26 @@ fn main_result() -> anyhow::Result<i32> {
...
+                    &benchmarks,
+                    profiles,
+                    scenarios,
+                    &mut errors,
+                    &profiler,
+                );
+                if let [diff] = &diffs[..] {
+                    let short = out_dir.join(format!("{}-latest", profiler.output_prefix()));
+                    std::fs::copy(diff, &short).expect("copy to short path");
+                    eprintln!("Original diff at: {}", diffs[0].to_string_lossy());
```
Kobzol, line 1162: `diffs[0]` -> `diff` :)

collector/src/compile/execute/profiler.rs
```diff
@@ -51,6 +52,68 @@ impl Profiler {
                 | Profiler::DepGraph
         )
     }
+
+    pub fn diff_prereqs(&self) -> anyhow::Result<()> {
```
Kobzol, line 56: Can we just move this check directly into `cachegrind_diff`? I don't think that it's worth it to create a separate interface on Profiler for this.

collector/src/compile/execute/profiler.rs
```diff
@@ -51,6 +52,68 @@ impl Profiler {
...
+            Cachegrind => "",
+            DepGraph => ".txt",
+
+            SelfProfile | PerfRecord | Oprofile | Samply | Callgrind | Dhat | DhatCopy | Massif
+            | Bytehound | Eprintln | LlvmLines | MonoItems | LlvmIr => "",
+        }
+    }
+
+    /// A file prefix to use for the diff output.
+    pub fn output_prefix(&self) -> &'static str {
```
Kobzol, line 95: Maybe `diff_prefix`?

But I think that we can also just unify `prefix` and `output_prefix` to make the code simpler, and always just append `-diff` to the prefix. We would need to change the rest of the code to fixup the cachegrind diff paths, but that's fine, I don't think that the specific `cgann-diff` name was that important.

collector/src/compile/execute/profiler.rs
```diff
@@ -51,6 +52,68 @@ impl Profiler {
...
+            Cachegrind => "cgout",
+            DepGraph => "dep-graph",
+
+            SelfProfile | PerfRecord | Oprofile | Samply | Callgrind | Dhat | DhatCopy | Massif
+            | Bytehound | Eprintln | LlvmLines | MonoItems | LlvmIr => "",
+        }
+    }
+
+    /// A postfix added to the file that gets diffed.
+    pub fn postfix(&self) -> &'static str {
```
Kobzol, line 83: Is the `.txt` postfix actually important for anything? If not, we could just get rid of this method.

Kobzol (commented):
Thanks, this makes sense! The cleanup is also nice.

Left a few nits.

collector/src/compile/execute/profiler.rs
oli-obk: Yea, differentiating it from the `.dot` file. Other profilers have multiple outputs, too. I wanna make this return a `Vec` in the future to diff all the individual pieces. Graphviz can produce diff graphs iirc

Kobzol: Ok :)

Kobzol:
There is also profiler diffing for runtime benchmarks (https://github.com/rust-lang/rustc-perf/blob/700be8cbc632063ae6d31a28005be7ee9874425a/collector/src/bin/collector.rs#L799), which could be moved to the new "API". But I don't think that it's super important, I can do it later.

Kobzol (approved):
Agreed. Thanks!

PR #2034, opened 2025-01-22
resolve cargo audit warns and dedupe few deps

klensy:
See commits. Sadly, `idna` pulled big tree of new deps.


collector/src/bin/collector.rs
```diff
@@ -1522,8 +1523,10 @@ fn print_binary_stats(
     table.with(tabled::settings::Style::sharp());
     table.with(
         Modify::new(Rows::last()).with(
-            Border::default()
+            Border::new()
                 .top('─')
+                .left('│')
+                .right('│')
```
klensy, line 1529: This didn't compiled until i added left/right. Looks +- the same for me.

Kobzol:
Thanks! I checked this and it adds like 20 new packages though :disappointed: Could we keep just the RUSTSEC updates, i.e. basically the first and third commits? That on its own actually reduces the package count by a few.

Kobzol:
I tried this: `cargo update ruzstd openssl tempfile futures-util idna` and that resulted in 4 less packages for `cargo build -p collector`.

klensy:
> cargo update ruzstd openssl tempfile futures-util idna

Yes, because your update didn't update `idna` :-)

Kobzol:
I see. I mean, the idna vulnerability doesn't exactly look terrifying, but so be it.

Thanks!

Kobzol (approved):

klensy:
> I see. I mean, the idna vulnerability doesn't exactly look terrifying, but so be it.
> 
> Thanks!

Yes, didn't look that scary.

PR #2042, opened 2025-02-12
Bump MSRV to 1.81

Kobzol:
For https://github.com/rust-lang/rustc-perf/pull/2039.

This actually helped me to uncover a Clippy bug :joy: https://github.com/rust-lang/rust-clippy/issues/14201

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-  pull_request: {}
+  pull_request: { }
   schedule:
     - cron: "0 12 * * 1" # Every Monday at 12:00 UTC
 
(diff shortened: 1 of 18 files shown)


lqd:
The lifetime elision from serde seems unusual and I don’t know what they usually recommend.

lqd (approved):

Kobzol:
Yeah, it's weird, but Clippy seemed adamant..

lqd:
Does serde recommend turning that lint off on these structs?

y21:
> The lifetime elision from serde seems unusual and I don’t know what they usually recommend.

FWIW there's a clippy PR that splits `needless_lifetime` into two lints in rust-lang/rust-clippy#13960, where the behavior of suggesting changing named lifetimes to `'_` (like here) will instead be part of an allow-by-default lint.

PR #2043, opened 2025-02-12
remove lazy_static

s7tya:
Since https://github.com/rust-lang/rustc-perf/pull/2042, now we can use `std::sync::LazyCell` instead of lazy_static.

(FYI: [lazy_static is deprecated](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214))


Kobzol (approved):
Awesome! Thanks.

PR #2046, opened 2025-02-25
Pin the rustc version used to compile collector

Kobzol:
Just to avoid potential surprises with the collector being recompiled with a different (unstable) rustc every day.

### DIFF
--- a/collector/collect.sh
+++ b/collector/collect.sh
@@ -16,8 +16,8 @@ while : ; do
 
         # Make sure we have a recent build, so that we can successfully build
         # the collector.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
Seems ok, though I wonder if we should be trying to wire this up to dependabot or something....

Kobzol:
If we bump MSRV in the repo, but forget to change it in the `collect.sh` script, we will find about that soon enough, I suppose :laughing: But usually I bump it by grepping the old version, so we should see all its occurrences.

PR #2051, opened 2025-03-04
Fix bootstrap rustc benchmark

Kobzol:
This was broken by the change done in https://github.com/rust-lang/rust/pull/137882.


lqd (approved):
I'm not sure if the PR will not be reverted but in the meantime we need data!

Kobzol:
I hope not, the PR made sense, maybe we'll just do some fix up of other bootstrap steps as a follow-up.

PR #2052, opened 2025-03-04
fix - correct column name for pull_request_build SQLite

Jamesbarford:
I believe from the schema the column name `sha` in this instance should be `bors_sha`


Kobzol (approved):
Although this code path will probably never be triggered, it's still a bug ofc. Thank you!

PR #2053, opened 2025-03-04
Fix stable benchmarks

Kobzol:
Fix failures in `html5ever` and `piston-image` on the beta/stable compiler. The performance change should be minimal:
image


lqd (approved):
This looks fine to me as well because:
- one use case is in a build script; another in a dependency (less likely to impact the leaf crate we're benchmarking)
- the impact is not in the compile times of piston-image
- we've bumped rustc-serialize in the stable benchmarks before for similar compatibility issues. This time for the `RustcDecodable` removal, which used to trigger FCW or similar warnings before IIUC.

We _could_ go the opposite way of updating the benchmarks and backfill the data with older rustc releases, if they successfully manage to build these benchmarks, but seeing this PR is not impactful to the actual profiling data, it's simpler to achieve.

Kobzol:
> We _could_ go the opposite way of updating the benchmarks and backfill the data with older rustc releases, if they successfully manage to build these benchmarks, but seeing this PR is not impactful to the actual profiling data, it's simpler to achieve.

We might need to do that once we update collector HW in the future, but we currently don't have enough tooling for this to make it.. *pleasant* :)

lqd:
Yeah for sure. A good feature request for the new architecture.

PR #2059, opened 2025-03-12
Add `target` column, defaulting to `x86_64-unknown-linux-gnu`

Jamesbarford:
The PR adds a `compiler_target` column to the `pstat_series` table. As well as passing through the `compiler_target`; currently it is `CompilerTarget::default()` which is set to `x86_64-unknown-linux-gnu`.

### DIFF
--- a/collector/src/compile/benchmark/compiler_target.rs
+++ b/collector/src/compile/benchmark/compiler_target.rs
@@ -1,0 +1,8 @@
+#[derive(Clone, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
+pub struct CompilerTarget(pub String);
+
+impl Default for CompilerTarget {
+    fn default() -> Self {
+        return Self("x86_64-unknown-linux-gnu".to_string())
+    }
+}
--- a/collector/src/compile/benchmark/compiler_target.rs
+++ b/collector/src/compile/benchmark/compiler_target.rs
@@ -1,8 +1,0 @@
-#[derive(Clone, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
-pub struct CompilerTarget(pub String);
-
-impl Default for CompilerTarget {
-    fn default() -> Self {
-        return Self("x86_64-unknown-linux-gnu".to_string())
-    }
-}
--- a/collector/src/compile/benchmark/target.rs
+++ b/collector/src/compile/benchmark/target.rs
@@ -1,5 +1,11 @@
+/// Target representing an Rust target tripple, for a full list of targets and 
+/// their support see;
+/// https://doc.rust-lang.org/nightly/rustc/platform-support.html
+///
+/// Presently we only support x86_64
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
 pub enum Target {
+    /// `x86_64-unknown-linux-gnu`
     X86_64UnknownLinuxGnu
 }
 
--- a/collector/src/compile/benchmark/target.rs
+++ b/collector/src/compile/benchmark/target.rs
@@ -1,4 +1,4 @@
-/// Target representing an Rust target tripple, for a full list of targets and 
+/// Target representing an Rust target triple, for a full list of targets and
 /// their support see;
 /// https://doc.rust-lang.org/nightly/rustc/platform-support.html
 ///
@@ -6,11 +6,11 @@
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
 pub enum Target {
     /// `x86_64-unknown-linux-gnu`
-    X86_64UnknownLinuxGnu
+    X86_64UnknownLinuxGnu,
 }
 
- impl Default for Target {
+impl Default for Target {
     fn default() -> Self {
         Self::X86_64UnknownLinuxGnu
-   }
+    }
 }
--- a/collector/src/compile/benchmark/target.rs
+++ b/collector/src/compile/benchmark/target.rs
@@ -1,0 +1,10 @@
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
+pub enum Target {
+    X86_64UnknownLinuxGnu
+}
+
+ impl Default for Target {
+    fn default() -> Self {
+        Self::X86_64UnknownLinuxGnu
+   }
+}
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -339,11 +339,17 @@ impl PartialOrd for Scenario {
         Some(self.cmp(other))
     }
 }
-/// The codegen backend used for compilation.
+
+/// Target representing an Rust target tripple, for a full list of targets and 
+/// their support see;
+/// https://doc.rust-lang.org/nightly/rustc/platform-support.html
+///
+/// Presently we only support x86_64
 #[derive(
     Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
 )]
 pub enum Target {
+    /// `x86_64-unknown-linux-gnu`
     X86_64UnknownLinuxGnu
 }
 
@@ -503,7 +509,7 @@ pub struct Index {
     artifacts: Indexed<Box<str>>,
     /// Id lookup of compile stat description ids
     /// For legacy reasons called `pstat_series` in the database, and so the name is kept here.
-    pstat_series: Indexed<(Benchmark, Profile, Scenario, CodegenBackend, Metric, Target)>,
+    pstat_series: Indexed<(Benchmark, Profile, Scenario, CodegenBackend, Target, Metric)>,
     /// Id lookup of runtime stat description ids
     runtime_pstat_series: Indexed<(Benchmark, Metric)>,
 }
@@ -617,7 +623,7 @@ mod index_serde {
     }
 }
 
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub enum DbLabel {
     StatisticDescription {
         benchmark: Benchmark,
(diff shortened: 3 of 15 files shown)


collector/src/compile/benchmark/compiler_target.rs
```diff
@@ -0,0 +1,8 @@
+#[derive(Clone, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
```
Kobzol, line 1: Could you please add a documentation comment on top of the struct to explain what does it represent?

database/src/lib.rs
```diff
@@ -472,7 +472,7 @@ pub struct Index {
     artifacts: Indexed<Box<str>>,
     /// Id lookup of compile stat description ids
     /// For legacy reasons called `pstat_series` in the database, and so the name is kept here.
-    pstat_series: Indexed<(Benchmark, Profile, Scenario, CodegenBackend, Metric)>,
+    pstat_series: Indexed<(Benchmark, Profile, Scenario, CodegenBackend, Metric, String)>,
```
Kobzol, line 475: Let's at least create a type alias `CompileTarget` in the database crate (in this file), to avoid having just `String` here, that's non-obvious.

database/src/pool/postgres.rs
```diff
@@ -278,7 +278,12 @@ static MIGRATIONS: &[&str] = &[
     alter table pstat_series drop constraint pstat_series_crate_profile_cache_statistic_key;
     alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, metric);
     "#,
-    r#"alter table pull_request_build add column backends text;"#,
+    // A pstat series shows 1 compiler_target
```
Kobzol, line 281: It seems like you have deleted one migration by accident.

database/src/pool/postgres.rs
```diff
@@ -617,7 +622,7 @@ where
             pstat_series: self
                 .conn()
                 .query(
-                    "select id, crate, profile, scenario, backend, metric from pstat_series;",
+                    "select id, crate, profile, scenario, backend, metric, compiler_target, from pstat_series;",
```
Kobzol, line 625:
```suggestion
                    "select id, crate, profile, scenario, backend, metric, compiler_target from pstat_series;",
```
There was an extra comma here.

database/src/pool/postgres.rs
```diff
@@ -632,6 +637,7 @@ where
                             row.get::<_, String>(3).as_str().parse().unwrap(),
                             CodegenBackend::from_str(row.get::<_, String>(4).as_str()).unwrap(),
                             row.get::<_, String>(5).as_str().into(),
+                            row.get::<_, String>(6).as_str().parse().unwrap(),
```
Kobzol, line 640:
```suggestion
                            row.get::<_, String>(6).as_str().to_string()
```
or `.into()`

database/src/selector.rs
```diff
@@ -250,20 +252,22 @@ impl BenchmarkQuery for CompileBenchmarkQuery {
...
         let mut statistic_descriptions: Vec<_> = index
             .compile_statistic_descriptions()
-            .filter(|(&(b, p, s, backend, m), _)| {
+            .filter(|(&(b, p, s, backend, m, ref compiler_target), _)| {
                 self.benchmark.matches(b)
                     && self.profile.matches(p)
                     && self.scenario.matches(s)
                     && self.backend.matches(backend)
                     && self.metric.matches(m)
+                    && self.compiler_target.matches(compiler_target.to_string())
```
Kobzol, line 261:
```suggestion
                    && self.compiler_target.matches(&**compiler_target)
```
Bit of a mouthful, but at least we don't need to clone the string, as this part is a bit perf. sensitive.

Kobzol (commented):
Thank you! I would rename it to `CompileTarget`, as compiler target is not a terminology that I know from anywhere and it could be a bit misleading.

I thought that using a String is fine for this, but now that I'm thinking about it, using an enum would have a non-trivial advantage in that it would be more efficient, especially memory wise. The `Index` and the queries go through tens of thousands of commits, and the webserver machine is not very beefy, so I would like to ask you to turn the target into an enum, which should take way less memory (and also be faster).

It will also make it easier to generate CLI help for selecting targets, as we will have a known set of targets.

Jamesbarford:
> Thank you! I would rename it to `CompileTarget`, as compiler target is not a terminology that I know from anywhere and it could be a bit misleading.
> 
> I thought that using a String is fine for this, but now that I'm thinking about it, using an enum would have a non-trivial advantage in that it would be more efficient, especially memory wise. The `Index` and the queries go through tens of thousands of commits, and the webserver machine is not very beefy, so I would like to ask you to turn the target into an enum, which should take way less memory (and also be faster).
> 
> It will also make it easier to generate CLI help for selecting targets, as we will have a known set of targets.

Thanks!

With respect to the `enum` being created, is this to make the column in the Postgres table an `enum`, in the Rust code an `enum` or both?

Kobzol:
I meant to handle it in the same way as we do for the other enums, e.g. `Profile`. So store it as an enum in Rust code, but still a normal string/`TEXT` in the DB (`pstat_series`).

(btw I left you some messages on Zulip regarding the naming :) )

Kobzol (commented):
Thanks, left a few nits.

It would be nice to also modify the `postgres-to-sqlite.rs` and `sqlite-to-postgres.rs` binaries in the `database` crate. It's not super important though, I can do it later if needed.

collector/src/compile/benchmark/target.rs
```diff
@@ -0,0 +1,10 @@
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
```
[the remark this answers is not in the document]
fbstj: should possibly also be `#[non_exhaustive]`?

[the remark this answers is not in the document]
Kobzol: Hmm, that's not needed, the only usage of this enum is within the workspace, and we actually want to exhaustively match on it. We will most likely only add a very small set of targets,

Kobzol (commented):
Thanks! Have you tried to gather benchmarks locally and visualize them in the local website? Just to check that this works. I will also try it locally later.

Jamesbarford:
> Thanks! Have you tried to gather benchmarks locally and visualize them in the local website? Just to check that this works. I will also try it locally later.

I ran the migration locally on a SQLite database I downloaded using the `fetch-latest` tool. The data looked to display ok and have no visual changes from what I had seen previously when running the application.

Kobzol:
Just a note that might be useful going forward: when performing DB changes, it's also useful to test it on Postgres, because the two DB implementations are completely separate.

I use the following `docker-compose.yml` file:

```yaml
version: '3'
services:
    postgres:
        image: 'postgres:15.3'
        ports:
            - '5432:5432'
        environment:
            POSTGRES_PASSWORD: 'postgres'
            POSTGRES_USER: 'postgres'
            POSTGRES_DB: 'perf'
```
and then specify the connection string using the `--db` parameter:
```bash
$ cargo run --bin collector bench_local `rustup +nightly which rustc` --include helloworld --db postgresql://postgres:postgres@127.0.0.1:5432/perf
```

I tested this on Postgres and it looks fine.

collector/src/compile/benchmark/target.rs
```diff
@@ -0,0 +1,16 @@
+/// Target representing an Rust target tripple, for a full list of targets and
```
Kobzol, line 1:
```suggestion
/// Target representing an Rust target triple, for a full list of targets and 
```

database/src/lib.rs
```diff
@@ -594,6 +631,7 @@ pub enum DbLabel {
         scenario: Scenario,
         backend: CodegenBackend,
         metric: Metric,
```
Kobzol, line 633: Could you please also keep the `target` before `metric` here, to keep `metric` at the end?

database/src/lib.rs
```diff
@@ -664,7 +703,7 @@ impl Index {
         self.pstat_series
             .map
             .keys()
-            .map(|(_, _, _, _, metric)| metric)
+            .map(|(_, _, _, _, metric, _)| metric)
```
Kobzol, line 706:
```suggestion
            .map(|(_, _, _, _, _, metric)| metric)
```
Otherwise it would actually select the target.

database/src/pool/postgres.rs
```diff
@@ -631,6 +637,7 @@ where
                             Profile::from_str(row.get::<_, String>(2).as_str()).unwrap(),
                             row.get::<_, String>(3).as_str().parse().unwrap(),
                             CodegenBackend::from_str(row.get::<_, String>(4).as_str()).unwrap(),
+                            Target::from_str(row.get::<_, String>(6).as_str()).unwrap(),
```
Kobzol, line 640:
```suggestion
                            Target::from_str(row.get::<_, String>(5).as_str()).unwrap(),
```
And the line below need to use `(6)`.

database/src/pool/sqlite.rs
```diff
@@ -385,7 +385,23 @@ static MIGRATIONS: &[Migration] = &[
...
+    Migration::without_foreign_key_constraints(
+    r#"
+        create table pstat_series_with_target(
+            id integer primary key not null,
+            crate text not null references benchmark(name) on delete cascade on update cascade,
+            profile text not null,
+            scenario text not null,
+            backend text not null,
+            metric text not null,
+            target text not null default 'x86_64-unknown-linux-gnu',
```
Kobzol, line 398: Since we're recreating the table, might as well again keep the metric column at the end here, it kind of makes more sense (id -> name of the benchmark -> parameters of the benchmark -> result of the benchmark).

Kobzol (approved):
Thank you! I think that this is ready now, but I will hold off merging until Monday, just to avoid deploying on a Friday evening =D

Kobzol:
Ah, yes, [here](https://github.com/rust-lang/rustc-perf/pull/2059/files#diff-d9ce9883872cb2171929ee33dea740e1e5db3b3e662db1159182f4ac53e82e27L388), you deleted an old migration.

Kobzol (approved):
Thank you! :)

PR #2066, opened 2025-04-11
Update cc dependency in all lockfiles

ognevny:
so the packages will be compilable under `*-pc-windows-gnullvm` host. I've tried opt-dist tool in Rust upstream and I found that rustc-perf can't be compiled. I'm unsure about `runtime-benchmarks` packages. please let me know if it's not applicable to change their lockfiles


Kobzol (approved):
Updating the runtime lockfiles is fine. Thanks!

ognevny:
that wasn't enough. actually there are more lockfiles, and not all of them can have required version of cc! (`v1.1.10` is required)

PR #2067, opened 2025-04-11
Update cc for remaining lockfiles

ognevny:
I forgot to update cc in compile-benchmarks crates. when I tried to do so, I found that not all packages got rquired version of cc (>=1.1.10), so I just blibfly ran `cargo update` for all packages to guarantee fresh version of cc...


ognevny:
this is a very huge diff. I'll try to update crates for each package manually

ognevny:
@Kobzol sorry for disturbing with my attempts on making opt-dist for specific host. I tried to minimize diff here, but I didn't manage to win cargo in fight with cc update (it updates cc to >=v1.1.10 only with full lockfile update)

Kobzol:
Hmm, I have much bigger issue with this than the previous PR :) We only update the compile benchmarks if they stop compiling on our supported targets (which is right pretty much only x64 Linux), to keep benchmark stability.

I don't want to do these global lockfile updates for the benchmarks in a single PR.

That being said, we are currently just before a big benchmark update that will update most of the benchmarks to newer versions, which should also bump cc. So after that update the situation might be better for your target. The update will take at least several weeks though (and someone actually has to start doing it..).

ognevny:
understood. I'll use my patch downstream then. waiting for updates

ognevny:
by the way, when you open a PR with compile-benchmark, please mork your PR with "closes #<number of my PR>" so I'll get a notification when your PR is merged :)

Kobzol:
There will probably be ~50 of such PRs (we need on PR to add a new benchmark and then later to remove the old one). I created a `benchmark-update` label to track this.

ognevny:
@Kobzol it seems that some outdated versions are still there :(

Kobzol:
Could you please post in which benchmarks is the outdated `cc` version?

ognevny:
[collector/compile-benchmarks/syn/Cargo.lock](https://github.com/rust-lang/rustc-perf/blob/f265cc6e965e99715a114467f2e96812db776bbb/collector/compile-benchmarks/syn/Cargo.lock#L44)
[collector/compile-benchmarks/piston-image/Cargo.lock](https://github.com/rust-lang/rustc-perf/blob/f265cc6e965e99715a114467f2e96812db776bbb/collector/compile-benchmarks/piston-image/Cargo.lock#L31)

Kobzol:
These two are "stable" benchmarks that we are keeping ideally unchanged for a long time, they are only used for benchmarking stable/beta releases and should not be used for PR or local benchmarks. Why do you need them to work?

ognevny:
> These two are "stable" benchmarks that we are keeping ideally unchanged for a long time, they are only used for benchmarking stable/beta releases and should not be used for PR or local benchmarks. Why do you need them to work?

IIUC these benchmarks are not used with `opt-dist`? in this case I'll close PR

ognevny:
> Ah, no, they shouldn't be used for `opt-dist`.

I don't have a run for current state, but syn seems to be used

Kobzol:
https://github.com/rust-lang/rust/blob/98fdcaed50dfc5346263ed40e9eacb1fc5c124ec/src/build_helper/src/lib.rs#L12-L11 this is the list of benchmarks that `opt-dist` currently uses. Note that `syn` is different from `syn-1.0.89`.

ognevny:
> https://github.com/rust-lang/rust/blob/98fdcaed50dfc5346263ed40e9eacb1fc5c124ec/src/build_helper/src/lib.rs#L12-L11 this is the list of benchmarks that `opt-dist` currently uses. Note that `syn` is different from `syn-1.0.89`.

so this list will be changed to include newer syn?

Kobzol:
Yeah, it's in my TODO list :)

ognevny:
thanks for clarification!

PR #2072, opened 2025-04-16
tests - ordering of the queue

Jamesbarford:
- More tests for the queue invariants. Adding them as guardrails and as cases we can `re-implement` to ensure the new mechanism behaves in the same way.
- I also renamed a few things for clarity

### DIFF
--- a/site/src/load.rs
+++ b/site/src/load.rs
@@ -367,7 +367,7 @@ fn calculate_missing_from(
         .collect::<Vec<_>>();
     let master_commits = queue
         .iter()
-        .map(|(mc, _)| mc.sha.clone())
+        .map(|(commit, _)| commit.sha.clone())
         .collect::<HashSet<_>>();
     for database::QueuedCommit {
         sha,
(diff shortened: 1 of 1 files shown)


site/src/load.rs
```diff
@@ -658,111 +752,229 @@ mod tests {
...
+
+        let expected = vec![create_commit_master_pair("abc", "def", 50, time, false)];
+        assert_eq!(expected, result);
+    }
+
+    #[test]
+    fn duplicate_commits() {
+        let time = chrono::DateTime::from_str("2021-09-01T00:00:00.000Z").unwrap();
+        let master_commits = vec![
+            create_master_commit("dup1", "m1", Some(69), time),
```
Kobzol, line 955: Let's hope this never happens in practice :grin:

Kobzol (approved):
Looks great, having more tests is always super useful. Also adding comments to the code that explain what it does, because the logic is quite complicated. Let me know if you want to modify something else, otherwise LGTM.

PR #2092, opened 2025-04-30
Move typenum-1.18.0 patch file to the correct directory

Kobzol:
Missed this while reviewing https://github.com/rust-lang/rustc-perf/pull/2088.


nnethercote (approved):
I swear I did a test run with incremental to check this. *shrug*  Thanks for fixing.

PR #2095, opened 2025-05-01
Update image to 0.25.6

Kobzol:
Funnily enough, debug compilation is faster, but opt is much slower, at least on icounts:

image


nnethercote (approved):

mayor6304:
Pls can you teach me, am a novice

PR #2119, opened 2025-05-15
Infra - wire up database for test

Jamesbarford:
I've added a Makefile too so we can spin it up and down locally.


database/src/pool.rs
```diff
@@ -294,3 +294,92 @@ impl Pool {
...
+#[cfg(test)]
+mod tests {
+    use chrono::Utc;
+    use std::str::FromStr;
+
+    use super::*;
+    use crate::{tests::run_db_test, Commit, CommitType, Date};
+
+    /// Create a Commit
+    fn create_commit(commit_sha: &str, time: chrono::DateTime<Utc>, r#type: CommitType) -> Commit {
```
Jamesbarford, line 307: I've copied this function from another set of tests from `load.rs`. Should we have something like a `test_utils.rs`

database/src/pool.rs
```diff
@@ -294,3 +294,92 @@ impl Pool {
...
+
+            let artifact_one_id_number = db.connection().await.artifact_id(&artifact_one).await;
+            let artifact_two_id_number = db.connection().await.artifact_id(&artifact_two).await;
+
+            // We cannot arbitrarily add random sizes to the artifact size
+            // table, as their is a constraint that the artifact must actually
+            // exist before attaching something to it.
+
+            // Artifact one inserts
+            db.connection()
```
Kobzol, line 348: There repeated `db.connection().await` calls are boilerplaty and distracting. Let's just create a `let db = db.connection().await;` variable at the start of the function and use that.

database/src/pool.rs
```diff
@@ -294,3 +294,92 @@ impl Pool {
...
+                .get_artifact_size(artifact_one_id_number)
+                .await;
+            let result_two = db
+                .connection()
+                .await
+                .get_artifact_size(artifact_two_id_number)
+                .await;
+
+            // artifact one
+            assert_eq!(Some(32 as u64), result_one.get("llvm.so").copied());
```
Kobzol, line 375:
```suggestion
            assert_eq!(Some(32u64), result_one.get("llvm.so").copied());
```

database/src/pool.rs
```diff
@@ -294,3 +294,92 @@ impl Pool {
...
+            let expected: Vec<Vec<Option<f64>>> = vec![];
+
+            assert_eq!(result, expected);
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn artifact_storage() {
```
Kobzol, line 332: What is this test supposed to be testing? Is it just a smoke test that artifact sizes can be stored?

database/Cargo.toml
```diff
@@ -26,3 +26,4 @@ csv = "1"
 x509-cert = { version = "0.2.5", features = ["pem"] }
 
 intern = { path = "../intern" }
+uuid = "1.16.0"
```
Kobzol, line 29: This should be in `[dev-dependencies]`, because it's only used in tests.

database/src/tests/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+            .unwrap();
+    }
+}
+
+pub(crate) async fn run_db_test<F, Fut>(f: F)
+where
+    F: FnOnce(TestContext) -> Fut,
+    Fut: Future<Output = anyhow::Result<TestContext>>,
+{
+    if let Ok(db_url) = std::env::var("TEST_DB_URL") {
```
Kobzol, line 96: Let's just crash here if `TEST_DB_URL` is unset. In triagebot there were some existing tests that I didn't want to break, but I don't think that many people run `cargo test` on the rustc-perf, and it's safer to just panic if the DB is missing, to avoid a situation where tests are being skipped unexpectedly.

Kobzol (commented):
Thank you! Looks great, left a few notes.

database/src/pool.rs
Jamesbarford: Yes, it's not exactly a complex test but shows how the current mechanics work and proves minimally that the database, for testing, is setup correctly

database/src/tests/mod.rs
Jamesbarford: I have `cfg`'d this as the pipeline failed when running windows as it does not support a docker container. Thus, how I've set things up is to skip the database tests on Windows.

Kobzol (commented):
I'm not such a big fan of the `Makefile`, especially since when running the containers with `-d`, it becomes too easy to let the container continue running in the background, which can slow down the system until restart. It also won't easily work on Windows. But I don't mind keeping it.

What we should add though is some documentation of this, if someone just runs `cargo test`, the failure won't tell them much.

Could you please add a `## How to test` section to `collector/README.md` that explains the need for the Postgres DB in 1-2 sentence and shows the `make test` command? Thank you.

Kobzol (approved):
Thank you!

PR #2144, opened 2025-05-29
Add a test for existence of `[workspace]` in compile benchmarks

Kobzol:
Some compile benchmarks didn't have `[workspace]` in their `Cargo.toml` (discovered [here](https://github.com/rust-lang/rust/pull/141490#issuecomment-2918053883)). This PR fixes that and adds a test for it, so that it doesn't happen again.

### DIFF
--- a/collector/compile-benchmarks/cranelift-codegen-0.119.0/Cargo.toml
+++ b/collector/compile-benchmarks/cranelift-codegen-0.119.0/Cargo.toml
@@ -250,3 +250,5 @@ check-cfg = [
     "cfg(pulley_tail_calls)",
     "cfg(pulley_assume_llvm_makes_tail_calls)",
 ]
+
+[workspace]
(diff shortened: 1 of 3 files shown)


collector/src/compile/benchmark/mod.rs
```diff
@@ -620,13 +620,23 @@ mod tests {
...
+        let benchmark_dir = Path::new(root).join("compile-benchmarks");
+        let benchmarks =
+            get_compile_benchmarks(&benchmark_dir, CompileBenchmarkFilter::All).unwrap();
         assert!(!benchmarks.is_empty());
+
+        for benchmark in benchmarks {
+            let dir = benchmark_dir.join(&benchmark.name.0);
+            let cargo_toml = std::fs::read_to_string(&dir.join("Cargo.toml"))
+                .expect(&format!("Cannot read Cargo.toml of {}", benchmark.name));
+            if !cargo_toml.contains("[workspace]") {
```
lqd, line 634: Maybe as an assert

lqd (approved):
As I said, I also had to do this manually to test the cranelift benchmark but hadn’t had the time to fix it yet. r=me with CI green

collector/src/compile/benchmark/mod.rs
Kobzol: Had that previously, but it didn't contain the right error message. I forgot that we can add arbitrary formatted messages to asserts :facepalm: Changed.

PR #2157, opened 2025-06-11
Feat; Create new table for collector configuration

Jamesbarford:
Creates a table for collector configuration; storing the target, whether the collector is `active` and able to take work. Has a heartbeat column so we can see if the collector goes offline. `benchmark_set` is for when we start splitting work between collectors

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -298,12 +298,18 @@ Columns:
 
 ### collector_config
 
-Information about the collector; it's target architecture, when it was added, whether it is active and when it last had activity denoted by `last_heartbeat_at`.
+Information about the collector; it's target architecture, when it was added,
+whether it is active and when it last had activity denoted by `last_heartbeat_at`.
 
-```
-sqlite> SELECT * FROM collector_config;
+Columns:
 
-id         target                     name  date_added     last_heartbeat_at  benchmark_set  is_active
----------  -------------------------  ----  -------------  ----------------   ---------      -------
-1          aarch64-unknown-linux-gnu  foo   2025-06-11...  2025-06-12 17...   2              0
-```
+* **id** (`id`): A unique identifier for the collector.
+* **target** (`text NOT NULL`): The ISA of the collector for example; `AArch64`.
+* **name** (`text NOT NULL`): Unique name for the collector.
+* **date_added** (`timestamptz NOT NULL`): When the collector was added
+* **last_heartbeat_at** (`timestamptz`): When the collector last updated this
+  column, a way to test if the collector is still alive.
+* **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
+  execute.
+* **is_active** (`boolean NOT NULL`): For controlling whether the collector is
+  active for use. Useful for adding/removing collectors.
--- a/database/schema.md
+++ b/database/schema.md
@@ -305,5 +305,5 @@ sqlite> SELECT * FROM collector_config;
 
 id         target                     date_added     last_heartbeat_at  benchmark_set  is_active
 ---------  -------------------------  -------------  ----------------   ---------      -------
-ea1f4e...  aarch64-unknown-linux-gnu  2025-06-11...  2025-06-12 17...   cea1bc...      0
+1          aarch64-unknown-linux-gnu  2025-06-11...  2025-06-12 17...   2              0
 ```
--- a/database/schema.md
+++ b/database/schema.md
@@ -303,7 +303,7 @@ Information about the collector; it's target architecture, when it was added, wh
 ```
 sqlite> SELECT * FROM collector_config;
 
-id         target                     date_added     last_heartbeat_at  benchmark_set  is_active
----------  -------------------------  -------------  ----------------   ---------      -------
-1          aarch64-unknown-linux-gnu  2025-06-11...  2025-06-12 17...   2              0
+id         target                     name  date_added     last_heartbeat_at  benchmark_set  is_active
+---------  -------------------------  ----  -------------  ----------------   ---------      -------
+1          aarch64-unknown-linux-gnu  foo   2025-06-11...  2025-06-12 17...   2              0
 ```
--- a/database/schema.md
+++ b/database/schema.md
@@ -304,7 +304,7 @@ whether it is active and when it last had activity denoted by `last_heartbeat_at
 Columns:
 
 * **id** (`id`): A unique identifier for the collector.
-* **target** (`text NOT NULL`): The ISA of the collector for example; `AArch64`.
+* **target** (`text NOT NULL`): The ISA of the collector for example; `aarch64-unknown-linux-gnu`.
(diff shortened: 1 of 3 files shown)


database/src/pool/postgres.rs
```diff
@@ -285,6 +285,17 @@ static MIGRATIONS: &[&str] = &[
     alter table pstat_series drop constraint test_case;
     alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, metric);
     "#,
+    r#"CREATE EXTENSION IF NOT EXISTS "uuid-ossp";"#,
+    r#"
+    CREATE TABLE IF NOT EXISTS collector_config (
+        id                UUID PRIMARY KEY,
```
Kobzol: We will have probably max. 4? collectors in the near future :sweat_smile: So I think that UUIDs are overkill. My experience with messing with Postgres extensions is that it can go wrong, I would prefer to avoid this and just use good ol' autoincremented IDs.

On the other hand, I'd like to assign human-readable names to the collectors, for easier debugging. So please add a `name TEXT NOT NULL` column, thanks.

database/src/pool/postgres.rs
```diff
@@ -285,6 +285,17 @@ static MIGRATIONS: &[&str] = &[
...
+        target            TEXT NOT NULL,
+        date_added        TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+        last_heartbeat_at TIMESTAMPTZ,
+        benchmark_set     UUID NOT NULL,
```
Kobzol: Same here, let's just use integers :)

database/src/pool/postgres.rs
Jamesbarford: 69971ef88e1d8e899eb21f71cbc08fae2b878d02

database/schema.md
```diff
@@ -258,3 +258,15 @@ aid         benchmark   error
...
 ```
+
+### collector_config
+
+Information about the collector; it's target architecture, when it was added, whether it is active and when it last had activity denoted by `last_heartbeat_at`.
+
+```
+sqlite> SELECT * FROM collector_config;
+
+id         target                     date_added     last_heartbeat_at  benchmark_set  is_active
```
Kobzol: Could you please update this with `name`? Thanks!

database/src/pool/sqlite.rs
```diff
@@ -404,6 +404,19 @@ static MIGRATIONS: &[Migration] = &[
         alter table pstat_series_with_target rename to pstat_series;
     "#,
     ),
+    Migration::without_foreign_key_constraints(
```
Kobzol: I wonder if we should "delay" the tables on the SQLite side until we finish the Postgres implementation? Just to potentially avoid needless migrations (i.e. Postgres could serve as a "testbed" for the DB). I'm not even sure if we'll ever need this table on the SQLite side, tbh.

Kobzol (commented):
Thanks! I think that the table looks good like it is, but also I don't think we necessarily need to merge this outright.

So I'd keep this lying here until we have some actual code (probably first in the website) that actually needs to use this table, and then write the corresponding DbPool queries so that we have a user of the table. Maybe we'll figure out that we need some additional columns by then or something.

database/schema.md
Jamesbarford: 34b927d872b8445af801966f2807e38f85c7f73c

database/src/pool/postgres.rs
```diff
@@ -285,6 +285,17 @@ static MIGRATIONS: &[&str] = &[
     alter table pstat_series drop constraint test_case;
     alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, metric);
     "#,
+    r#"
+    CREATE TABLE IF NOT EXISTS collector_config (
+        id                SERIAL PRIMARY KEY,
+        target            TEXT NOT NULL,
+        name              TEXT NOT NULL,
```
Kobzol: Let's add a UNIQUE constraint for the name.

database/src/pool/postgres.rs
```diff
@@ -285,6 +285,17 @@ static MIGRATIONS: &[&str] = &[
     alter table pstat_series drop constraint test_case;
     alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, metric);
     "#,
+    r#"
+    CREATE TABLE IF NOT EXISTS collector_config (
+        id                SERIAL PRIMARY KEY,
+        target            TEXT NOT NULL,
```
Kobzol, line 315: Let's add a UNIQUE index constraint on `(target, benchmark_set, is_active)` to ensure that there aren't ever two machines running at the same time benchmarking the same thing.

database/schema.md
```diff
@@ -296,3 +295,15 @@ Columns:
   * `completed`: Completed request.
 * **backends** (`text NOT NULL`): Comma-separated list of codegen backends to benchmark. If empty, the default set of codegen backends will be benchmarked.
 * **profiles** (`text NOT NULL`): Comma-separated list of profiles to benchmark. If empty, the default set of profiles will be benchmarked.
+
+### collector_config
+
+Information about the collector; it's target architecture, when it was added, whether it is active and when it last had activity denoted by `last_heartbeat_at`.
+
+```
```
Kobzol: Same comment as on the other PR, maybe format the fields according to `benchmark_request` above, instead of the SQLite output?

Jamesbarford: 055650d18acc7869d7032470417ad93f52c16b90 :+1:

database/schema.md
```diff
@@ -296,3 +295,21 @@ Columns:
...
+
+### collector_config
+
+Information about the collector; it's target architecture, when it was added,
+whether it is active and when it last had activity denoted by `last_heartbeat_at`.
+
+Columns:
+
+* **id** (`id`): A unique identifier for the collector.
+* **target** (`text NOT NULL`): The ISA of the collector for example; `AArch64`.
```
Kobzol: This is not an actual valid example, because we will be using target tuples :D But it doesn't matter that much.

database/src/pool/postgres.rs
```diff
@@ -309,6 +309,22 @@ static MIGRATIONS: &[&str] = &[
...
+        name              TEXT NOT NULL UNIQUE,
+        date_added        TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+        last_heartbeat_at TIMESTAMPTZ,
+        benchmark_set     INTEGER NOT NULL,
+        is_active         BOOLEAN DEFAULT FALSE NOT NULL,
+
+        -- Given the current setup, we do not want 2 collectors that are active
+        -- with the same target using the same benchmark set.
+        CONSTRAINT collector_config_target_bench_active_uniq
```
Kobzol: I realized that this should have a WHERE condition with `is_active` being true. Otherwise the DB wouldn't allow us to have two inactive machines with the same target and benchmark set.

Kobzol (approved):
Thanks!

PR #2158, opened 2025-06-12
Do not compile extended tools in bootstrap benchmark

Kobzol:
I really don't think that we have to compile things like `llvm-bitcode-linker` and `wasm-component-ld` in the bootstrap benchmark. Especially the compilation is performed with `-j1`, so compiling tens of dependencies that these crates have is quite painful.

This should speed up the bootstrap build by a couple of minutes, I think.

### DIFF
--- a/collector/src/compile/execute/rustc.rs
+++ b/collector/src/compile/execute/rustc.rs
@@ -98,6 +98,9 @@ async fn record(
         "build.cargo={}",
         toolchain.components.cargo.to_str().unwrap()
     ))
+    // Do not compile all default tools
+    .arg("--set")
+    .arg("build.extended=false")
     .status()
     .context("configuring")?;
     assert!(status.success(), "configure successful");


collector/src/compile/execute/rustc.rs
```diff
@@ -98,6 +98,9 @@ async fn record(
         "build.cargo={}",
         toolchain.components.cargo.to_str().unwrap()
     ))
+    // Do not compile all default tools
+    .arg("--set")
+    .arg("build.extended=false")
```
Mark-Simulacrum, line 103: ....why is this on by default? It's literally *extended* :)

But seems fine.

Mark-Simulacrum (approved):

collector/src/compile/execute/rustc.rs
Kobzol: rustc-perf runs configure, which creates a config file based on the dist profile. And dist uses extended = true, which I guess makes sense for it.

Mark-Simulacrum: Hm, the goal of the bootstrap timing is really more focused on rustc devs IMO, not distros, so we might want to change that to compiler or so... though maybe this is enough in practice.

lqd: Hm², the dist profile doesn't `download-ci-llvm`, so I hope we're not building llvm on the collector.

Kobzol: We enable download CI LLVM a few lines before, don't worry :sweat_smile:

PR #2161, opened 2025-06-17
Check test cases with measurements

Kobzol:
With the [new design](https://hackmd.io/wq30YNEIQMSFLWWcWDSI9A), it will be possible to backfill results into the DB, for example if you ask on a PR that you want to see results for the cranelift backend (which is not benchmarked by default), the collector will go back and actually backfill cranelift backend data for the parent master commit.

To support that, we need to expand the notion of a benchmark being "done". Right now, we record a `(artifact, benchmark_name)` tuple into the DB (called a step) when a benchmark begins, and then if we ever encounter the same tuple again, we don't benchmark it again. That's not ideal, because if an error happened and no data was generated, you won't be able to retry the collection without removing everything for the given artifact from the DB. And mainly, you cannot backfill more results (e.g. by running only Debug first, and then backfilling Opt, which is useful also for local experiments).

This PR expands the concept of a benchmark being done by actually checking which compile-time test cases are present in the DB. We cheat a bit to have better perf - if there is at least one recorded statistic in the DB for a given test case, we consider it to be done (so we essentially ignore missing iterations, but that should be a niche edge case).

Even though this logic is mostly useful for the new scheme, which is not implemented yet, I decided to also implement it for the current benchmarking logic, because it's useful for local experiments.


database/src/pool/postgres.rs
```diff
@@ -1397,4 +1427,7 @@ mod tests {
         let certs = make_certificates().await;
         assert!(!certs.is_empty());
     }
+
+    #[tokio::test]
+    async fn foo() {}
```
Jamesbarford: Think this got missed!

Jamesbarford (dismissed):

database/src/pool/postgres.rs
Kobzol: Oops :) Thanks!

Kobzol:
It's a bit complicated :sweat_smile: Detailed are in https://hackmd.io/wq30YNEIQMSFLWWcWDSI9A. The TLDR is:
1) We have a permanent table (`benchmark_request`) that stores all requests for benchmarking artifacts. Once a benchmark is completed there, we mark it as `completed`, and it will never be resubmitted to a collector again (thus we won't auto backfill old PRs when benchmark parameters are added).
2) We have a transient table (`job_queue`), which contains specific jobs for a certain collector machine to benchmark. When a try build needs to backfill certain benchmark parameters, it will also create jobs for the master parent commit that will backfill the missing results into the DB.

Note that this PR does not actually allow backfill with the old system, it should work the same as before. It just allows backfill in the future with the new system, while being compatible with both systems.

collector/src/compile/benchmark/mod.rs
```diff
@@ -273,19 +275,65 @@ impl Benchmark {
...
-                    ));
+                    // Do we have any scenarios left to compute?
+                    let mut remaining_scenarios = scenarios
+                        .iter()
+                        .flat_map(|scenario| {
+                            self.create_test_cases(scenario, profile, backend, target)
+                                .into_iter()
+                                .map(|test_case| (*scenario, test_case))
+                        })
+                        .filter(|(_, test_case)| !already_computed.contains(test_case))
```
Mark-Simulacrum: Shouldn't this condition be more complex? For example, if we add a new incr-patched scenario, we can't run the build for that without going through incr-clean (at least) - and probably the preceding patches, too.

collector/src/compile/benchmark/mod.rs
```diff
@@ -458,6 +507,42 @@ impl Benchmark {
...
     }
+
+    fn create_test_cases(
+        &self,
+        scenario: &Scenario,
+        profile: &Profile,
+        backend: &CodegenBackend,
+        target: &Target,
+    ) -> Vec<CompileTestCase> {
+        self.patches
```
Mark-Simulacrum: Are we expecting this to get more complicated later? It seems a bit wasteful (I guess fine in practice...) to produce a relatively large vec and then dedup most of it away.

collector/src/compile/benchmark/mod.rs
```diff
@@ -273,19 +275,65 @@ impl Benchmark {
...
+                        })
+                        .filter(|(_, test_case)| !already_computed.contains(test_case))
+                        .map(|(scenario, _)| scenario)
+                        .collect::<HashSet<Scenario>>()
+                        .into_iter()
+                        .collect::<Vec<Scenario>>();
+                    if remaining_scenarios.is_empty() {
+                        continue;
+                    }
+                    remaining_scenarios.sort();
```
Mark-Simulacrum: nit: if we're going to sort, maybe just use that for the dedup rather than intermediating through the HashSet?

collector/src/compile/benchmark/mod.rs
Kobzol: If we have an extra patch scenario, `if scenarios.iter().any(|s| s.is_incr()) {` will take care of re-running the base `IncrFull` scenario. We might then have an additional data point for it, but that shouldn't matter much.

Kobzol: We need to handle the preceding patches though, good point.

collector/src/compile/benchmark/mod.rs
Kobzol: Removed the vec/set dedup and sorting.

Mark-Simulacrum (approved):

Kobzol:
Ok, let's try.

PR #2163, opened 2025-06-18
Feat - cron queue for master commits

Jamesbarford:
- Toggle the cron running using `RUN_CRON` environment variable
- Change the frequency of the cron using environment variable `QUEUE_UPDATE_INTERVAL_SECONDS`
- Every `30` seconds by default would populate the `benchmark_requests` table with master commits


database/src/lib.rs
```diff
@@ -798,3 +798,154 @@ pub struct ArtifactCollection {
...
+    InProgress,
+    Completed,
+}
+
+impl fmt::Display for BenchmarkRequestStatus {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            BenchmarkRequestStatus::WaitingForArtifacts => write!(f, "waiting_for_artifacts"),
+            BenchmarkRequestStatus::WaitingForParent => write!(f, "waiting_for_parent"),
+            BenchmarkRequestStatus::Queued => write!(f, "queued"),
```
Kobzol: I think that queued and in progress are mostly the same thing, as per the document? Maybe `Queued` is a better name though. For the purpose of the benchmark request, it shouldn't matter much if the request is only queued or actually running.

database/src/lib.rs
```diff
@@ -798,3 +798,154 @@ pub struct ArtifactCollection {
...
+    Master {
+        sha: String,
+        parent_sha: Option<String>,
+        pr: u32,
+    },
+    /// A release only has a tag
+    Release { tag: String },
+}
+
+impl fmt::Display for BenchmarkRequestType {
```
Kobzol, line 857: I know that we sometimes `impl Display` so that we can just YOLO certain structs into a single TEXT column of a DB, which is already kind of an abuse of the `Display` trait, but here the display trait purposefully ignores certain fields, which seems like too much of a hack. Let's just add a method on the struct, such as `commit_type(&self) -> &str` that will return the commit type string.

database/src/pool/postgres.rs
```diff
@@ -285,6 +286,22 @@ static MIGRATIONS: &[&str] = &[
     alter table pstat_series drop constraint test_case;
     alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, metric);
     "#,
+    r#"
+    CREATE TABLE IF NOT EXISTS benchmark_requests (
```
Kobzol: All our other tables use a singular name, so it should be `benchmark_request`. Sorry for naming it wrong in the document :)

database/src/pool/postgres.rs
```diff
@@ -285,6 +286,22 @@ static MIGRATIONS: &[&str] = &[
...
+        id           SERIAL PRIMARY KEY,
+        tag          TEXT NOT NULL UNIQUE,
+        parent_sha   TEXT,
+        commit_type  TEXT,
```
Kobzol: `TEXT NOT NULL`

database/src/pool/postgres.rs
```diff
@@ -285,6 +286,22 @@ static MIGRATIONS: &[&str] = &[
...
+        pr           INTEGER,
+        created_at   TIMESTAMPTZ NOT NULL,
+        completed_at TIMESTAMPTZ,
+        status       TEXT NOT NULL,
+        backends     TEXT,
```
Kobzol: Let's make it `TEXT NOT NULL` for both `backends` and `profiles`, to avoid having two states describing the same thing (`NULL` and an empty string).

database/src/pool/postgres.rs
```diff
@@ -285,6 +286,22 @@ static MIGRATIONS: &[&str] = &[
...
+        profiles     TEXT
+    );
+    CREATE INDEX IF NOT EXISTS benchmark_requests_status_idx on benchmark_requests (status);
```
Kobzol: Could you please try to fill like 10000 fake entries into the table and see what `EXPLAIN <query>` comes up with? Seems like index on `status` will be useful, not sure if we also need one on `commit_type` though. I think we will only be filtering by `status`?

database/src/lib.rs
```diff
@@ -798,3 +798,154 @@ pub struct ArtifactCollection {
     pub duration: Duration,
     pub end_time: DateTime<Utc>,
 }
+
+#[derive(Debug)]
+pub enum BenchmarkRequestCommitType {
```
Kobzol: This is probably unused now?

database/src/pool/postgres.rs
```diff
@@ -1365,6 +1382,66 @@ where
...
+            }
+            | BenchmarkRequestType::Master {
+                sha,
+                parent_sha,
+                pr,
+            } => {
+                self.conn()
+                    .execute(
+                        r#"
+                        INSERT INTO
```
Kobzol: To avoid repeating the SQL query (which always introduces another possibility of making a mistake when the query needs to be modified), I'd suggest adding `tag(&self) -> &str`, `pr(&self) -> Option<u32>, parent(&self) -> Option<&str>` and `commit_type(&self) -> &str` method to `BenchmarkRequest`. Then we can easily just have the query once here.

I thought about making it a prepared statement, but it will probably not be executed often enough for it to be worth it.

Also, I think that inserting `u32` into Postgres will fail, and it needs to be converted to `i32` for the query. Which is why I'd appreciate at least some basic tests :laughing: E.g. at least insert try/master/release artifact into the DB.

database/src/pool/sqlite.rs
```diff
@@ -1252,6 +1252,10 @@ impl Connection for SqliteConnection {
             )
             .unwrap();
     }
+
+    async fn insert_benchmark_request(&self, _benchmark_request: &BenchmarkRequest) {
+        panic!("Queueing for SQLite has not been implemented, if you are wanting to test the queueing functionality please use postgres. Presuming you have docker installed, at the root of the repo you can run `make start-postgres` to spin up a postrgres database.");
```
Kobzol: `postrgres` -> `postgres` :D

site/src/lib.rs
```diff
@@ -9,6 +9,7 @@ pub mod server;
 mod average;
 mod benchmark_metadata;
 mod comparison;
+mod queue_jobs;
```
Kobzol: Maybe `job_queue` sounds more natural?

site/src/load.rs
```diff
@@ -282,6 +282,27 @@ impl SiteCtxt {
 
         commits
     }
+
+    /// Store the latest master commits or do nothing if all of them are
```
Kobzol: I would ideally like to move all the new logic to a separate module (could be the `job_queue` module for now), so that we can easily see what's the "old" and the "new" logic.

site/src/load.rs
```diff
@@ -282,6 +282,27 @@ impl SiteCtxt {
...
+    /// already in the database
+    pub async fn enqueue_master_commits(&self) {
+        let conn = self.conn().await;
+        let master_commits = &self.get_master_commits().commits;
+
+        for master_commit in master_commits {
```
Kobzol: We should start a transaction before doing this, I think, to have a faster bulk insert. Although in most cases this will be like 1-2 commits at most, so probably it doesn't even matter.

site/src/queue_jobs.rs
```diff
@@ -0,0 +1,23 @@
+use std::sync::Arc;
+
+use crate::load::SiteCtxt;
+use parking_lot::RwLock;
+use tokio::time::{self, Duration};
+
+/// Inserts and manages the queue at `seconds` interval
+pub async fn cron_enqueue_jobs(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u64) {
```
Kobzol: We should add a "cron entrypoint" function that will perform the whole logic of the cron job (essentially right now only line 17 here), and pass `&SiteCtxt` to it. That will make testing easier (it would be annoying to test this function when it loops and waits for time intervals).

Kobzol (commented):
Thank you! ~~I haven't tested it locally yet because it didn't compile (some things need to be shuffled around/fixed I think), but left a couple of comments. I'll test it then.~~ Nevermind, that was some Cargo bug.

Kobzol:
`called `Result::unwrap()` on an `Err` value: Error { kind: ToSql(1), cause: Some(WrongType { postgres: Int4, rust: "u32" }) }` Yup, guessed it right :laughing:

database/src/lib.rs
Jamesbarford: What should the default status be? `waiting_for_parent`? `in_progress` makes sense for when something is happening, I'm not implicitly suggesting a new name. I'm meaning `queued` here to mean `newly_inserted`

Kobzol: I see. I would say:
- For try builds created with `@rust-timer queue`, it's `waiting_for_artifacts`
- For try builds created with `@rust-timer build`, it's `waiting_for_parent`
- For master builds, it's `waiting_for_parent`
- For release artifacts, I suppose that they should be enqueued immediately and they would only ever observe `in_progress` or `completed`? That might be a bit tricky though.

What about this: since we should always try to enqueue requests to the job queue immediately once their parents are ready, maybe we can treat `queued` and `waiting_for_parent` as the same thing. We could name it something like `waiting_for_enqueing` (better name is welcome :D ), and it would mean that the jobs for it haven't been created yet. Then when we encounter such a request, we check if parents are ready. If not, we keep the current state. If yes, then we atomically create the jobs and switch the state to `in_progress`.

Then the default state would be:
- For try builds created with `@rust-timer queue`, it's `waiting_for_artifacts`
- For try builds created with `@rust-timer build`, it's `waiting_for_enqueing`
- For master builds, it's `waiting_for_enqueing`
- For release artifacts, it's `waiting_for_enqueing`

We can pass the default state to the try build constructor, and use `waiting_for_enqueing` for master and release artifacts.

Jamesbarford: Yes having `waiting_for_parent` as the default insert status and then switching to `in_progress` seems like a good approach :+1:

database/src/pool/postgres.rs
Jamesbarford: Sure.

Inserted 100k with a `status` of `waiting_for_parent`.


## No Index
```sql
EXPLAIN WITH commits AS (
	SELECT * FROM benchmark_requests WHERE status = 'waiting_for_parent'
)
SELECT * FROM commits;

-- Produces;

Seq Scan on benchmark_requests  (cost=0.00..2678.00 rows=100000 width=77)
```

## With index
```sql
EXPLAIN WITH commits AS (
	SELECT * FROM benchmark_request WHERE status = 'waiting_for_parent'
)
SELECT * FROM commits;

-- Produces;

Seq Scan on benchmark_requests  (cost=0.00..2678.00 rows=100000 width=77)
```
It's the same as the status is the same for all columns and postgres is smart enough to figure out it's quicker to not use an index.

... So I changed the status;

```sql
EXPLAIN WITH commits AS (
	SELECT * FROM benchmark_request WHERE status = 'in_progress'
)
SELECT * FROM commits;

-- Produces;

Index Scan using benchmark_request_status_idx on benchmark_requests  (cost=0.29..4.31 rows=1 width=77)
```

site/src/queue_jobs.rs
Jamesbarford: :+1: I created `cron_main`

database/src/pool/postgres.rs
Jamesbarford: 500b2ac461480ce8e652d4ab22adf2cd597ee388

database/src/pool/postgres.rs
Jamesbarford: 500b2ac461480ce8e652d4ab22adf2cd597ee388

database/src/pool/postgres.rs
Kobzol: Right, it's hard to benchmark these things without realistic data. I guess that a realistic case is that 99% of stuff is completed, and there's a few in progress/queued things that we want to select.

database/src/lib.rs
```diff
@@ -855,6 +836,24 @@ pub enum BenchmarkRequestType {
     Release { tag: String },
 }
 
+impl BenchmarkRequestType {
+    pub fn commit_type_str(&self) -> &str {
+        match self {
+            BenchmarkRequestType::Try {
+                sha: _,
```
Kobzol, line 843: If you don't need the fields at all, you can also just write `::Try { .. }`. Reduces visual noise.

database/src/lib.rs
```diff
@@ -948,4 +947,61 @@ impl BenchmarkRequest {
             profiles: profiles.to_string(),
         }
     }
+
+    /// Get either the `sha` for a `try` or `master` commit or a `tag` for a
+    /// `release`
+    pub fn tag(&self) -> &str {
+        match &self.commit_type {
+            BenchmarkRequestType::Try {
+                sha,
```
Kobzol: Same here, e.g. `Try { sha, .. }`.

database/src/lib.rs
```diff
@@ -948,4 +947,61 @@ impl BenchmarkRequest {
...
+            BenchmarkRequestType::Try {
+                sha: _,
+                parent_sha,
+                pr: _,
+            }
+            | BenchmarkRequestType::Master {
+                sha: _,
+                parent_sha,
+                pr: _,
+            } => match parent_sha {
```
Kobzol: `parent_sha.as_deref()`

database/src/lib.rs
```diff
@@ -948,4 +947,61 @@ impl BenchmarkRequest {
...
+                Some(parent_sha) => Some(parent_sha),
+                None => None,
```
Kobzol: I realized now that we should store the parent as just `String` for try/master builds, they always need to have *some* parent, and we should know it when the request is being created. We still need to distinguish parent is ready/parent is not ready, but that is done by the status.

site/src/job_queue.rs
```diff
@@ -0,0 +1,59 @@
...
+            ctxt_clone.enqueue_master_commits().await;
+        }
+
+        interval.tick().await;
+        println!("Cron job executed at: {:?}", std::time::SystemTime::now());
+    }
+}
+
+/// Entry point for the cron
+pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u64) {
```
Kobzol, line 43: I probably explained this wrongly :sweat_smile: What I meant was to have an entrypoint function for the cron logic that is being invoked every N seconds (so now for `ctxt_clone.enqueue_master_commits().await;`), so that we can test the cron logic without involving the timer at all.

site/src/job_queue.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::load::SiteCtxt;
+use chrono::Utc;
+use database::{BenchmarkRequest, BenchmarkRequestStatus};
+use parking_lot::RwLock;
+use tokio::time::{self, Duration};
+
+impl SiteCtxt {
+    /// Store the latest master commits or do nothing if all of them are
+    /// already in the database
+    pub async fn enqueue_master_commits(&self) {
```
Kobzol: This is just a style thing, but I don't think it needs to be a method? It can just be a function that takes `SiteCtxt`. It should also be private, it will only be callable from here, we don't want other parts of the website using the enqueing logic, except for a few selected functions that we will expose.

database/src/pool/postgres.rs
Jamesbarford: ## Setup
I've tried again using a new setup, with all the statuses and inserting in random-ish locations;

- Insert 100 "completed"
- Insert 5   "waiting_for_parent"
- Insert 100 "completed"
- Insert 5   "waiting_for_artifact"
- Insert 100 "completed"
- Insert 10  "in_progress"
- Insert 90k "completed"

## Indexes
With the following (B-TREE default) where I filter out the status `'completed'`;
```SQL
CREATE INDEX benchmark_request_status_idx ON benchmark_request (status) WHERE status != 'completed';
```
Rationale behind this is, like you say, the status are 99% of the time going to be complete thus a full `SEQ SCAN` makes sense if we want to look at completed. It will also make the index tiny on disk.

It's been a hot minute since I've used SQL in anger and admittedly forget we could use `EXPLAIN ANALYZE` which gives a much better description of what's going on. With the above index and the following query;

```sql
EXPLAIN ANALYZE SELECT * FROM benchmark_request WHERE status = 'waiting_for_parent';
```

We get the following;

```
 Index Scan using benchmark_request_status_idx on benchmark_request  (cost=0.12..4.14 rows=1 width=68) (actual time=0.028..0.029 rows=0 loops=1)
   Index Cond: (status = 'waiting_for_parent'::text)
 Planning Time: 1.183 ms
 Execution Time: 0.070 ms
```

Which looks about right in my opinion when comparing it against the query with no index;

```
 Seq Scan on benchmark_request  (cost=0.00..2356.25 rows=1 width=68) (actual time=9.908..9.909 rows=0 loops=1)
   Filter: (status = 'waiting_for_parent'::text)
   Rows Removed by Filter: 91220
 Planning Time: 0.365 ms
 Execution Time: 9.938 ms

```

Thus I think our index should be;
```sql
CREATE INDEX benchmark_request_status_idx ON benchmark_request (status) WHERE status != 'completed';
```
Which has, according to `EXPLAIN ANALYZE` a dramatic speed boost, covers our use case and should be small to store on disk.

site/src/load.rs
Jamesbarford: Do we need this? We could dynamically create the query and wrap that in a `BEGIN/COMMIT;` for maximal performance however for 1 or 2 commits that seems excessive?

database/src/lib.rs
Jamesbarford: Nice :+1:, yes, a lot more readable.

database/src/lib.rs
Jamesbarford: This makes sense :+1:

site/src/job_queue.rs
Jamesbarford: Right, I think I've got there

database/src/pool/postgres.rs
Jamesbarford: And for comparison with an index without the `WHERE status != 'complete'`;
```SQL
CREATE INDEX benchmark_request_status_idx ON benchmark_request (status);
```

```
 Index Scan using benchmark_request_status_idx on benchmark_request  (cost=0.29..4.31 rows=1 width=68) (actual time=0.145..0.146 rows=0 loops=1)
   Index Cond: (status = 'waiting_for_parent'::text)
 Planning Time: 0.180 ms
 Execution Time: 0.229 ms
```

Seems a little slower though running both multiple times produces similar-ish results

site/src/load.rs
Kobzol: Yeah I don't think we need it, as the number of commits here will usually be very small.

database/src/pool/postgres.rs
Kobzol: Holy cow, I had no idea you can make such a conditional index. I think that the query that we will actually want to do in the cron job is just `status != 'completed'`, to batch load everything and then handle it in Rust. I suppose that the index works equally well for this case both without and with the `WHERE` clause in the index?

If yes, I'd have a small preference for the "normal" index without WHERE, as I think it is good enough for our scale, and seems less magical (also it would be easier to rename the status column values in the future, although hopefully we won't have to do that).

Jamesbarford: > I think that the query that we will actually want to do in the cron job is just status != 'completed'

I think this is why the index with the condition makes sense. Hypothetically if we did index `"completed"` it would never be used as a SEQ SCAN would be cheaper to execute. Thus we'd be wasting disk space on an index we'd not use;

```sql
EXPLAIN ANALYZE SELECT * FROM benchmark_request WHERE status = 'completed';
```

Produces
```
 Seq Scan on benchmark_request  (cost=0.00..2356.25 rows=91220 width=68) (actual time=0.009..13.934 rows=91220 loops=1)
   Filter: (status = 'completed'::text)
 Planning Time: 1.566 ms
 Execution Time: 16.451 ms
```

I got this idea from within the `postgres.rs` file which has;

https://github.com/rust-lang/rustc-perf/blob/d7d06470bf013d6ef56dbdb5fc4d9f6073af7cf9/database/src/pool/postgres.rs#L223

So I'm strictly speaking not introducing something new to the codebase.

Kobzol: TIL :) So just to clarify, if you have the index with `WHERE status != 'completed'` and then you do a query `WHERE status != 'completed'`, it will be fast, right? :laughing:

Jamesbarford: It should be :smile: given the below information;
```sql
EXPLAIN ANALYZE SELECT * FROM benchmark_request WHERE status != 'completed';
```

```
 Index Scan using benchmark_request_status_idx on benchmark_request  (cost=0.12..4.14 rows=1 width=68) (actual time=0.019..0.020 rows=0 loops=1)
 Planning Time: 0.649 ms
 Execution Time: 0.053 ms
```

site/src/job_queue.rs
```diff
@@ -0,0 +1,57 @@
...
+    // Put the master commits into the `benchmark_requests` queue
+    enqueue_master_commits(site_ctxt).await;
+}
+
+/// Entry point for the cron
+pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u64) {
+    let ctxt = site_ctxt.clone();
+    let mut interval = time::interval(Duration::from_secs(seconds));
+
+    if let Some(ctxt_clone) = {
```
Kobzol: This condition needs to be inside the loop. First we should tick the interval, than check if the ctxt is filled, and then run the `cron_enqueue_jobs` function. Both because the context is missing when `cron_main` is called (so the cron doesn't run at all), and also because if we took the context here, we would be using an old context for the whole duration of the cron job.

site/src/job_queue.rs
```diff
@@ -0,0 +1,57 @@
...
+        let guard = ctxt.read();
+        guard.as_ref().cloned()
+    } {
+        loop {
+            cron_enqueue_jobs(&ctxt_clone).await;
+            interval.tick().await;
+            println!("Cron job executed at: {:?}", std::time::SystemTime::now());
```
Kobzol: Let's use `log::info!("Cron job executed");` so that we get proper timestamps and logging, and also print a message both at the start and end of the cron job logic (so that we can see e.g. if the logic started, but then timeouted somewhere).

Kobzol (commented):
Note: the `created_at` column is currently set to the date of master commits. I guess it doesn't matter that much, as in most cases it should be inserted into the table relatively soon after being merged. But it's a bit weird, since we probably treat `created_at` as the time when the actual entry in the table was created.

That also reminded me that we will want to show the total duration of the benchmark run on the website. This duration was simple before, but it gets more complicated with the job queue and backfilling. Possibly later we might want to add a column to `benchmark_request` that will store the duration of the whole benchmark? Essentially the time from the start of the first job to the completion of the last job, I suppose. But maybe we can reuse `artifact_collection_duration` for that. Anyway, we can deal with that later.

site/src/job_queue.rs
Jamesbarford: Ah! I think this commit should handle that; a145daa6abb53ff3ac22631c896130509311fa21 

I `.clone()` outside of the loop which I think makes sense to increment the reference count. Then do as you say inside the loop and have moved the `interval.tick()` to be the first call to be invoked in the loop

Kobzol: The clone is not necessary as you already hold a value of the `Arc` from the parameter, but it doesn't really matter.

PR #2164, opened 2025-06-18
Make sure to run all clippy lints at warn level

blyxyas:
Previously running all the Clippy bench suite with the general command

```
cargo run --bin collector profile_local perf-record ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustc --profiles Clippy --scenarios Full --exact-match ctfe-stress-5
```

This was because we were running deny-by-default lints and the collect is expecting everything to compile correctly.
Now rustc-perf sets every Clippy lint to `warn` to avoid that problem.

While it could be tempting to just open PRs to the respective failing crates to fix the issues, it's still healthy to run every single lint, even allow-by-defaults.

----

Note that if you run rustc-perf with a non-`master` Clippy, it will still fail because there are some unfiltered CTFE checks. These will be removed in the next release so they're not really a problem.

cc @Kobzol


collector/src/compile/execute/mod.rs
```diff
@@ -329,9 +329,16 @@ impl<'a> CargoProcess<'a> {
...
-                Profile::Check | Profile::Clippy => {
+                Profile::Check => {
                     cmd.arg("--profile").arg("check");
                 }
+                Profile::Clippy => {
+                    cmd.arg("--profile").arg("check");
+                    // Make sure that we run all lints, or else would 
+                    // be pointless for allow-by-default lint benchmarks
+                    // and would cause errors with deny-by-default lints.
+                    cmd.env("RUSTFLAGS", "-Wclippy::all");
```
Kobzol: This overrides externally set rustflags, but it shouldn't be a big deal for the Clippy benchmarks. Just noting it here if anyone runs into it.

blyxyas: In this new patch we even append instead of overwriting! :)

collector/src/compile/execute/mod.rs
```diff
@@ -329,9 +329,21 @@ impl<'a> CargoProcess<'a> {
...
+                Profile::Clippy => {
+                    cmd.arg("--profile").arg("check");
+                    // Make sure that we run all lints, or else would
+                    // be pointless for allow-by-default lint benchmarks
+                    // and would cause errors with deny-by-default lints.
+                    //
+                    // Note that this takes priority over inherited `-Aclippy::*`s
+                    // and similar.
+                    let mut rustflags = env::var("RUSTFLAGS").unwrap_or_default();
+                    rustflags.push_str("-Wclippy::all");
```
Kobzol: Thanks! :heart: In that case here should be a space though, otherwise it would be e.g. `RUSTFLAGS="-Ctarget-cpu=native-Wclippy::all`.

```suggestion
                    rustflags.push_str(" -Wclippy::all");
```

blyxyas: Done! =^w^=

PR #2165, opened 2025-06-23
Port the detailed query page to Vue

Kobzol:
@oli-obk asked me to add some additional query profile data to the table. In order to do that, I moved the page to Vue (with big help from Claude :) ). This also allows us to do client-side sorting of the table, and fix a few sorting related bugs (e.g. sorting by name was somehow buggy before).

### DIFF
--- a/site/frontend/src/pages/detailed-query/page.vue
+++ b/site/frontend/src/pages/detailed-query/page.vue
@@ -567,17 +567,4 @@ code {
 #artifact-table td {
   padding: 0 0 0 20px;
 }
-
-button {
-  margin-right: 10px;
-  padding: 5px 10px;
-  background-color: #f0f0f0;
-  border: 1px solid #ccc;
-  border-radius: 3px;
-  cursor: pointer;
-
-  &:hover {
-    background-color: #e0e0e0;
-  }
-}
 </style>


site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+const tableData = computed(() => {
+  const rows = unsortedTableData.value;
+  if (rows.length === 0) return rows;
+
+  // Separate totals row from data rows
+  const totalsRow = rows.find((row) => row.isTotal);
+  const dataRows = rows.filter((row) => !row.isTotal);
+
+  // Sort data rows based on current sort column and direction
+  const sortedDataRows = [...dataRows].sort((a, b) => {
```
panstromek: You can avoid the clone, because the `.filter` call above already cloned the array

```suggestion
  const sortedDataRows = dataRows.sort((a, b) => {
```

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+        aValue = a.label;
+        bValue = b.label;
+    }
+
+    // Handle string vs number comparison
+    let comparison: number;
+    if (typeof aValue === "string" && typeof bValue === "string") {
+      comparison = aValue.localeCompare(bValue);
+    } else {
+      comparison = (aValue as number) - (bValue as number);
```
panstromek, line 148: Can we ever have `string` vs `number` comparison here or is it safe to assume those will both be numbers at this point?

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+    return currentSortDirection.value === "asc" ? comparison : -comparison;
+  });
+
+  // Return totals row first, then sorted data rows
+  return totalsRow ? [totalsRow, ...sortedDataRows] : sortedDataRows;
+});
+const artifactData = computed(() => createArtifactData(data.value));
+
+function handlePerfettoClick(event: Event, link: string, title: string) {
+  event.preventDefault();
```
panstromek: Vue has `@click.prevent` shorthand syntax for this, btw.

[edit] I see it's used below in other places, so maybe this is intentional?

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+    return {
+      "data-sorted-by": currentSortDirection.value,
+    };
+  }
+  return {};
+}
+
+onMounted(async () => {
+  await loadData();
+});
```
panstromek: You can just call this directly, there's no need to wait for mount (onMounted would be needed only if you need to touch the DOM manually).

```suggestion
loadData();
```

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+            :class="{'total-row': row.isTotal}"
+          >
+            <td>{{ row.label }}</td>
+            <td :title="row.timePercent.title">
+              {{ row.timePercent.formatted }}
+            </td>
+            <td>{{ row.timeSeconds.toFixed(3) }}</td>
+            <td class="delta" v-html="row.timeDelta.formatted"></td>
+            <td>{{ row.executions }}</td>
+            <td class="delta" v-html="row.executionsDelta.formatted"></td>
```
panstromek: `v-html` is pretty discouraged because it's XSS attack vector, but I guess as a transition state it could be ok? This is basically the same thing as the original code AFAICT.

Vue doesn't have React's `dangerouslySetInnerHTML` equivalent of `unsafe` but it's the same thing

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+      ) {
+        comparison = bSecondary - aSecondary; // Higher percentage change comes first
+      }
+    }
+
+    return currentSortDirection.value === "asc" ? comparison : -comparison;
+  });
+
+  // Return totals row first, then sorted data rows
+  return totalsRow ? [totalsRow, ...sortedDataRows] : sortedDataRows;
```
panstromek, line 164: btw. you could also skip the clone here if you use `sortedDataRows.unshift(totalsRow)`

site/frontend/src/pages/detailed-query/utils.ts
```diff
@@ -0,0 +1,380 @@
...
+    classes = "neutral";
+  }
+  if (Math.abs(delta) <= 0.05) {
+    classes = "neutral";
+  }
+  let text;
+  if (isIntegralDelta) {
+    text = delta.toString();
+  } else {
+    text = delta.toFixed(3);
```
panstromek: btw `toFixed` is called quite a lot here and during render. I've found it to be quite slow a few times before (Once I even tried to optimize compare page and it was one of the slow things in the profile.) Could be something to look into if the page is noticeably laggy.

site/frontend/src/pages/detailed-query/page.vue
panstromek: we also swallow the error, but that was present in the original code, too, so that's something we could improve later

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -0,0 +1,557 @@
...
+<template>
+  <div>
+    <div v-if="loading">
+      <p>Loading...</p>
+    </div>
+    <div v-else id="content">
+      <h3 id="title">
+        {{ titleData.text }}
+        <template v-if="selector?.base_commit">
+          <br />diff vs base {{ selector.base_commit.substring(0, 10) }},
```
Jamesbarford, line 306: We seem to be doing;

```js
selector.base_commit.substring(0, 10)
```

And on line 227 & line 298;

```js
selector?.base_commit?.substring(0, 10)
```

Which suggests that `base_commit` could be undefined? I'm not sure what the consequences of it being undefined are exactly however I'm assuming things could crash? I don't know how vue works however can we extract this to a variable outside of the template?

site/frontend/src/pages/detailed-query/page.vue
Kobzol: If that happens, something would be very very wrong with the data :sweat_smile:

site/frontend/src/pages/detailed-query/page.vue
Kobzol: Wasn't intentional, thanks.

site/frontend/src/pages/detailed-query/page.vue
Kobzol: Personally I prefer the simpler code here, as it's more obvious, I always have to google what do the individual shift functions do :D

site/frontend/src/pages/detailed-query/page.vue
Kobzol: Yeah, this shouldn't be here, I'll refactor it, thanks.

site/frontend/src/pages/detailed-query/page.vue
panstromek: the `v-if` on the above line (250) guards against the undefined case

site/frontend/src/pages/detailed-query/utils.ts
Kobzol: I don't think we have enough data for that to be come a problem, but if it becomes laggy, it's good to know where to look!

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -202,41 +216,82 @@ async function loadData() {
...
+  return "header-sort";
 }
 
-onMounted(async () => {
-  await loadData();
-});
+function DeltaComponent({delta}: {delta: DeltaData | null}) {
+  if (delta === null) {
+    return <span>-</span>;
+  }
```
panstromek, line 256: this is very unusual pattern in Vue 😄 , does it actually work? Bit of a "Reactism"

Kobzol: Yeah, it's JSX :laughing: I sometimes use it when I need to reuse a small component and don't want to create a full SFC file for it.

Kobzol: It requires some build system setup, but it seems to work fine, it's already used elsewhere on the site (`tabs.vue`).

panstromek:
> Let's try.

We have `@bors try` at home

PR #2166, opened 2025-06-24
Feat: Logic for finding the next request to be processed and tests

Jamesbarford:
Tees up the next request in the `benchmark_request` table (does not create jobs in what will be the `job_queue` table.

- Tests for database
- Tests for the sorting
  - On sorting; I did implement the sorting logic in SQL - it's probably more performant however it's about 60ish lines and would be fairly difficult to maintain. As we shouldn't be pull back vast amounts of data, and we only do this when something is not `in_progress` I think we should be ok? Though I can bring back the SQL version if needed.
- I couldn't see an obvious path for stubbing out some of the fields on `SiteCtxt` thus split the sorting logic to; `get_next_benchmark_request(...)`
- We want to rename `waiting_for_parent`, I'm a bit lost as to what the name should be so am very open to suggestions. `waiting` along with some code comments about what `waiting` means for the different types of commits was the best I could think of however I held off

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -13,9 +13,7 @@ pub mod interpolate;
 pub mod metric;
 pub mod pool;
 pub mod selector;
-
-#[cfg(test)]
-mod tests;
+pub mod tests;
 
 pub use pool::{Connection, Pool};
 
@@ -799,7 +797,7 @@ pub struct ArtifactCollection {
     pub end_time: DateTime<Utc>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum BenchmarkRequestStatus {
     WaitingForArtifacts,
     ArtifactsReady,
@@ -818,12 +816,34 @@ impl fmt::Display for BenchmarkRequestStatus {
     }
 }
 
-#[derive(Debug)]
+impl<'a> tokio_postgres::types::FromSql<'a> for BenchmarkRequestStatus {
+    fn from_sql(
+        ty: &tokio_postgres::types::Type,
+        raw: &'a [u8],
+    ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
+        // Decode raw bytes into &str with Postgres' own text codec
+        let s: &str = <&str as tokio_postgres::types::FromSql>::from_sql(ty, raw)?;
+
+        match s {
+            "waiting_for_artifacts" => Ok(Self::WaitingForArtifacts),
+            "artifacts_ready" => Ok(Self::ArtifactsReady),
+            "in_progress" => Ok(Self::InProgress),
+            "completed" => Ok(Self::Completed),
+            other => Err(format!("unknown benchmark_request_status '{other}'").into()),
+        }
+    }
+
+    fn accepts(ty: &tokio_postgres::types::Type) -> bool {
+        <&str as tokio_postgres::types::FromSql>::accepts(ty)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
 pub enum BenchmarkRequestType {
     /// A Try commit
     Try {
         sha: String,
-        parent_sha: String,
+        parent_sha: Option<String>,
         pr: u32,
     },
     /// A Master commit
@@ -864,7 +884,7 @@ impl fmt::Display for BenchmarkRequestType {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct BenchmarkRequest {
     pub commit_type: BenchmarkRequestType,
     pub created_at: DateTime<Utc>,
@@ -896,7 +916,7 @@ impl BenchmarkRequest {
 
     pub fn create_try(
         sha: &str,
-        parent_sha: &str,
+        parent_sha: Option<&str>,
         pr: u32,
         created_at: DateTime<Utc>,
         status: BenchmarkRequestStatus,
@@ -907,7 +927,7 @@ impl BenchmarkRequest {
             commit_type: BenchmarkRequestType::Try {
                 pr,
                 sha: sha.to_string(),
-                parent_sha: parent_sha.to_string(),
+                parent_sha: parent_sha.map(|it| it.to_string()),
             },
             created_at,
             completed_at: None,
@@ -964,8 +984,11 @@ impl BenchmarkRequest {
 
     pub fn parent_sha(&self) -> Option<&str> {
         match &self.commit_type {
-            BenchmarkRequestType::Try { parent_sha, .. }
-            | BenchmarkRequestType::Master { parent_sha, .. } => Some(parent_sha),
+            BenchmarkRequestType::Try { parent_sha, .. } => match parent_sha {
+                Some(parent_sha) => Some(parent_sha),
+                _ => None,
+            },
+            BenchmarkRequestType::Master { parent_sha, .. } => Some(parent_sha),
             BenchmarkRequestType::Release { tag: _ } => None,
         }
     }
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -1,7 +1,7 @@
 use crate::selector::CompileTestCase;
 use crate::{
-    ArtifactCollection, ArtifactId, ArtifactIdNumber, BenchmarkRequest, CodegenBackend,
-    CompileBenchmark, Target,
+    ArtifactCollection, ArtifactId, ArtifactIdNumber, BenchmarkRequest, BenchmarkRequestStatus,
+    CodegenBackend, CompileBenchmark, Target,
 };
 use crate::{CollectionId, Index, Profile, QueuedCommit, Scenario, Step};
 use chrono::{DateTime, Utc};
@@ -195,6 +195,20 @@ pub trait Connection: Send + Sync {
         &self,
         artifact_row_id: &ArtifactIdNumber,
     ) -> anyhow::Result<HashSet<CompileTestCase>>;
+
+    /// Gets the benchmark requests matching the status. Optionally provide the
+    /// number of days from whence to search from
+    async fn get_benchmark_requests_by_status(
+        &self,
+        statuses: &[BenchmarkRequestStatus],
+    ) -> anyhow::Result<Vec<BenchmarkRequest>>;
+
+    /// Update the status of a `benchmark_request`
+    async fn update_benchmark_request_status(
+        &mut self,
+        benchmark_request: &BenchmarkRequest,
+        benchmark_request_status: BenchmarkRequestStatus,
+    ) -> anyhow::Result<()>;
 }
 
 #[async_trait::async_trait]
@@ -406,7 +420,7 @@ mod tests {
 
             let try_benchmark_request = BenchmarkRequest::create_try(
                 "b-sha-2",
-                "parent-sha-2",
+                Some("parent-sha-2"),
                 32,
                 time,
                 BenchmarkRequestStatus::ArtifactsReady,
@@ -489,6 +503,102 @@ mod tests {
                 .await
                 .unwrap()
                 .is_empty());
+
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn get_benchmark_requests_by_status() {
+        // Ensure we get back the requests matching the status with no date
+        // limit
+        run_postgres_test(|ctx| async {
+            let db = ctx.db_client();
+            let time = chrono::DateTime::from_str("2021-09-01T00:00:00.000Z").unwrap();
+            let master_benchmark_request = BenchmarkRequest::create_master(
+                "a-sha-1",
+                "parent-sha-1",
+                42,
+                time,
+                BenchmarkRequestStatus::ArtifactsReady,
+                "llvm",
+                "",
+            );
+
+            let try_benchmark_request = BenchmarkRequest::create_try(
+                "b-sha-2",
+                Some("parent-sha-2"),
+                32,
+                time,
+                BenchmarkRequestStatus::Completed,
+                "cranelift",
+                "",
+            );
+
+            let release_benchmark_request = BenchmarkRequest::create_release(
+                "1.8.0",
+                time,
+                BenchmarkRequestStatus::ArtifactsReady,
+                "cranelift,llvm",
+                "",
+            );
+
+            let db = db.connection().await;
+            db.insert_benchmark_request(&master_benchmark_request).await;
+            db.insert_benchmark_request(&try_benchmark_request).await;
+            db.insert_benchmark_request(&release_benchmark_request)
+                .await;
+
+            let requests = db
+                .get_benchmark_requests_by_status(&[BenchmarkRequestStatus::ArtifactsReady])
+                .await
+                .unwrap();
+
+            assert_eq!(requests.len(), 2);
+            assert_eq!(requests[0].status, BenchmarkRequestStatus::ArtifactsReady);
+            assert_eq!(requests[1].status, BenchmarkRequestStatus::ArtifactsReady);
+
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn update_benchmark_request_status() {
+        // Insert one item into the database, change the status and then
+        // get the item back out again to ensure it has changed status
+        run_postgres_test(|ctx| async {
+            let db = ctx.db_client();
+            let time = chrono::DateTime::from_str("2021-09-01T00:00:00.000Z").unwrap();
+            let master_benchmark_request = BenchmarkRequest::create_master(
+                "a-sha-1",
+                "parent-sha-1",
+                42,
+                time,
+                BenchmarkRequestStatus::ArtifactsReady,
+                "llvm",
+                "",
+            );
+
+            let mut db = db.connection().await;
+            db.insert_benchmark_request(&master_benchmark_request).await;
+
+            db.update_benchmark_request_status(
+                &master_benchmark_request,
+                BenchmarkRequestStatus::InProgress,
+            )
+            .await
+            .unwrap();
+
+            let requests = db
+                .get_benchmark_requests_by_status(&[BenchmarkRequestStatus::InProgress])
+                .await
+                .unwrap();
+
+            assert_eq!(requests.len(), 1);
+            assert_eq!(requests[0].tag(), master_benchmark_request.tag());
+            assert_eq!(requests[0].status, BenchmarkRequestStatus::InProgress);
 
             Ok(ctx)
         })
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1,9 +1,9 @@
 use crate::pool::{Connection, ConnectionManager, ManagedConnection, Transaction};
 use crate::selector::CompileTestCase;
 use crate::{
-    ArtifactCollection, ArtifactId, ArtifactIdNumber, Benchmark, BenchmarkRequest, CodegenBackend,
-    CollectionId, Commit, CommitType, CompileBenchmark, Date, Index, Profile, QueuedCommit,
-    Scenario, Target,
+    ArtifactCollection, ArtifactId, ArtifactIdNumber, Benchmark, BenchmarkRequest,
+    BenchmarkRequestStatus, BenchmarkRequestType, CodegenBackend, CollectionId, Commit, CommitType,
+    CompileBenchmark, Date, Index, Profile, QueuedCommit, Scenario, Target,
 };
 use anyhow::Context as _;
 use chrono::{DateTime, TimeZone, Utc};
@@ -1443,6 +1443,113 @@ where
             })
             .collect())
     }
+
+    async fn get_benchmark_requests_by_status(
+        &self,
+        statuses: &[BenchmarkRequestStatus],
+    ) -> anyhow::Result<Vec<BenchmarkRequest>> {
+        // There is a small period of time where a try commit's parent_sha
+        // could be NULL, this query will filter that out.
+        let query = "
+                SELECT
+                    tag,
+                    parent_sha,
+                    pr,
+                    commit_type,
+                    status,
+                    created_at,
+                    completed_at,
+                    backends,
+                    profiles
+                FROM benchmark_request
+                WHERE status = ANY($1)"
+            .to_string();
+
+        let rows = self
+            .conn()
+            .query(&query, &[&statuses])
+            .await
+            .context("Failed to get benchmark requests")?;
+
+        let benchmark_requests = rows
+            .iter()
+            .map(|row| {
+                let tag = row.get::<_, &str>(0);
+                let parent_sha = row.get::<_, Option<&str>>(1);
+                let pr = row.get::<_, Option<i32>>(2);
+                let commit_type = row.get::<_, &str>(3);
+                let status = row.get::<_, BenchmarkRequestStatus>(4);
+                let created_at = row.get::<_, DateTime<Utc>>(5);
+                let completed_at = row.get::<_, Option<DateTime<Utc>>>(6);
+                let backends = row.get::<_, &str>(7);
+                let profiles = row.get::<_, &str>(8);
+
+                match commit_type {
+                    "try" => {
+                        let mut try_benchmark = BenchmarkRequest::create_try(
+                            tag,
+                            parent_sha,
+                            pr.unwrap() as u32,
+                            created_at,
+                            status,
+                            backends,
+                            profiles,
+                        );
+                        try_benchmark.completed_at = completed_at;
+                        try_benchmark
+                    }
+                    "master" => {
+                        let mut master_benchmark = BenchmarkRequest::create_master(
+                            tag,
+                            parent_sha.unwrap(),
+                            pr.unwrap() as u32,
+                            created_at,
+                            status,
+                            backends,
+                            profiles,
+                        );
+                        master_benchmark.completed_at = completed_at;
+                        master_benchmark
+                    }
+                    "release" => {
+                        let mut release_benchmark = BenchmarkRequest::create_release(
+                            tag, created_at, status, backends, profiles,
+                        );
+                        release_benchmark.completed_at = completed_at;
+                        release_benchmark
+                    }
+                    _ => panic!(
+                        "Invalid `commit_type` for `BenchmarkRequest` {}",
+                        commit_type
+                    ),
+                }
+            })
+            .collect();
+        Ok(benchmark_requests)
+    }
+
+    async fn update_benchmark_request_status(
+        &mut self,
+        benchmark_request: &BenchmarkRequest,
+        benchmark_request_status: BenchmarkRequestStatus,
+    ) -> anyhow::Result<()> {
+        let tx = self
+            .conn_mut()
+            .transaction()
+            .await
+            .context("failed to start transaction")?;
+
+        tx.execute(
+            "UPDATE benchmark_request SET status = $1 WHERE tag = $2;",
+            &[&benchmark_request_status, &benchmark_request.tag()],
+        )
+        .await
+        .context("failed to execute UPDATE benchmark_request")?;
+
+        tx.commit().await.context("failed to commit transaction")?;
+
+        Ok(())
+    }
 }
 
 fn parse_artifact_id(ty: &str, sha: &str, date: Option<DateTime<Utc>>) -> ArtifactId {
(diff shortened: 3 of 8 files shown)


site/src/job_queue.rs
```diff
@@ -33,10 +34,97 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+        // create the relevant jobs in the `job_queue`
+        conn.update_benchmark_request_status(&next_request, BenchmarkRequestStatus::InProgress)
+            .await;
+    }
+}
+
 /// For queueing jobs, add the jobs you want to queue to this function
 async fn cron_enqueue_jobs(site_ctxt: &Arc<SiteCtxt>) {
     // Put the master commits into the `benchmark_requests` queue
     enqueue_master_commits(site_ctxt).await;
```
Kobzol: We use the term `enqueue` for putting things into the job queue, not into `benchmark_request`. So to avoid confusion, I would now rename `enqueue_master_commits` to something like `create_master_commit_requests` or something like that.

site/src/job_queue.rs
```diff
@@ -33,10 +34,97 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+    if let Some(next_request) = get_next_benchmark_request(&mut pending, &completed) {
+        // TODO; we simply flip the status for now however this should also
+        // create the relevant jobs in the `job_queue`
+        conn.update_benchmark_request_status(&next_request, BenchmarkRequestStatus::InProgress)
+            .await;
+    }
+}
+
 /// For queueing jobs, add the jobs you want to queue to this function
 async fn cron_enqueue_jobs(site_ctxt: &Arc<SiteCtxt>) {
```
Kobzol: Let's call `let conn = ctxt.conn().await;` once at the beginning of this function, and pass it to the individual subfunctions. No need to create a new DB connection or (return it back to the pool) for each function call.

site/src/job_queue.rs
```diff
@@ -33,10 +34,97 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+    // No requests to process or we have something currently in progress
+    if pending
+        .iter()
+        .any(|r| r.status == BenchmarkRequestStatus::InProgress)
+    {
+        return;
+    }
+
+    // We draw back the last 30 days of completed requests
+    let completed = conn
```
Kobzol: This should be the list of *all* completed benchmark requests, that we ideally preload into a cached hashmap in memory to allow fast lookups (or something like that). The 30 day limit applies to the job queue, not to the benchmark requests. In theory, someone could do a try build with a very old parent (this is supported now), which was benchmarked more than 30 days ago.

site/src/job_queue.rs
```diff
@@ -33,10 +34,97 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+    });
+
+    let completed_set: HashSet<String> = completed.iter().map(|r| r.tag().to_string()).collect();
+
+    let cutoff = Utc::now() - chrono::Duration::days(30);
+
+    let pending_by_sha: HashMap<String, &BenchmarkRequest> =
+        pending.iter().map(|r| (r.tag().to_string(), r)).collect();
+
+    for req in pending.iter() {
```
Kobzol: Interesting, we actually don't need to build an explicit topological ordering here and do the recursive parent lookup, because we can simply take a look if the parent is completed or not. Nevertheless, we will need to build the actual full queue somewhere, because we should show it on the status page, and we also need it for estimating how long it will take to benchmark a specific PR.

database/src/pool.rs
```diff
@@ -183,6 +183,21 @@ pub trait Connection: Send + Sync {
     /// Add an item to the `benchmark_requests`, if the `benchmark_request`
     /// exists it will be ignored
     async fn insert_benchmark_request(&self, benchmark_request: &BenchmarkRequest);
+
+    /// Gets the benchmark requests matching the status. Optionally provide the
+    /// number of days from whence to search from
+    async fn get_benchmark_requests_by_status(
```
Kobzol, line 201: I know that the old code didn't uphold this, but I'd rather avoid unwraps in new code, and explicitly return `Result`s everywhere. Especially since the cron job is a background task, and if it panics, the whole system grinds to a halt. We might want to add some auto-restarting functionality for it, just in case.

Kobzol (commented):
Left a few comments, but it would be best to discuss in sync I think.

site/src/job_queue.rs
Jamesbarford: 441e017fdab0bd32c645a26acbe8188e0d674263 `create_benchmark_request_master_commits()`

site/src/job_queue.rs
Jamesbarford: May you please point in the direction of where something similar happens? However the old implementation looks to cap it at the last 30 days; https://github.com/rust-lang/rustc-perf/blob/817d4b121813c79911384b2a2a078050e21dfed6/site/src/load.rs#L351

site/src/job_queue.rs
Jamesbarford: We discussed this and I've implemented essentially what we have now with some minor modifications along with a test showing the ordering to be the same; https://github.com/rust-lang/rustc-perf/pull/2166/commits/6093f88dfcdd4219244f8d1b66d732257bd58de5

database/src/pool.rs
Jamesbarford: I've used anyhow; ac7b0ff7e3787f8c1b3a00f022f85a57a1094f17

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+        .get_benchmark_requests_by_status(
+            &[
+                BenchmarkRequestStatus::InProgress,
+                BenchmarkRequestStatus::ArtifactsReady,
+            ],
+            None,
+        )
+        .await?;
+
+    // No requests to process or we have something currently in progress
```
Kobzol: The comment is inaccurate, if `pending` is empty, the function will continue below this condition.

site/src/job_queue.rs
Kobzol: That's something a bit different, the old implementation caps last 30 days of master commits, not last 30 days of completed benchmark runs. As noted on the call, in theory someone could start a try build with e.g. a 2 month old master commit (it's rare, but it's something that could happen in the new design). So we should just load all completed requests from the DB. Right now let's load them from the DB, as the new requests table is still small, in the future we should cache them in memory.

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
                 "",
                 "",
             );
             conn.insert_benchmark_request(&benchmark).await;
         }
     }
+    Ok(())
+}
+
+/// This function requires the sorts the priority order of the queue. It assumes
```
Kobzol: The comment seems a bit mangled :)

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+fn sort_benchmark_requests<'a>(
+    done: &mut HashSet<String>,
+    unordered_queue: &'a mut [BenchmarkRequest],
+) -> &'a [BenchmarkRequest] {
+    // A topological sort, where each "level" is additionally altered such that
+    // try commits come first, and then sorted by PR # (as a rough heuristic for
+    // earlier requests).
+
+    // Ensure the queue is ordered by status and the `commit_type`
+    unordered_queue.sort_unstable_by_key(|bmr| (bmr.status.rank(), bmr.commit_type.rank()));
```
Kobzol: We assume that the status is "ArtifactsReady", so we don't need to sort by it, right? I don't think that we need to sort at all, because `partition_in_place` doesn't maintain relative order anyway (`The relative order of partitioned items is not maintained.` from https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition_in_place).

What we should do though is add an assert that all items in `unordered_queue` have status "ArtifactsReady".

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+
+    let mut finished = 0;
+    while finished < unordered_queue.len() {
+        // The next level is those elements in the unordered queue which
+        // are ready to be benchmarked (i.e., those with parent in done or no
+        // parent).
+        let level_len = partition_in_place(unordered_queue[finished..].iter_mut(), |bmr| {
+            bmr.parent_sha().is_none_or(|parent| done.contains(parent))
```
Kobzol, line 60: I would `expect()` the `parent_sha()` here, if it is missing, we have a bug in code.

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+        });
+
+        // No commit is ready for benchmarking. This can happen e.g. when a try parent commit
+        // was forcefully removed from the master branch of rust-lang/rust. In this case, just
+        // let the commits be benchmarked in the current order that we have, these benchmark runs
+        // just won't have a parent result available.
+        if level_len == 0 {
```
Kobzol, line 67: This will also happen if the current parent is in progress, right? We should either also send the in-progress commits to this function, or add the in-progress commits to `done` before calling the function. I know that right now you assume that in-progress artifacts don't exist when this function is called, but we will need to support this for the virtual queue anyway, so I think we can implement it right away, otherwise this logic would have to be changed in the next PR.

site/src/job_queue.rs
```diff
@@ -24,19 +27,118 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+            return unordered_queue;
+        }
+
+        let level = &mut unordered_queue[finished..][..level_len];
+        level.sort_unstable_by_key(|bmr| {
+            (
+                bmr.parent_sha().is_some(),
```
Kobzol: This is not required, we only assume try/master `ArtifactsReady` here and they must have a parent. We shouldn't ever sort release artifacts here, these will be sorted separately. We should sort instead by the commit type - master before try.

database/src/lib.rs
```diff
@@ -852,6 +863,18 @@ impl BenchmarkRequestType {
             BenchmarkRequestType::Release { tag: _ } => "release",
         }
     }
+
+    /// For getting the priority of the request type;
+    /// - Release
+    /// - Master
+    /// - Try
+    pub fn rank(&self) -> u8 {
```
Kobzol: The rank functions should be in `site`, not `database`, IMO. They return ordering that is specific to some logic in `site`, elsewhere we might order these things differently.

site/src/job_queue.rs
Jamesbarford: Yes, sorting by status is pointless. I'll remove that and add an assert to ensure the status is "ArtifactsReady".

We do, however, want to sort by `commit_type.rank()` as the order will be arbitrary when it comes out of the database.

The old implementation assumes the order to be;
```
[<Master>, <Master>, <Master>, <Try>, <Try>, <Try>]
```
And if we pass an arbitrary order to this sort function we do not get the same behaviour as the old implementation.

We now want the order to be;
```
[<Release>, <Master>, <Master>, <Try>, <Try>, <Try>]
```
So we do want to sort by `commit_type`, to preserve the old behaviour.

I was thinking we could sort by the `commit_type` in the SQL, it's a trivial `CASE WHEN...`. However it would then fragment the sorting logic which I think makes it more complicated than it need be. Along with the number of items being small, sorting by `commit_type` in Rust seemed to make sense.

site/src/job_queue.rs
Jamesbarford: Releases however do not have `parent_sha`'s

site/src/job_queue.rs
Jamesbarford: We've sorted by by the `commit_type` at the start of this function. So we do not need to sort it again, sorting here causes the order to be reshuffled and breaks the expected behaviour in the unit tests.

site/src/job_queue.rs
Jamesbarford: I'd intend to have a separate function for displaying the queue that did something like (or use fold to split them into buckets);
```rs
fn get_queue_for_ui() -> Vec<?> {
    let mut queue = conn
        .get_benchmark_requests_by_status(
            &[
                BenchmarkRequestStatus::ArtifactsReady,
            ],
            None,
        )
        .await?;
    let completed_set = conn
        .get_benchmark_requests_by_status(
            &[
                BenchmarkRequestStatus::Completed,
            ],
            None,
        )
        .await?;
    let in_progress = conn
        .get_benchmark_requests_by_status(
            &[
                BenchmarkRequestStatus::InProgress,
            ],
            None,
        )
        .await?;

    completed_set.insert_all(in_progress.iter());

    let sorted_queue = sort_queue(queue, completed_set);
    return in_progress.extend(queue.iter().cloned());
}
```

Presently `enqueue_next_job(...)` is self contained calling `get_next_benchmark_request(...)`. So I'm not seeing what we need to rewrite? It should "just work"?

database/src/lib.rs
Jamesbarford: :+1: Deleted and made them functions in `job_queue.rs`.

site/src/job_queue.rs
Kobzol: But we sort by the partition level first, only then by commit type. In other words, a try request can come ahead of a master request from this function, which is the same as what could happen in the old logic. We also shouldn't ever encounter release (nor in-progress) artifacts in this function, it should only handle master/try artifacts (we might want to add this to the assert :) ).

`partition_in_place` explicitly documents that relative order of items is not kept (we cannot rely on it), so any sorting before calling this function is sort of pointless. The old code was a bit weird here, it assumed that in progress requests are at the start, and that they always have a parent, and that `partition_in_place` won't reorder them but that's relying on implementation details of the partitioning function. Now that we simplified the logic and we only deal with master/try commits here, we shouldn't need any sort.

site/src/job_queue.rs
Kobzol: We shouldn't ever encounter release requests here, right? These should be sorted outside, and put in front of all master/try requests.

We need to filter for them though, because they will also have the `ArtifactsReady` status.

site/src/job_queue.rs
Kobzol: So my assumption was that the code for the virtual queue for UI and the code for determining what to enqueue next is the same code :laughing: And that `sort_benchmark_requests` doesn't ever deal with release artifacts.

site/src/job_queue.rs
Kobzol: I realized that when creating the try benchmark requests, the current code does not actually store the parent SHA, so it is NULL even for try builds. The problem is that the enqueuing of a try benchmark request and the start of a try build usually happen in parallel (`@bors try @rust-timer queue`), so the entry in `pull_request_build` has `parent_sha = NULL`, even though usually the parent SHA is known like 5 seconds after the entry has been created (because the try build usually starts quite quickly). This is caused by us using multiple separate bots for this functionality :(

So we will need to deal with the case where a try benchmark request has `NULL` parent. That actually changes the situation a bit.

Kobzol:
I think we can discuss the sorting logic in sync, now that we can take a look at existing code.

site/src/job_queue.rs
Jamesbarford: > even though usually the parent SHA is known like 5 seconds after the entry has been created

As that is a small amount of time we could simplify things by altering the query to maintain the invariant that a `try` commit has to have a parent; 5c26009c10ca8c2fde5d8965dcc64fbf423ee66d

Kobzol: Sorry, I again missed one invariant.. :laughing: In this function we should only be dealing with `ArtifactsReady` artifacts, these will always have a parent SHA. Try builds that are `WaitingForArtifacts` shouldn't ever come into this function.

Jamesbarford: I've got an assert in the function;
```rust
    // Ensure all the items are ready to be sorted, if they are not this is
    // undefined behaviour
    assert!(unordered_queue
        .iter()
        .all(|bmr| bmr.status == BenchmarkRequestStatus::ArtifactsReady));
```
To ensure that is upheld.

database/src/pool/postgres.rs
```diff
@@ -1413,6 +1414,121 @@ where
             .await
             .unwrap();
     }
+
+    async fn get_benchmark_requests_by_status(
+        &self,
+        statuses: &[BenchmarkRequestStatus],
+        days: Option<i32>,
```
Kobzol: I don't think we have any use-case for the `days` parameter anymore? We should be loading all completed (and all pending) requests from the DB, regardless of their age.

database/src/pool/postgres.rs
```diff
@@ -1413,6 +1414,121 @@ where
...
+            self.conn().query(&query, &[&statuses, &days]).await
+        } else {
+            self.conn().query(&query, &[&statuses]).await
+        }
+        .context("Failed to get benchmark requests")?;
+
+        let benchmark_requests = rows
+            .iter()
+            .map(|row| {
+                let tag = row.get::<_, String>(0);
```
Kobzol: Here we allocate `String`, then we pass it to `create_xyz` as `&str`, and inside that function we turn it into a `String` again. I think we can do `row::get::<_, &str>` to avoid the first allocation for each string here.

database/src/pool/postgres.rs
```diff
@@ -1434,6 +1550,53 @@ fn parse_artifact_id(ty: &str, sha: &str, date: Option<DateTime<Utc>>) -> Artifa
...
+            // Only compile if the type is acceptable
+            tokio_postgres::types::to_sql_checked!();
+        }
+    };
+}
+
+impl_to_postgresql_via_to_string!(BenchmarkRequestType);
+impl_to_postgresql_via_to_string!(BenchmarkRequestStatus);
+
+impl<'a> FromSql<'a> for BenchmarkRequestStatus {
```
Kobzol: Could you please move this just below `impl fmt::Display for BenchmarkRequestStatus`? To keep the hardcoded strings next to each other, so that it's harder to miss updating both if the string representation changes or a new variant is added in the future.

database/src/tests/mod.rs
```diff
@@ -16,18 +16,19 @@ enum TestDb {
...
 /// a database.
-pub(crate) struct TestContext {
+pub struct TestContext {
     test_db: TestDb,
     // Pre-cached client to avoid creating unnecessary connections in tests
     client: Pool,
 }
 
 impl TestContext {
+    #[allow(dead_code)]
```
Kobzol: Let's just `#![allow(dead_code)]` this whole module, will be easier I guess.

site/src/job_queue.rs
```diff
@@ -31,12 +33,145 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
...
+            )
+    }));
+
+    let mut finished = 0;
+    while finished < request_queue.len() {
+        // The next level is those elements in the unordered queue which
+        // are ready to be benchmarked (i.e., those with parent in done or no
+        // parent).
+        let level_len = partition_in_place(request_queue[finished..].iter_mut(), |bmr| {
+            matches!(bmr.commit_type, BenchmarkRequestType::Release { .. })
```
Kobzol:
```suggestion
            
```
I don't think we need this, based on the assert above? :)

site/src/job_queue.rs
```diff
@@ -51,8 +186,412 @@ pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u
...
+        conn.get_benchmark_requests_by_status(&[BenchmarkRequestStatus::InProgress], None)
+            .await
+            .unwrap()
+            .first()
+            .cloned()
+    }
+
+    fn queue_order_matches(queue: &[BenchmarkRequest], expected: &[&str]) -> bool {
+        let queue_shas: Vec<&str> = queue.iter().map(|req| req.tag()).collect();
+        if queue_shas.len() != expected.len() {
```
Kobzol: Maybe just `assert_eq!(expected, &queue_shas)`? The return value probably also isn't needed.

site/src/job_queue.rs
```diff
@@ -51,8 +186,412 @@ pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u
...
+
+            let in_progress = get_in_progress(&*db).await;
+
+            assert!(in_progress.is_none());
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    /// Parent completed < 30 d ago -> child is picked
```
Kobzol: We don't have any 30 day logic here, right? :thinking: That will only be in the job queue.

Kobzol (commented):
Left a few comments, but it looks pretty closely to how I envisioned it. I was a bit confused by the comments about "30 days" in tests though.

database/src/pool/postgres.rs
Jamesbarford: Removed; 88a431cb8a591ebca195481a1985b58f01b2f628

database/src/pool/postgres.rs
Jamesbarford: :+1: 88a431cb8a591ebca195481a1985b58f01b2f628

database/src/pool/postgres.rs
Jamesbarford: :+1: 88a431cb8a591ebca195481a1985b58f01b2f628

database/src/tests/mod.rs
Jamesbarford: :+1: 88a431cb8a591ebca195481a1985b58f01b2f628

site/src/job_queue.rs
Jamesbarford: :+1: 0c477d5dc65a15bf27141d6f7ffa717595a5459f

site/src/job_queue.rs
Jamesbarford: :+1: 0c477d5dc65a15bf27141d6f7ffa717595a5459f

site/src/job_queue.rs
Jamesbarford: :+1: 0c477d5dc65a15bf27141d6f7ffa717595a5459f

collector/Cargo.toml
```diff
@@ -44,6 +44,7 @@ inquire = "0.7.5"
 
 benchlib = { path = "benchlib" }
 database = { path = "../database" }
+uuid = { version = "1.16.0", features = ["v4"] }
```
Kobzol: This shouldn't be needed? The `collector` crate doesn't seem to use `uuid`.

database/src/pool/postgres.rs
```diff
@@ -1413,6 +1413,114 @@ where
             .await
             .unwrap();
     }
+
+    async fn get_benchmark_requests_by_status(
+        &self,
+        statuses: &[BenchmarkRequestStatus],
+    ) -> anyhow::Result<Vec<BenchmarkRequest>> {
+        // There is a small period of time where a try commit's parent_sha
```
Kobzol, line 1451: Sorry, I think I confused you (and myself, again), with the remark about "a few seconds between bors and rustc-perf comments being sent on GitHub". What I meant was that if the two bots were merged, we could know the parent SHA of try builds immediately after starting their build. But since we don't have that (bors and rustc-timer are separate), and we only figure out the parent SHA after the try build **finishes**, not **starts** (this could be changed, but it would be quite complicated, I think), the parent SHA of try builds will actually be `NULL` for quite a long time (e.g. 2-3 hours or more), and it will be `NULL` exactly when the status of the try build will be `WaitingForArtifacts`. So if we called this function with `WaitingForArtifacts` (which we will probably do eventually), it's expected that the parent will be `NULL`.

That means that we will actually need to use `Option<String>` for the `parent_sha` for try builds, sorry. Ideally we'd have an enum that describes this situation properly, but here we have two separate concepts (status and type), and the status can be in different states based on the type (waiting for artifacts is only for try builds, having a parent SHA does not make sense for release builds, etc.). We could model it with one mega enum + a tag that just says "try/master/release", but that would be even more annoying, I guess.

So please just use `parent_sha: Option<String>` for try builds and remove `AND (commit_type ...` and that should solve it.

I swear I'm not giving you conflicting advice by intention :laughing: I'm just discovering all the edge cases and things that we need to handle as we iterate on the code...

database/src/pool/postgres.rs
```diff
@@ -1413,6 +1413,114 @@ where
...
+            .conn()
+            .query(&query, &[&statuses])
+            .await
+            .context("Failed to get benchmark requests")?;
+
+        let benchmark_requests = rows
+            .iter()
+            .map(|row| {
+                let tag = row.get::<_, &str>(0);
+                let parent_sha = row.get::<_, Option<String>>(1);
```
Kobzol: `&str` can be used also for `parent_sha`, `commit_type`, `backends` and `profiles`.

site/src/job_queue.rs
```diff
@@ -31,12 +33,164 @@ async fn enqueue_master_commits(ctxt: &Arc<SiteCtxt>) {
             conn.insert_benchmark_request(&benchmark).await;
         }
     }
+    Ok(())
+}
+
+/// Sorts try and master requests that are in the `ArtifactsReady` status.
+/// Doesn't consider in-progress requests or release artifacts.
+fn sort_benchmark_requests(done: &HashSet<String>, request_queue: &mut [BenchmarkRequest]) {
+    // A topological sort, where each "level" is additionally altered such that
```
Kobzol: This comment is not accurate anymore.

site/src/job_queue.rs
```diff
@@ -51,8 +205,404 @@ pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u
...
+
+            let in_progress = get_in_progress(&*db).await;
+
+            assert_eq!(in_progress.unwrap().tag(), "v1.2.3");
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    /// Parent exists but is > 30 d old -> parent gets picked
```
Kobzol: Outdated comment?

site/src/job_queue.rs
```diff
@@ -51,8 +205,404 @@ pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u
...
+    #[tokio::test]
+    async fn get_next_benchmark_request_oldest_first() {
+        run_postgres_test(|ctx| async {
+            let mut db = ctx.db_client().connection().await;
+            let c1 = create_master("x", "x", 1, "days521")
```
Kobzol, line 371: Two master requests with the same PR should be invalid :thinking: Like, we don't need to check this in code, but it should never happen in practice.

site/src/job_queue.rs
```diff
@@ -51,8 +205,404 @@ pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, seconds: u
...
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn get_next_benchmark_request_large_mixture() {
+        run_postgres_test(|ctx| async {
+            let mut db = ctx.db_client().connection().await;
+            // Fresh parents that will unblock some children
+            let parent_master = create_master("parent_m", "x", 999, "days5")
```
Kobzol: Same here, we should ideally use different PR numbers.

site/src/job_queue.rs
Jamesbarford: b3a37f305c796d711357a1cf0eb8781c72ba0dd5 :+1:

site/src/job_queue.rs
Jamesbarford: b3a37f305c796d711357a1cf0eb8781c72ba0dd5

database/src/pool/postgres.rs
Jamesbarford: b3a37f305c796d711357a1cf0eb8781c72ba0dd5

database/src/pool/postgres.rs
Jamesbarford: All makes sense b3a37f305c796d711357a1cf0eb8781c72ba0dd5

collector/Cargo.toml
Jamesbarford: Interesting, for some reason the pipeline failed at one point due to `uuid` not existing. I've done a clean build locally and, yes, it can be removed; 6523fc135a89013a571934099e8debb9e9cd313f

Kobzol (approved):
Thanks!

PR #2168, opened 2025-06-24
Add query cache hits to detailed query table

Kobzol:
This is blocked on the query hits number actually being gathered in some [non-terribly-expensive way](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Recording.20query.20cache.20hit.20counts/with/525540091) in rustc :sweat_smile:

### DIFF
--- a/site/frontend/src/pages/detailed-query/page.vue
+++ b/site/frontend/src/pages/detailed-query/page.vue
@@ -21,9 +21,10 @@ const showIncr = ref(true);
 const showDelta = ref(true);
 
 type SortDirection = "asc" | "desc";
+type ColumnName = keyof TableRowData;
 
 // Client-side sorting state
-const currentSortColumn = ref<string>("timeSeconds");
+const currentSortColumn = ref<ColumnName>("timeSeconds");
 const currentSortDirection = ref<SortDirection>("desc");
 
 // Computed properties for UI data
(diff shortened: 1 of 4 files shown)


site/frontend/src/pages/detailed-query/utils.ts
```diff
@@ -244,85 +259,66 @@ export function createTableData(
...
+    result.push(createRowData(false, query, queryDelta));
+  });
+
+  return result;
+}
+
+function createRowData(
+  isTotal: boolean,
+  value: ProfileElement,
+  delta: ProfileElementDelta | undefined
```
Jamesbarford, line 276:
```suggestion
  delta?: ProfileElementDelta
```

site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -75,6 +77,15 @@ const tableData = computed(() => {
             ? Math.abs(b.executionsDelta.percentage)
             : 0;
         break;
+      case "cacheHits": // Hits
+        aValue = a.cacheHits;
+        bValue = b.cacheHits;
+        // Use percentage change as secondary sort for equal absolute values
+        aSecondary =
+          a.cacheHitsDelta !== null ? Math.abs(a.cacheHitsDelta.percentage) : 0;
```
Jamesbarford: Stylistic comment.

We could make this more concise in all instances;
```js
aSecondary = Math.abs(a.cacheHitsDelta?.percentage ?? 0)
```

site/frontend/src/pages/detailed-query/utils.ts
Kobzol: I have that in Javascript, I consider these to be different andI have to think about :laughing: To me, `?` says that the argument may be missing, while with `| undefined`, I tell to myself that it has to be present, but its value can be `undefined`. But I know it's actually the same in JS :(

site/frontend/src/pages/detailed-query/page.vue
Kobzol: Good suggestion! Done.

site/frontend/src/pages/detailed-query/utils.ts
panstromek: Yea, It's not the same thing, but it's a bit obscure. Typescript will yell at you if you don't provide `undefined` as argument, in the `|undefined` case.

Whether you passed `undefined` or ommited the argument is observable by looking at the `arguments` object in the function.

Jamesbarford (approved):

Kobzol:
Ok, let's try this!

PR #2184, opened 2025-07-05
bump deps

klensy:
This resolves few cargo audit issues and drops few deps around.

Btw, should i replace sha1 with sha256 check instead? github suggest to do it https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers


klensy:
benchlib pulls clap which affects lock files for runtime benches.

Cargo.toml
```diff
@@ -6,7 +6,8 @@ resolver = "2"
 [workspace.dependencies]
 anyhow = "1"
 chrono = "0.4"
-clap = "4.1"
+# this is latest version with small number of deps
+clap = "=4.1.14"
```
Kobzol: I don't think that reducing the number of deps is so critical here that we would need to pin specific versions, that should be left to really rare circimstances. Just leave it at `"4"`.

Kobzol (commented):
Nice to get rid of `ring`! I don't think we have to get so draconian about pinning dependency versions though.

Cargo.toml
klensy: One of sources of dep number increase here https://github.com/rust-lang/rustc-perf/pull/2034#issuecomment-2607818469 was clap, so i tried to prevent it. Ok.

Kobzol (approved):
Thanks!

PR #2186, opened 2025-07-06
Remove unused dev dependency

Kobzol:
It is no longer needed since `uuid` is now a normal dependency of the `database` crate.


Jamesbarford (approved):
LGTM

PR #2187, opened 2025-07-06
Revert #2161

Kobzol:
This reverts https://github.com/rust-lang/rustc-perf/pull/2161. I noticed that the recomputation logic has some issue on production, and it needlessly reruns stuff when a benchmark was restarted. Also the computation of step durations on the status page suffers when a benchmark is restarted, because we now restart steps, which is something the status page wasn't really prepared for.

The recomputation bug could be fixed, but now I think that we should merge this PR only after we have the new system, to avoid the problems with the step durations.


Kobzol:
For future reference: the bug can be reproduced with the `Doc` profile. The issue is that we think that e.g. `Doc + IncrPatched` is missing, while in reality it is never computed (even though it is a part of the profile x scenario matrix). We have to precompute/prefilter the matrix before checking DB.

Mark-Simulacrum (approved):

Kobzol:
I just realized, if we revert this, and don't land https://github.com/rust-lang/rustc-perf/pull/2188, then we'll be potentially skipping more test cases (because the old logic would just skip everything once). But it should always be at most one benchmark (crate) skipped, so hopefully that's still not so terrible.

Kobzol:
Landing as a hotfix, it should help us make progress on the collector in the face of S3 upload failures.

PR #2188, opened 2025-07-06
Do not panic on S3 file upload failures

Kobzol:
We have started hitting this a few times today when uploading the self-profile artifacts to S3:
```
upload failed: ../../../tmp/.tmp1aGYxm to s3://rustc-perf/self-profile/67394/derive/check/full/self-profile-32456021.mm_profdata.sz SSL validation failed for https://rustc-perf.s3.us-west-1.amazonaws.com/self-profile/67394/derive/check/full/self-profile-32456021.mm_profdata.sz EOF occurred in violation of protocol (_ssl.c:2406)
```
I don't think it's worth it to crash/interrupt the whole collection when a self-profile file can't be uploaded, we use them very sparsely anyway, and if the unlikely case happened that an important self-profile case was missing, we can just rerun locally.

### DIFF
--- a/collector/src/compile/execute/bencher.rs
+++ b/collector/src/compile/execute/bencher.rs
@@ -345,9 +345,9 @@ impl SelfProfileS3Upload {
         let start = std::time::Instant::now();
         let status = self.0.wait().expect("waiting for child");
         if !status.success() {
-            panic!("S3 upload failed: {:?}", status);
+            log::error!("S3 upload failed: {status:?}");
+        } else {
+            log::trace!("uploaded to S3, additional wait: {:?}", start.elapsed());
         }
-
-        log::trace!("uploaded to S3, additional wait: {:?}", start.elapsed());
     }
 }


Mark-Simulacrum:
> unlikely case happened that an important self-profile case was missing, we can just rerun locally

I'm not sure I agree. Running locally isn't completely trivial -- you need to stand up the site, figure out the right arguments, etc.?

I believe this error is not entirely unknown (https://github.com/rust-lang/promote-release/issues/77) but unfortunately we've never been able to narrow down the exact conditions that trigger it (including IPs we're hitting, maybe a tcpdump) to help AWS fix it.

I think we should add a retry to the upload for now rather than just ignoring the upload failure. If there's something more streamlined compared to panic that we should do (maybe treat it as a benchmark failure?) that's fine with me too.

Mark-Simulacrum (approved):

Kobzol:
Merging as a hotfix, to unblock the collector. S3 upload seems to be acting up today..

PR #2189, opened 2025-07-06
Log duration of self-profile parsing

Kobzol:
I profiled rustc-perf locally, and it seems like we spent a bunch of time on parsing the self-profile files. This is quite wasteful, because they can be quite large, and we iterate through the whole file, load and allocate all the event labels, and then only take the artifact numbers from it, which is like 6 numbers.

I have code locally that optimizes this so that the loading is 2x faster, but I want to first see how much time we actually spend on this in production. My estimate is around 1 minute per benchmark run.

### DIFF
--- a/collector/src/compile/execute/mod.rs
+++ b/collector/src/compile/execute/mod.rs
@@ -22,6 +22,7 @@ use std::pin::Pin;
 use std::process::{self, Command};
 use std::str;
 use std::sync::LazyLock;
+use std::time::Instant;
 
 pub mod bencher;
 mod etw_parser;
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):

Kobzol:
I wasn't far off, it was ~66s in the last benchmark run. That means that if I made it twice faster, we would save 30 seconds for each benchmark run :thinking: Not sure if it is worth it, tbh, because I had to essentially copy out the code for getting the artifact sizes out of analyzeme.

PR #2190, opened 2025-07-06
switch sha1 signature check to sha256

klensy:
https://github.com/rust-lang/rustc-perf/pull/2184#issuecomment-3039581010

Headers check changed according to https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers


Kobzol (approved):
Thank you! :)

PR #2192, opened 2025-07-07
Feat; add releases to the benchmark_request queue

Jamesbarford:
- Adds releases to the `benchmark_request` queue

### DIFF
--- a/site/src/job_queue.rs
+++ b/site/src/job_queue.rs
@@ -107,18 +107,23 @@ async fn create_benchmark_request_releases(
     // TODO; delete at some point in the future
     let cutoff: chrono::DateTime<Utc> = chrono::DateTime::from_str("2025-06-01T00:00:00.000Z")?;
 
-    for release_string in releases.lines().rev().take(20) {
-        if let Some((name, date_time)) = parse_release_string(release_string) {
-            if date_time >= cutoff {
-                let release_request = BenchmarkRequest::create_release(
-                    &name,
-                    date_time,
-                    BenchmarkRequestStatus::ArtifactsReady,
-                    "",
-                    "",
-                );
-                conn.insert_benchmark_request(&release_request).await;
-            }
+    let releases: Vec<_> = releases
+        .lines()
+        .rev()
+        .filter_map(parse_release_string)
+        .take(20)
+        .collect();
+
+    for (name, date_time) in releases {
+        if date_time >= cutoff {
+            let release_request = BenchmarkRequest::create_release(
+                &name,
+                date_time,
+                BenchmarkRequestStatus::ArtifactsReady,
+                "",
+                "",
+            );
+            conn.insert_benchmark_request(&release_request).await;
         }
     }
     Ok(())


site/src/job_queue.rs
```diff
@@ -36,6 +41,93 @@ async fn create_benchmark_request_master_commits(
...
 }
 
+/// Parses strings in the following formats extracting the Date & release tag
+/// `static.rust-lang.org/dist/2016-05-24/channel-rust-1.9.0.toml`
+/// `static.rust-lang.org/dist/2016-05-31/channel-rust-nightly.toml`
+/// `static.rust-lang.org/dist/2016-06-01/channel-rust-beta.toml`
+/// `static.rust-lang.org/dist/2025-06-26/channel-rust-1.89-beta.toml`
+/// `static.rust-lang.org/dist/2025-06-26/channel-rust-1.89.0-beta.toml`
+/// `static.rust-lang.org/dist/2025-06-26/channel-rust-1.89.0-beta.2.toml`
+fn parse_release_string(url: &str) -> anyhow::Result<Option<(String, DateTime<Utc>)>> {
```
Kobzol: Hmm, we already have `parse_published_artifact_tag` and `missing_published_artifacts`. But it's true that we didn't parse the release dates before, which was a shame, as they are missing in the DB, and it was using the DB index with artifact SHAs for checking what was missing. So I suppose that this temporary duplication is fine.

site/src/job_queue.rs
```diff
@@ -36,6 +41,93 @@ async fn create_benchmark_request_master_commits(
...
+    static VERSION_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"(\d+\.\d+\.\d+)").unwrap());
+
+    // Grab ".../YYYY-MM-DD/FILE.toml" components with Path helpers.
+    let file = Path::new(url)
+        .file_name()
+        .and_then(|n| n.to_str())
+        .ok_or_else(|| anyhow::anyhow!("URL lacks a file name"))?;
```
Kobzol:
```suggestion
        .ok_or_else(|| anyhow::anyhow!("URL {url} lacks a file name"))?;
```
Otherwise it's quite hard to debug what was wrong. (same in other `anyhow::anyhow!` calls below).

site/src/job_queue.rs
```diff
@@ -36,6 +41,93 @@ async fn create_benchmark_request_master_commits(
...
+    // * channel-rust-1.75.0-beta.1.toml
+    //
+    // Which should get ignored for now, they're not consumable via rustup yet.
+    if file.contains("beta") && file != "channel-rust-beta.toml" {
+        return Ok(None);
+    }
+
+    // Parse the YYYY-MM-DD segment and stamp it with *current* UTC time.
+    let naive = NaiveDate::parse_from_str(date_str, "%Y-%m-%d")?;
+    let now = Utc::now();
```
Kobzol: I don't think adding timestamps based on when was the artifact parsed is ideal, because then when we call the function multiple times, it will return different results. I would just assign a fixed time to all releases, e.g. the beginning of a day (`00:00`).

site/src/job_queue.rs
```diff
@@ -36,6 +41,93 @@ async fn create_benchmark_request_master_commits(
...
+/// already in the database
+async fn create_benchmark_request_releases(
+    conn: &dyn database::pool::Connection,
+) -> anyhow::Result<()> {
+    let releases: String = reqwest::get("https://static.rust-lang.org/manifests.txt")
+        .await?
+        .text()
+        .await?;
+    // TODO; delete at some point in the future
+    let cutoff: chrono::DateTime<Utc> = chrono::DateTime::from_str("2025-06-01T00:00:00.000Z")?;
```
Kobzol, line 108: The original code loaded 20 most recent artifacts, I think that we can do the same here. You can still keep the cut-off, but we should also have some long-term cap on the number of release artifacts parsed. You can replicate the original code, so iterate through lines in reverse order, try to parse release artifacts, and take the first 20 of them.

site/src/job_queue.rs
Kobzol: I think that returning just `Option` is enough, if it can be parsed as a release, we just don't take it into account. Same as the original code.

site/src/job_queue.rs
Jamesbarford: I've removed all the `anyhow` errors as per you other comment and only return an `Option`; 60d7321b02f5c7f2944ef477924bb3fb870f0efa

site/src/job_queue.rs
Jamesbarford: 60d7321b02f5c7f2944ef477924bb3fb870f0efa

site/src/job_queue.rs
```diff
@@ -111,8 +107,8 @@ async fn create_benchmark_request_releases(
     // TODO; delete at some point in the future
     let cutoff: chrono::DateTime<Utc> = chrono::DateTime::from_str("2025-06-01T00:00:00.000Z")?;
 
-    for release_string in releases.lines() {
-        if let Some((name, date_time)) = parse_release_string(release_string)? {
+    for release_string in releases.lines().rev().take(20) {
```
Kobzol: This is a bit different than we need - here you load the last 20 lines, but we need the last 20 release artifacts (imagine e.g. that we had 20 lines with mostly nightly releases, then we'd take only a few actual release artifacts). Check how the `parse_published...` function was used, we can just copy paste it.

Jamesbarford: My previous method, I think, was implementing what is there currently. The current function parses _all_ of the list then makes a decision if I understand correctly;

```rust
// site/src/load.rs L:236

        let artifacts: Vec<_> = artifact_list
            .lines()
            .rev()
            .filter_map(parse_published_artifact_tag) // Will parse every item in the list
            .take(20)
            .filter(|artifact| {
                !tested_artifacts.contains(artifact.as_str())
                    && !in_progress_tagged_artifacts.contains(artifact.as_str())
            })
            .collect();
```

So I think what I was doing previously was fine?

Jamesbarford: 069f182c3f8a60eea7ddaf10df348db241bba761 I've changed it :+1:

Kobzol: Let's say that we take the last 3 (instead of 20) releases, to simplify this. If we have this input:
```
nightly-A
1.10.0
nightly-B
1.11.0
1.12.0
nightly-C
nightly-D
```
then
`for release_string in releases.lines().rev().take(3)` will iterate over `nightly-D`, `nightly-C` and `1.12.0`, and extract only `1.12.0`.

What we want instead is to parse `1.12.0`, `1.11.0` and `1.10.0`.

```rust
artifact_list
            .lines()
            .rev()
            .filter_map(parse_published_artifact_tag) // Will parse every item in the list
            .take(3)
```
goes through the lines in reverse order, keeps only the release artifact lines in the iterator chain, and then takes the first three. That's exactly what we want.

The `.collect()` is not needed, but otherwise the new logic is fine :+1:

Kobzol (approved):
Thanks! I tested it locally, and it seems that it works fine. I also noticed that one of the benchmark requests is already marked as being in progress.. we should remove this logic from the code, even though it's WIP, it will mess up with out invariants later, because we haven't enqueued any jobs yet, so we might have to fix up the DB manually.

PR #2196, opened 2025-07-08
Chore; make `tag` optional for Try benchmark requests

Jamesbarford:
- Removes database table `benchmark_request` constraint for `tag` to be `NOT NULL`
- Change allocations for `.tag()` to return `Option<&str>`

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -985,10 +985,7 @@ impl BenchmarkRequest {
 
     pub fn parent_sha(&self) -> Option<&str> {
         match &self.commit_type {
-            BenchmarkRequestType::Try { parent_sha, .. } => match parent_sha {
-                Some(parent_sha) => Some(parent_sha),
-                _ => None,
-            },
+            BenchmarkRequestType::Try { parent_sha, .. } => parent_sha.as_deref(),
             BenchmarkRequestType::Master { parent_sha, .. } => Some(parent_sha),
             BenchmarkRequestType::Release { tag: _ } => None,
         }


database/src/pool/postgres.rs
```diff
@@ -301,6 +301,11 @@ static MIGRATIONS: &[&str] = &[
     );
     CREATE INDEX IF NOT EXISTS benchmark_request_status_idx on benchmark_request (status) WHERE status != 'completed';
     "#,
+    // Remove that tag cannot be NULL
+    r#"ALTER TABLE benchmark_request ALTER COLUMN tag DROP NOT NULL;"#,
+    // Prevent multiple try commits without a `sha` and the same `pr` number
+    // being added to the table
+    r#"CREATE UNIQUE INDEX benchmark_request_pr_commit_type_idx ON benchmark_request (pr, commit_type) WHERE tag IS NULL;"#,
```
Kobzol: Hmm. We also want to prevent concurrent perf. runs if one try build already has artifacts, but the other one doesn't.

I wonder how it works currently. Because it is definitely possible to do e.g. two benchmark requests on two different try builds on the same PR (we need to keep that working). Could you please check how it is handled now?

Jamesbarford: I might be getting the wrong end of the stick here. However I believe that does work; I broke the test for building the queue initially by having the index as;

```sql
CREATE UNIQUE INDEX benchmark_request_pr_commit_type_idx ON benchmark_request (pr, commit_type);
```
Which meant that we couldn't have a `Try` with a status as `in_progress` and attempt to queue another run.

Whereas with;
 ```sql
CREATE UNIQUE INDEX benchmark_request_pr_commit_type_idx ON benchmark_request (pr, commit_type) WHERE tag IS NULL;
```
Means that we can have as many `Try`'s as we like for a PR provided there is a defined `tag` (sha specifically for a `Try`).

Comment (Jamesbarford) [reply database/src/pool/postgres.rs]: I'm not entirely sure what you're referring too however we could have the WHERE clause as;

```sql
WHERE tag IS NULL OR status IN ('in_progress')
```

Which would prevent another `try` being queued for the same Pr if;
- The pr matches and there is no tag
OR
- The pr matches and there is a `Try` which currently has it's status as `in_progress`. Which effectively blocks any `Try`'s from being queued for a PR if there is one currently in progress.

Kobzol: > Which meant that we couldn't have a Try with a status as in_progress and attempt to queue another run.

Hmm. I kind of assumed that this is exactly what we want. The previous code seemingly kind of allowed something like this, but I think it would create a mess in the DB if it actually happened. And having multiple parallel perf. runs on a PR is quite confusing, when you receive the results comment, it will be hard to keep track to which commit it is attached to.

That's also why we don't allow parallel try builds on single PR.

However, we also can't just use only `(pr, commit_type)`, because that would completely prevent doing multiple perf. runs on the same PR.

I think that `WHERE status != 'completed'` might be closest to what we want. We don't allow parallel try builds, but if a previous try build was completed, we now ignore it. What do you think?

Jamesbarford: That would mean that we would not be able to run another benchmark on a "job/benchmark" that has completed?

Would that also mean that if someone is working on a PR and wants to run a couple of benchmarks, say test changes between commits, they would not be able to until they opened a new PR?

Kobzol: I'll try to describe the use-cases.

What we want to work:
- Start a try build + queue a try benchmark request, wait for the try build to complete, perf run the try benchmark request. Then start + queue another try build, etc. (this is the most common use-case).
- Create a benchmark request for an already completed try build (that wasn't queued before). This can be repeated, as long as no benchmark request is currently waiting/ready/in-progress.

What we do not want to work:
- Queue a try benchmark request while another try request on the same PR is already waiting/ready/in-progress.
- Create a benchmark request for an already completed try build while another try request on the same PR is already waiting/ready/in-progress.

I think that if we only had the unique index `WHERE status != 'completed'`, this would be upheld, right?

Jamesbarford: Yes that makes sense as `tag IS NULL` is better described by the status of `WaitingForArtifacts`. Which your suggestion of `WHERE status != 'completed'` would cover :+1:

Jamesbarford: I've updated the index and now the test too; 316e314e2bb34d650a7b15c723dd531361134ace

site/src/job_queue.rs
```diff
@@ -182,7 +182,9 @@ fn sort_benchmark_requests(done: &HashSet<String>, request_queue: &mut [Benchmar
             )
         });
         for c in level {
-            done.insert(c.tag().to_string());
+            if let Some(tag) = c.tag() {
```
Kobzol: Let's `expect` the tag here (and add a comment that explains that waiting for artifacts PRs shouldn't ever appear here). If a request with `NULL` SHA ever gets here, we have a bug in the code.

Kobzol (commented):
Thanks! I tried it locally, the migration was applied cleanly, it allows inserting NULL SHAs, multiple non-completed try builds cannot exist at the same time, but multiple release builds can (so `NULL` in the `pr` column is treated as distinct values for the unique index). So it seems to work as we wanted it to.

site/src/job_queue.rs
Jamesbarford: 89b5e88e378defcd1ac74bd3468e72af840130da

PR #2197, opened 2025-07-09
Feat; queuing try commits

Jamesbarford:


database/src/pool/postgres.rs
```diff
@@ -1525,6 +1525,42 @@ where
...
         Ok(())
     }
+
+    async fn attach_shas_to_try_benchmark_request(
+        &mut self,
+        pr: u32,
+        sha: &str,
+        parent_sha: &str,
+    ) -> anyhow::Result<()> {
+        let tx = self
```
Kobzol: Why the transaction? We only run a single SQL statement, that is essentially a mini-transaction anyway, right?

database/src/pool/postgres.rs
```diff
@@ -1525,6 +1525,42 @@ where
...
+        tx.execute(
+            "UPDATE
+                benchmark_request
+            SET
+                tag = $1,
+                parent_sha = $2,
+                status = $3
+            WHERE
+                pr = $4
+                AND commit_type = 'try';",
```
```suggestion
                AND commit_type = 'try'
                AND tag IS NULL;",
```
We need to only modify try builds on this PR that are not completed and don't have a SHA already. We might also add `WHERE status = "waiting-for-artifscts"` for good measure.

Please add a test for this, that when you have a completed try request in the DB, and then you create another one waiting for artifacts and attach SHA to it, it won't overwrite the SHA of the previous one.

site/src/github.rs
```diff
@@ -233,6 +234,22 @@ pub async fn rollup_pr_number(
...
+    pr: u32,
+    sha: &str,
+    parent_sha: &str,
+) {
+    if run_new_queue() {
+        if let Err(e) = conn
+            .attach_shas_to_try_benchmark_request(pr, sha, parent_sha)
+            .await
+        {
+            log::error!("Cron job failed to execute {}", e);
```
Kobzol: Error log isn't accurate.

site/src/request_handlers/github.rs
```diff
@@ -72,6 +74,29 @@ async fn handle_issue(
...
+        let try_request = BenchmarkRequest::create_try(
+            None,
+            None,
+            pr,
+            chrono::Utc::now(),
+            BenchmarkRequestStatus::WaitingForArtifacts,
+            backends,
+            "",
+        );
+        conn.insert_benchmark_request(&try_request).await;
```
Kobzol, line 96: I wonder if we should return an error here, so that we can e.g. tell to the user that a request already exists, so they can't queue another one, otherwise it will be essentially a silent failure. But we can do that in a future PR.

Jamesbarford: Yeah I reckon so, happy to do it in a follow up PR

database/src/pool/postgres.rs
Jamesbarford: I've used the enum to lower the number of hardcoded strings, but yes makes sense

site/src/github.rs
```diff
@@ -274,15 +274,10 @@ pub async fn enqueue_shas(
...
-        attach_shas_to_try_benchmark_request(
-            &mut *conn,
-            pr_number,
-            &try_commit.sha,
-            &try_commit.sha,
-        )
-        .await;
+        let conn = ctxt.conn().await;
+
+        attach_shas_to_try_benchmark_request(&*conn, pr_number, &try_commit.sha, &try_commit.sha)
```
Kobzol: Wrong SHA is being passed here for the parent.

database/src/pool.rs
```diff
@@ -579,4 +579,68 @@ mod tests {
...
+                ])
+                .await
+                .unwrap();
+
+            assert_eq!(requests.len(), 2);
+            let completed_try = requests
+                .iter()
+                .find(|req| req.status == BenchmarkRequestStatus::Completed);
+            assert!(completed_try.is_some());
+            assert_eq!(completed_try.unwrap().pr(), Some(&pr));
```
Kobzol, line 634: Please assert that the SHA of the completed build hasn't changed (but the SHA of the second one did), that is the problematic situation that would be broken without the `WHERE` in the SQL query.

Kobzol (approved):
Thanks! Feel free to merge once CI is green.

PR #2198, opened 2025-07-09
Chore; Missed database error handling & small refactor

Jamesbarford:
- Makes `insert_benchmark_request` return an error
- Deleted some un-needed methods
- Refactor the `BenchmarkRequestStatus` so we only have the magic strings written once

### DIFF
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -181,7 +181,7 @@ pub trait Connection: Send + Sync {
     async fn purge_artifact(&self, aid: &ArtifactId);
 
     /// Add an item to the `benchmark_requests`, if the `benchmark_request`
-    /// exists it will be ignored
+    /// exists an Error will be returned
     async fn insert_benchmark_request(
         &self,
         benchmark_request: &BenchmarkRequest,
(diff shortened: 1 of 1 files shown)


database/src/pool.rs
```diff
@@ -182,7 +182,10 @@ pub trait Connection: Send + Sync {
 
     /// Add an item to the `benchmark_requests`, if the `benchmark_request`
```
Kobzol, line 183: The comment is inaccurate now.

database/src/pool.rs
```diff
@@ -436,12 +439,23 @@ mod tests {
             );
 
             let db = db.connection().await;
-            db.insert_benchmark_request(&master_benchmark_request).await;
-            db.insert_benchmark_request(&try_benchmark_request).await;
-            db.insert_benchmark_request(&release_benchmark_request)
-                .await;
+            assert!(db
```
Kobzol: Just `unwrap()`, it's shorter and will actually show the failure (`Err` value) in the panic, unlike this assert.

Kobzol (approved):

database/src/pool.rs
Jamesbarford: Sure, for where I'm testing the error I've left it as `assert!(<>.is_err())`

PR #2200, opened 2025-07-10
Feat; job_queue table definition & mark a request as complete if all jobs are finished

Jamesbarford:
- Creates the `job_queue` table.
- Inserts into the `job_queue`.
- Marks a `benchmark_request` as complete if all jobs associated with the request have the status of `success` or `failure`.

I've omitted `job_splitting` as this PR is modestly big. However, bar marking things as complete, the PR is fairly mechanical.

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -321,11 +321,28 @@ collector which benchmarks it should execute. The jobs will be kept in the
 table for ~30 days after being completed, so that we can quickly figure out
 what master parent jobs we need to backfill when handling try builds.
 
-```
-psql# SELECT * FROM job_queue limit 1;
+Columns:
 
-id   request_id   target    backend   benchmark_set  collector_id    created_at                   started_at                   completed_at                 status     retry   error 
----  -----------  --------  --------  -------------  --------------  ---------------------------  ---------------------------  ---------------------------  ---------  ------  --------
-23            7   AArch64   llvm      5              collector-1     2025-07-10 09:00:00.123+00   2025-07-10 09:05:02.456+00   2025-07-10 09:15:17.890+00   complete       0  
-```
->>>>>>> d0ed82ff (Feat; job_queue table definition & mark a request as complete if all jobs are finished)
+- **id** (`bigint` / `serial`): Primary-key identifier for the job row;
+  auto-increments with each new job.
+- **request_id** (`bigint`): References the parent benchmark request that
+  spawned this job.
+- **target** (`text NOT NULL`): Hardware/ISA the benchmarks must run on
+  (e.g. AArch64, x86_64).
+- **backend** (`text NOT NULL`): Code generation backend the collector should
+  test (e.g. llvm, cranelift).
+- **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
+  execute.
+- **collector_id** (`text`): Id of the collector that claimed the job
+  (populated once the job is started).
+- **created_at** (`timestamptz NOT NULL`): Datetime when the job was queued.
+- **started_at** (`timestamptz`): Datetime when the collector actually began
+  running the benchmarks; NULL until the job is claimed.
+- **completed_at** (`timestampt`): Datetime when the collector finished
+  (successfully or otherwise); used to purge rows after ~30 days.
+- **status** (`text NOT NULL`): Current job state. `queued`, `in_progress`,
+  `success`, or `failure`.
+- **retry** (`int NOT NULL`): Number of times the job has been re-queued after
+  a failure; 0 on the first attempt.
+- **error** (`text`): Optional error message or stack trace from the last
+  failed run; NULL when the job succeeded.
--- a/database/schema.md
+++ b/database/schema.md
@@ -313,3 +313,19 @@ Columns:
   execute.
 * **is_active** (`boolean NOT NULL`): For controlling whether the collector is
   active for use. Useful for adding/removing collectors.
+
+### job_queue
+
+This table stores ephemeral benchmark jobs, which specifically tell the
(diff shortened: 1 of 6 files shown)


database/schema.md
```diff
@@ -296,3 +295,18 @@ Columns:
...
 * **profiles** (`text NOT NULL`): Comma-separated list of profiles to benchmark. If empty, the default set of profiles will be benchmarked.
+
+### job_queue
+
+This table stores ephemeral benchmark jobs, which specifically tell the
+collector which benchmarks it should execute. The jobs will be kept in the
+table for ~30 days after being completed, so that we can quickly figure out
+what master parent jobs we need to backfill when handling try builds.
+
+```
```
Kobzol: Can you please use the list-based field description that I used above? I think it's better to describe the fields in more detail. While these tables look cool, they are also a big pain to modify, and don't contain the field type and some possible additional invariants or comments.

database/src/lib.rs
```diff
@@ -1009,4 +1013,73 @@ impl BenchmarkRequestIndex {
     pub fn completed_requests(&self) -> &HashSet<String> {
         &self.completed
     }
+
+    pub fn add_tag(&mut self, tag: &str) {
+        self.completed.insert(tag.to_string());
```
Kobzol, line 1057: This should also add the tag to `all`, if we're trying to keep the cache consistent.

database/src/lib.rs
```diff
@@ -1009,4 +1013,73 @@ impl BenchmarkRequestIndex {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct BenchmarkJob {
+    target: Target,
+    backend: CodegenBackend,
+    benchmark_set: u32,
```
Kobzol: Let's create a newtype for the set, i.e. `struct BenchmarkSet(u32)`.

database/src/lib.rs
```diff
@@ -1009,4 +1013,73 @@ impl BenchmarkRequestIndex {
...
+    collector_id: String,
+    created_at: Option<DateTime<Utc>>,
```
Kobzol: `created_at` should be non-optional, I think? It always has to be set if the job exists.

database/src/pool.rs
```diff
@@ -210,6 +210,26 @@ pub trait Connection: Send + Sync {
         sha: &str,
         parent_sha: &str,
     ) -> anyhow::Result<()>;
+
+    /// Try and mark the benchmark_request as completed. Will return `true` if
+    /// it has been marked as completed else `false` meaning there was no change
+    async fn try_mark_benchmark_request_as_completed(
```
```suggestion
    async fn mark_benchmark_request_as_completed(
```
I think that the word "try" is too overloaded in this context.

database/src/pool.rs
```diff
@@ -210,6 +210,26 @@ pub trait Connection: Send + Sync {
...
+        &self,
+        benchmark_request: &mut BenchmarkRequest,
+    ) -> anyhow::Result<bool>;
+
+    /// Given a benchmark request get the id
+    async fn get_benchmark_request_id(
```
Kobzol: Tags are unique, let's just use them as the identifier, to avoid the usual ORM issue of "I have a domain struct represented in the DB, how do I attach the ID to it, when it can be None before the struct is inserted into the DB" etc. etc. So we won't need this function, and we can simply use tags when inserting benchmark jobs and marking them as completed.

database/src/pool/postgres.rs
```diff
@@ -309,6 +309,28 @@ static MIGRATIONS: &[&str] = &[
...
     // being added to the table
     r#"CREATE UNIQUE INDEX benchmark_request_pr_commit_type_idx ON benchmark_request (pr, commit_type) WHERE status != 'completed';"#,
+    r#"
+    CREATE TABLE IF NOT EXISTS job_queue (
+        id            SERIAL PRIMARY KEY,
+        request_id    INTEGER NOT NULL,
+        target        TEXT NOT NULL,
+        backend       TEXT NOT NULL,
+        benchmark_set INTEGER NOT NULL,
+        collector_id  TEXT NOT NULL,
```
Kobzol: We finally got to a place where we need collectors, let's merge https://github.com/rust-lang/rustc-perf/pull/2157 first then, so that we can use a proper foreign key here.

Kobzol (commented):
Did a quick scan and left a bunch of comments, but it would likely be best to discuss this at our call.

database/src/pool/postgres.rs
Jamesbarford: Good idea, I've updated that PR

database/src/pool.rs
Jamesbarford: Yup that makes sense to me :+1:. Though still useful at the database level to map to the `benchmark_request` but is more of an internal implementation detail

database/src/pool/postgres.rs
```diff
@@ -1593,6 +1615,98 @@ where
...
+                WHERE
+                    benchmark_request.tag = $2
+                AND benchmark_request.commit_type = $3
+                AND benchmark_request.status = $4
+                AND NOT EXISTS (
+                    SELECT
+                        1
+                    FROM
+                        job_queue
+                    WHERE job_queue.request_id = benchmark_request.id
```
Kobzol, line 1747: Once we also add the parent benchmark tag here, could you please run `EXPLAIN` to make sure that the **jobs** (not requests) are looked up by index? Also would be nice to check if requests are accessed by index.

Jamesbarford:
Superseded by; https://github.com/rust-lang/rustc-perf/pull/2214

PR #2201, opened 2025-07-10
Refactor benchmark requests

Kobzol:
I went through the current benchmark request code and did a bunch of changes, mostly in order to encode more invariants into the type system, and to only use public API for dealing with these requests. This means that tests now also have to be using the public API, instead of constructing custom state arbitrarily.

Best reviewed commit-by-commit. Sorry, this will probably conflict a lot with your next changes, it took me a while to get through this.


Jamesbarford (changes requested):
This looks good! Nice work! While you're making these changes we could also add `benchmark_request` to schema.md?

Kobzol:
Good point, added.

PR #2203, opened 2025-07-19
collector: use `patch` instead of `git apply`

ognevny:
so it'll be possible to run collector inside of tarball

### DIFF
--- a/collector/src/compile/benchmark/patch.rs
+++ b/collector/src/compile/benchmark/patch.rs
@@ -60,8 +60,19 @@ impl Patch {
     pub fn apply(&self, dir: &Path) -> anyhow::Result<()> {
         log::debug!("applying {} to {:?}", self.name, dir);
 
-        let mut cmd = Command::new("git");
-        cmd.current_dir(dir).args(["apply"]).arg(&*self.path);
+        // It's harder to use patch(1) in pure Windows MSVC toolchain
+        let mut cmd = if cfg!(all(windows, target_env = "msvc")) {
+            let mut command = Command::new("git");
+            command.current_dir(dir).args(["apply"]).arg(&*self.path);
+            command
+        } else {
+            let mut command = Command::new("patch");
+            command
+                .current_dir(dir)
+                .args(["-p1", "-i"])
+                .arg(&*self.path);
+            command
+        };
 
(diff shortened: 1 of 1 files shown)


collector/src/compile/benchmark/patch.rs
```diff
@@ -60,8 +60,8 @@ impl Patch {
     pub fn apply(&self, dir: &Path) -> anyhow::Result<()> {
         log::debug!("applying {} to {:?}", self.name, dir);
 
-        let mut cmd = Command::new("git");
-        cmd.current_dir(dir).args(["apply"]).arg(&*self.path);
+        let mut cmd = Command::new("patch");
```
Kobzol: This probably won't work on Windows, right? Could you talk a bit more about the use-case that you have?

ognevny: on Windows there should be way to invoke patch (e.g. from MSYS2 or Cygwin)

just download dist src tarball and try to run opt-dist in it. collector will fail even if git binary is available because in tarball there is no git repos

Kobzol: And you need it to run opt-dist on a tarball? I wonder how other people do it, there are definitely a few distro people who use opt-dist. Maybe they just use git or have some local patches.

> on Windows there should be way to invoke patch (e.g. from MSYS2 or Cygwin)

Currently rustc-perf should hopefully work on "native" Windows (MSVC), so it should stay that way I'm fine with special casing the logic for Windows/MSVC using e.g. `#[cfg]`.

ognevny: > And you need it to run opt-dist on a tarball? I wonder how other people do it, there are definitely a few distro people who use opt-dist. Maybe they just use git or have some local patches.

MSYS2 and many linux distros build Rust from dist tarball, not git repository

ognevny: tried cfg workaround

Kobzol (approved):
Thanks. It would be best if we applied the patch in Rust, but I haven't found any well-established crates for that.

ognevny:
> but I haven't found any well-established crates for that.

I'll try to integrate `diffy` here

edit: oh, it's not so easy

PR #2204, opened 2025-07-21
Implement benchmark comparison TUI using Ratatui

Kobzol:
I finally found some time and motivation to finish https://github.com/rust-lang/rustc-perf/pull/1991. This should be much easier to use than the previous simple TUI.

cmp-ratatui

CC @orhun

Closes: https://github.com/rust-lang/rustc-perf/pull/1991


JonathanBrouwer (approved):
I used this code to debug performance problems of my rustc branch today, works very well

lqd (approved):
Great job, no notes.

PR #2207, opened 2025-07-23
Feat; Split request job into benchmark jobs

Jamesbarford:
Adds the database table `job_queue` along with creating a `BenchmarkJob` from a `BenchmarkRequest`.

This does not include enqueuing a parent if there is a missing backend/profile.

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -1092,6 +1092,12 @@ impl fmt::Display for BenchmarkJobStatus {
 #[derive(Debug, Clone, PartialEq)]
 pub struct BenchmarkSet(u32);
 
+/// A single unit of work generated from a benchmark request. Split by profiles
+/// and backends
+///
+/// Each request is split into several `BenchmarkJob`s. Collectors poll the
+/// queue and claim a job only when its `benchmark_set` matches one of the sets
+/// they are responsible for.
 #[derive(Debug, Clone, PartialEq)]
 pub struct BenchmarkJob {
     target: Target,
@@ -1102,49 +1108,4 @@ pub struct BenchmarkJob {
     created_at: DateTime<Utc>,
     status: BenchmarkJobStatus,
     retry: u32,
-}
-
-impl BenchmarkJob {
-    pub fn create_queued(
-        target: Target,
-        backend: CodegenBackend,
-        profile: Profile,
-        request_tag: &str,
-        benchmark_set: u32,
-    ) -> Self {
-        BenchmarkJob {
-            target,
-            backend,
-            profile,
-            created_at: Utc::now(),
-            request_tag: request_tag.to_string(),
-            benchmark_set: BenchmarkSet(benchmark_set),
-            status: BenchmarkJobStatus::Queued,
-            retry: 0,
-        }
-    }
-
-    pub fn request_tag(&self) -> &str {
-        &self.request_tag
-    }
-
-    pub fn target(&self) -> Target {
-        self.target
-    }
-
-    pub fn backend(&self) -> CodegenBackend {
-        self.backend
-    }
-
-    pub fn profile(&self) -> Profile {
-        self.profile
-    }
-
-    pub fn benchmark_set(&self) -> u32 {
-        self.benchmark_set.0
-    }
-
-    pub fn status(&self) -> &BenchmarkJobStatus {
-        &self.status
-    }
 }
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -1052,9 +1052,14 @@ impl BenchmarkRequestIndex {
 #[derive(Debug, Clone, PartialEq)]
 pub enum BenchmarkJobStatus {
     Queued,
-    InProgress,
-    Success,
-    Failure,
+    InProgress {
+        started_at: DateTime<Utc>,
+    },
+    Completed {
+        started_at: DateTime<Utc>,
+        completed_at: DateTime<Utc>,
+        success: bool,
+    },
 }
 
 const BENCHMARK_JOB_STATUS_QUEUED_STR: &str = "queued";
@@ -1066,9 +1071,14 @@ impl BenchmarkJobStatus {
     pub fn as_str(&self) -> &str {
         match self {
             BenchmarkJobStatus::Queued => BENCHMARK_JOB_STATUS_QUEUED_STR,
-            BenchmarkJobStatus::InProgress => BENCHMARK_JOB_STATUS_IN_PROGRESS_STR,
-            BenchmarkJobStatus::Success => BENCHMARK_JOB_STATUS_SUCCESS_STR,
-            BenchmarkJobStatus::Failure => BENCHMARK_JOB_STATUS_FAILURE_STR,
+            BenchmarkJobStatus::InProgress { .. } => BENCHMARK_JOB_STATUS_IN_PROGRESS_STR,
+            BenchmarkJobStatus::Completed { success, .. } => {
+                if *success {
+                    BENCHMARK_JOB_STATUS_SUCCESS_STR
+                } else {
+                    BENCHMARK_JOB_STATUS_FAILURE_STR
+                }
+            }
         }
     }
 }
@@ -1090,32 +1100,26 @@ pub struct BenchmarkJob {
     request_tag: String,
     benchmark_set: BenchmarkSet,
     created_at: DateTime<Utc>,
-    started_at: Option<DateTime<Utc>>,
-    completed_at: Option<DateTime<Utc>>,
     status: BenchmarkJobStatus,
     retry: u32,
 }
 
 impl BenchmarkJob {
-    pub fn new(
+    pub fn create_queued(
         target: Target,
         backend: CodegenBackend,
         profile: Profile,
         request_tag: &str,
         benchmark_set: u32,
-        created_at: DateTime<Utc>,
-        status: BenchmarkJobStatus,
     ) -> Self {
         BenchmarkJob {
             target,
             backend,
             profile,
+            created_at: Utc::now(),
             request_tag: request_tag.to_string(),
             benchmark_set: BenchmarkSet(benchmark_set),
-            created_at,
-            started_at: None,
-            completed_at: None,
-            status,
+            status: BenchmarkJobStatus::Queued,
             retry: 0,
         }
     }
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -227,6 +227,11 @@ impl Profile {
             Profile::Clippy => "clippy",
         }
     }
+
+    /// Set of default profiles that should be benchmarked for a master/try artifact.
+    pub fn default_profiles() -> Vec<Self> {
+        vec![Profile::Check, Profile::Debug, Profile::Doc, Profile::Opt]
+    }
 }
 
 impl std::str::FromStr for Profile {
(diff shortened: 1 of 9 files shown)


database/src/lib.rs
```diff
@@ -227,6 +227,11 @@ impl Profile {
             Profile::Clippy => "clippy",
         }
     }
+
+    /// Set of default profiles that should be benchmarked for a master/try artifact.
+    pub fn default_profiles() -> Vec<Self> {
```
Kobzol, line 232: Not that I expect these to change often, but we now have the logic for deciding what is the default duplicated in the collector `Profile` and the database `Profile`. Right now, the collector one is more important, because that's what decides what is benchmarked. In the future, the website will be deciding what gets benchmarked, so the database one will be more important. It would be nice to have this list only at a single place (I'd say the collector for now), and then load it from there and convert the collector profiles to the database profiles.

It would be nice to have logic for converting from the collector data structures to the database data structures somewhere, as it would be needed both for what I described above, and also for https://github.com/rust-lang/rustc-perf/pull/2161. Since the database doesn't depend on the collector, these conversions should live in the collector crate.

database/src/lib.rs
```diff
@@ -1010,3 +1048,99 @@ impl BenchmarkRequestIndex {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct BenchmarkJob {
+    target: Target,
+    backend: CodegenBackend,
+    profile: Profile,
+    request_tag: String,
+    benchmark_set: BenchmarkSet,
+    created_at: DateTime<Utc>,
+    started_at: Option<DateTime<Utc>>,
+    completed_at: Option<DateTime<Utc>>,
```
Kobzol: Similarly to the request, we can represent this without the options, e.g. like this:
```rust
enum BenchmarkJobStatus {
   Queued,
   InProgress { started_at: DateTime<Utc> },
   Completed {
       started_at: DateTime<Utc>,
       completed_at: DateTime<Utc>,
       success: bool
   }
}
```

database/src/pool/postgres.rs
```diff
@@ -342,7 +342,9 @@ static MIGRATIONS: &[&str] = &[
         CONSTRAINT job_queue_request_fk
             FOREIGN KEY (request_tag)
             REFERENCES benchmark_request(tag)
-            ON DELETE CASCADE
+            ON DELETE CASCADE,
+
+       CONSTRAINT job_queue_unique UNIQUE(request_tag, target, backend, profile)
```
Kobzol: We also need to include the benchmark set here, otherwise we couldn't split the jobs across multiple collectors.

database/src/pool/postgres.rs
```diff
@@ -1608,6 +1630,34 @@ where
             .collect();
         Ok(requests)
     }
+
+    async fn enqueue_benchmark_job(&self, benchmark_job: &BenchmarkJob) -> anyhow::Result<()> {
```
Kobzol: This is always tricky with these we-have-ORM-at-home situations, but it shouldn't really be possible to ever insert a job that is not in the queued state into the DB. I wonder if we should just pass the individual fields arguments (tag, target, backend, profile, benchmark_set) to this function, or wrap them in some other struct. That would also allow us to remove the `BenchmarkJob::new` constructor.

database/src/pool/postgres.rs
```diff
@@ -324,6 +324,28 @@ static MIGRATIONS: &[&str] = &[
...
     "#,
+    r#"
+    CREATE TABLE IF NOT EXISTS job_queue (
+        id            SERIAL PRIMARY KEY,
+        request_tag   TEXT NOT NULL,
+        target        TEXT NOT NULL,
+        backend       TEXT NOT NULL,
+        profile       TEXT NOT NULL,
+        benchmark_set INTEGER NOT NULL,
+        collector_id  TEXT,
```
Kobzol: `collector_id` could be an integer that references the primary key of the `collector_config` table, I guess?

site/src/job_queue/mod.rs
```diff
@@ -174,12 +178,63 @@ pub async fn build_queue(
...
+    let mut jobs = vec![];
+
+    // Target x benchmark_set x backend x profile -> BenchmarkJob
+    for target in Target::all() {
+        for benchmark_set in 0..benchmark_set_count(
+            collector::compile::benchmark::target::Target::from_db_target(&target),
+        ) {
+            for backend in backends.iter() {
+                for profile in profiles.iter() {
+                    let job = BenchmarkJob::new(
```
Kobzol: Similar to benchmark requests, I'd like to only use public API to interact with jobs, and avoid giving everyone the option to create invalid states. So ideally the `new` function shouldn't take status and `created_at`; it should automatically set `created_at` to now (unless we *really* need to set the time in tests), and it should automatically set the status to queued.

The `database` crate should be the only one who will be able to create the job from arbitrary state, because it needs to do it for deserialization from the DB.

site/src/job_queue/mod.rs
```diff
@@ -174,12 +178,63 @@ pub async fn build_queue(
...
 /// Enqueue the job into the job_queue
 async fn enqueue_next_job(
     conn: &dyn database::pool::Connection,
     index: &mut BenchmarkRequestIndex,
 ) -> anyhow::Result<()> {
-    let _queue = build_queue(conn, index).await?;
+    let queue = build_queue(conn, index).await?;
+    for request in queue {
+        if request.status() != BenchmarkRequestStatus::InProgress {
+            for job in create_benchmark_jobs(&request)? {
```
Kobzol: We need to wrap this code in a transaction.

site/src/job_queue/mod.rs
```diff
@@ -174,12 +178,63 @@ pub async fn build_queue(
...
+                    let job = BenchmarkJob::new(
+                        target,
+                        *backend,
+                        *profile,
+                        benchmark_request.tag().unwrap(),
+                        benchmark_set as u32,
+                        Utc::now(),
+                        BenchmarkJobStatus::Queued,
+                    );
+                    jobs.push(job);
```
Kobzol: Here we can also create the jobs for the parent. It will have the same parameters, just with the parent tag.

database/src/pool/postgres.rs
```diff
@@ -1608,6 +1630,34 @@ where
             .collect();
         Ok(requests)
     }
+
+    async fn enqueue_benchmark_job(&self, benchmark_job: &BenchmarkJob) -> anyhow::Result<()> {
+        self.conn()
+            .execute(
+                r#"
+            INSERT INTO job_queue(
```
Kobzol, line 1659: This will need `ON CONFLICT DO NOTHING`.

database/src/pool/postgres.rs
Jamesbarford: I've removed it in favour of `create_queued` and passing the individual fields as arguments to the function. Though have excluded `status`

database/src/pool/postgres.rs
Jamesbarford: An edge case we need to handle is a job having a collector id which is not in the collector configuration table.

I was thinking a foreign key however we'd then not be able to have a test for the `enqueue_benchmark_job(...)` without also inserting something into the `collector_config` table which would require creating more methods which we possibly won't be using. I think it needs a bit more thought so have left it out for now

database/src/lib.rs
Jamesbarford: We discussed this and have come to the conclusion it can live in the database for now as the collectors eventually will not need to know what `default_profiles` are. Thus we've removed it from the collector

site/src/job_queue/mod.rs
Jamesbarford: Yes I will immediately do that in a follow up. Or are you suggesting something like;

```rust
if let Some(parent_tag) = benchmark_request.parent_tag() {
   // create job
}
```

In which case it seems simple enough to do that here

site/src/job_queue/mod.rs
Jamesbarford: I've hidden this in `BenchmarkJob::create_queued(...)`

database/src/pool/postgres.rs
Kobzol: Having a test only function for adding a collector config doesn't sound so terrible, if needed. But we can just let the `collector_id` be `NULL` in tests, it will only be used for diagnostics for the status page anyway.

site/src/job_queue/mod.rs
Kobzol: Yeah, just this. I don't think there's more to it under the scheme that we have discussed. Just try to also insert the parent jobs. We could do a "micro-optimization" by avoiding the insert if there are no non-default profiles/backends selected, but it doesn't really matter if we try to insert also the parent jobs too. And at least it will be resilient to (the incredibly rare) changes of backend/profile defaults.

site/src/job_queue/mod.rs
```diff
@@ -174,24 +177,84 @@ pub async fn build_queue(
     Ok(queue)
 }
 
+/// From a benchmark_request create all the required jobs
+pub fn create_benchmark_jobs(
```
Kobzol: We can just pass the DB connection to this function and create the jobs immediately. If we go through the trouble of creating the actual job instance here, we might have as well sent it to the database function :laughing: I'd like to avoid having a constructor for `BenchmarkJob` to be publicly available.

site/src/job_queue/mod.rs
```diff
@@ -381,4 +456,30 @@ mod tests {
         })
         .await;
     }
+
+    #[test]
+    fn create_benchmark_jobs_default() {
```
Kobzol: I don't think that this test is super useful, tbh. I'd just call `create_benchmark_jobs` to actually populate the DB, and then use DB functions to load jobs from the DB and assert their properties. We don't have the loading functions yet, so either you can add them in this PR and keep the test, or I'd remove the test and only add it once we can actually check what happened in the database.

Kobzol (commented):
Thanks for the changes, left a few other thoughts.

I'd like to remove `BenchmarkJob::create_queued`, because it creates the wrong incentives - it shouldn't be usable in "normal" code, because everything should go through the DB, and it ideally shouldn't be usable in tests, because then we test something that doesn't actually happen in production - even the tests should go through the DB if possible. This is a bit different from the benchmark request, where it made sense to create it in code and then insert it into the DB, because we had several variations of the request that could exist. But the job is more specialized, and it doesn't make sense to create it outside of the database.

It would also be nice to add at least some brief doc comments in code on top of the job, just to explain what is it for.

site/src/job_queue/mod.rs
Jamesbarford: I've removed the test for now

database/src/pool/postgres.rs
Jamesbarford: 8e8a03aa2571d26acf314373a121776dd9b8688d :+1:

site/src/job_queue/mod.rs
```diff
@@ -197,20 +197,40 @@ pub fn create_benchmark_jobs(
...
                 }
             }
         }
     }
 
-    Ok(jobs)
+    tx.conn()
+        .update_benchmark_request_status(request_tag, BenchmarkRequestStatus::InProgress)
+        .await?;
+    Ok(())
```
Kobzol, line 239: The tx commit got lost :)

Jamesbarford: :fearful: woops, I've put it back; 754f63b3488657f542e1f43d33945ff4a2e17565

Kobzol (approved):
Thanks! I tested this locally and noticed that multiple requests were marked as in progress, because the enqueuing loop didn't end after enqueuing the first request. I fixed that and added more logging, which helped me with debugging locally.

Btw you can run local Postgres DB, run the website with the CRON environment variables and it will actually start creating the master/release artifacts and enqueuing jobs, so it can be easily tested.

PR #2210, opened 2025-07-25
Make missing test case check more granular

Kobzol:
This is a reland of https://github.com/rust-lang/rustc-perf/pull/2161, which was reverted in https://github.com/rust-lang/rustc-perf/pull/2187. I fixed the issue with recomputing incremental doc builds, and also made sure that when something is recomputed, the corresponding step won't be restarted from scratch (`start_time`), to avoid weirdly short steps (now they will be weirdly long instead, but that's good, as it can tell us on the status page that some recomputation happened).

This will be needed soon for the new benchmarking scheme.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1176,19 +1176,14 @@ where
             == 1
     }
     async fn collector_end_step(&self, aid: ArtifactIdNumber, step: &str) {
-        let did_modify = self
-            .conn()
+        self.conn()
             .execute(
                 "update collector_progress set end_time = statement_timestamp() \
-                where aid = $1 and step = $2 and start_time is not null and end_time is null;",
+                where aid = $1 and step = $2 and start_time is not null;",
                 &[&(aid.0 as i32), &step],
             )
             .await
-            .unwrap()
-            == 1;
-        if !did_modify {
-            log::error!("did not end {} for {:?}", step, aid);
-        }
+            .unwrap();
     }
     async fn collector_remove_step(&self, aid: ArtifactIdNumber, step: &str) {
         self.conn()
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1168,7 +1168,7 @@ where
         self.conn()
             .execute(
                 "update collector_progress set start_time = statement_timestamp() \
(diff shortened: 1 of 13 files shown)


database/src/pool/postgres.rs
```diff
@@ -1157,27 +1168,22 @@ where
         self.conn()
             .execute(
                 "update collector_progress set start_time = statement_timestamp() \
-                where aid = $1 and step = $2 and end_time is null;",
+                where aid = $1 and step = $2 and start_time is null and end_time is null;",
```
Jamesbarford, line 1171: Why has the `and end_time is null` been removed?

database/src/pool/postgres.rs
```diff
@@ -1157,27 +1168,22 @@ where
...
+                where aid = $1 and step = $2 and start_time is not null;",
                 &[&(aid.0 as i32), &step],
             )
             .await
-            .unwrap()
-            == 1;
-        if !did_modify {
-            log::error!("did not end {} for {:?}", step, aid);
-        }
+            .unwrap();
```
Jamesbarford, line 1186: Wonder if we should make this return an `anyhow::Result<()>` ?

Jamesbarford (changes requested):

database/src/pool/postgres.rs
Kobzol: You mean in `collector_end_step`? Because we can finish a step multiple times now. I don't want to reset the start time when that happens, but I want to reset the end time. It's not fully needed, because in reality what we do is *start* a step multiple times, not necessarily also finish it, but it shouldn't hurt.

database/src/pool/postgres.rs
Kobzol: Ideally yes, but that's true for most methods in this DB implementation, so it's IMO orthogonal. Eventually we should get rid of the unwraps here and migrate everything to results. Not that it would change much particularly for the collector though, it would probably just crash on the first failure anyway.

database/src/pool/postgres.rs
Jamesbarford: Makes sense :+1:

Kobzol:
Second attempt, let's see what happens in production..

PR #2212, opened 2025-07-29
Feat; collector dequeue job

Jamesbarford:
- Add a command for adding a collector to the database
- Add a command called `DequeueJob` (open to alternative names) that will take a job from the `job_queue` and print to standard out if a job exists.
- Exposed a few properties on the structs `BenchmarkJob` and `CollectorConfig` using `pub fn <name_of_property>()` methods
- More mapping functions between the collector / database enums
- Altered the `job_queue` table from having a `collector_id` to `collector_name` as the name column in the collector is unique, cannot be NULL and saves a lookup to get the `id` using the `collector_name`. I've retained the foreign key constraint.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -368,7 +368,7 @@ static MIGRATIONS: &[&str] = &[
     // to use as we do not expose the `collector_id` in the code
     r#"
     ALTER TABLE job_queue DROP CONSTRAINT IF EXISTS job_queue_collector;
-    ALTER TABLE job_queue ADD COLUMN IF NOT EXISTS collector_name TEXT;
+    ALTER TABLE job_queue ADD COLUMN collector_name TEXT;
     ALTER TABLE job_queue
         ADD CONSTRAINT job_queue_collector
             FOREIGN KEY (collector_name)


collector/src/compile/benchmark/collector_config.rs
```diff
@@ -0,0 +1,53 @@
+use chrono::{DateTime, Utc};
+use database::BenchmarkSet;
+
+use super::target::Target;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct CollectorConfig {
```
Kobzol: Let's just use the database version.

database/src/pool/postgres.rs
```diff
@@ -1710,6 +1724,181 @@ where
...
+                        updated_queue
+                    WHERE
+                        updated_queue.request_tag = benchmark_request.tag
+                )
+                SELECT
+                    *
+                FROM
+                    updated_queue;",
+                &[
+                    &BenchmarkJobStatus::Queued,
```
Kobzol: Let's change this to a string and remove `impl_to_postgresql_via_to_string!(BenchmarkJobStatus);`.

database/src/pool/postgres.rs
```diff
@@ -1710,6 +1724,181 @@ where
...
+                        job_queue.target,
+                        job_queue.backend,
+                        job_queue.profile,
+                        job_queue.request_tag,
+                        job_queue.benchmark_set,
+                        job_queue.created_at,
+                        job_queue.status,
+                        job_queue.started_at,
+                        job_queue.retry
+                ), updated_request AS (
```
Kobzol: The modification of the request status is not needed, it is already in progress.

collector/src/bin/collector.rs
```diff
@@ -1266,6 +1299,65 @@ Make sure to modify `{dir}/perf-config.json` if the category/artifact don't matc
...
+            // collector's setup
+            let collector_config: CollectorConfig = rt
+                .block_on(conn.get_collector_config(&collector_name))?
+                .into();
+
+            let collector_target = collector_config.target();
+            if collector_target.as_str() != target {
+                panic!(
+                    "Mismatching target for collector expected `{}` got `{}`",
+                    collector_target, target
```
Kobzol:
```suggestion
                    "Mismatching target for collector expected `{collector_target}` got `{target}`",
```

database/src/pool/postgres.rs
```diff
@@ -1710,6 +1724,181 @@ where
...
+                    WHERE
+                        job_queue.id = picked.id
+                    RETURNING
+                        job_queue.target,
+                        job_queue.backend,
+                        job_queue.profile,
+                        job_queue.request_tag,
+                        job_queue.benchmark_set,
+                        job_queue.created_at,
+                        job_queue.status,
```
Kobzol: We don't need to query the status, because we set it.

database/src/pool/postgres.rs
```diff
@@ -1710,6 +1724,181 @@ where
...
+                    SET
+                        collector_name = $4,
+                        started_at = NOW(),
+                        status = $5
+                    FROM
+                        picked
+                    WHERE
+                        job_queue.id = picked.id
+                    RETURNING
+                        job_queue.target,
```
Kobzol: We don't have to query the target and benchmark set, because we set it.

database/src/pool/postgres.rs
```diff
@@ -1710,6 +1724,181 @@ where
...
+        target: &Target,
+        benchmark_set: &BenchmarkSet,
+    ) -> anyhow::Result<Option<BenchmarkJob>> {
+        // We take the oldest job from the job_queue matching the benchmark_set,
+        // target and status of 'queued'
+        let row_opt = self
+            .conn()
+            .query_opt(
+                "
+                WITH picked AS (
```
Kobzol, line 1821: Could you try to run `EXPLAIN` and check that the `job_queue` index is applied here?

Jamesbarford: It's not, we get a `Seq Scan` on the CTE (common table expression) `WITH picked` however we are getting an index scan on the CTE `updated_queue` as it uses the primary key. Creating the following;

```sql
CREATE INDEX IF NOT EXISTS job_queue_status_target_benchmark_set_idx (status, target, benchmark_set);
```

Uses a `Bitmap heap scan`. So I guess we create the index?

database/src/pool/postgres.rs
Jamesbarford: 168260b0a5ae9a9db681b4ffd45d9479b928ed5d :+1:

database/src/pool/postgres.rs
Jamesbarford: 168260b0a5ae9a9db681b4ffd45d9479b928ed5d :+1:

database/src/pool/postgres.rs
Jamesbarford: 168260b0a5ae9a9db681b4ffd45d9479b928ed5d :+1:

database/src/pool/postgres.rs
Jamesbarford: 1c200c6e2f0168096c7edb4e8e9fba8cb813244e :+1: Did it for `BenchmarkRequestStatus` too

collector/src/compile/benchmark/collector_config.rs
Jamesbarford: a4c2ddc45a7a30846f951c31548b4c1a16c27579 :+1:

collector/src/bin/collector.rs
Jamesbarford: a4c2ddc45a7a30846f951c31548b4c1a16c27579 +1

database/src/pool/postgres.rs
Jamesbarford: Created; a094fa3fe8c8d2c4004f521b07b11b7d1e091e81

Kobzol: Wait, even with the index it uses bitmap heap scan, not a "proper" index access?

database/src/pool/postgres.rs
```diff
@@ -361,6 +364,19 @@ static MIGRATIONS: &[&str] = &[
     );
     CREATE INDEX IF NOT EXISTS job_queue_request_tag_idx ON job_queue (request_tag);
     "#,
+    // The name is unique and not null in the collector config and is simpler
+    // to use as we do not expose the `collector_id` in the code
+    r#"
+    ALTER TABLE job_queue DROP CONSTRAINT IF EXISTS job_queue_collector;
+    ALTER TABLE job_queue ADD COLUMN IF NOT EXISTS collector_name TEXT;
```
Kobzol: Why did you add the `IF NOT EXISTS` clause? That shouldn't be needed, right?

database/src/pool/postgres.rs
Jamesbarford: It uses a Bitmap Heap scan with a bitmap index scan when using the index.

(foo_bar, is what I called the index locally)
```
 CTE Scan on updated_queue  (cost=100.77..100.79 rows=1 width=184) (actual time=1.519..1.523 rows=1 loops=1)
   CTE picked
     ->  Limit  (cost=92.44..92.45 rows=1 width=18) (actual time=1.382..1.384 rows=1 loops=1)
           ->  LockRows  (cost=92.44..101.71 rows=742 width=18) (actual time=1.381..1.381 rows=1 loops=1)
                 ->  Sort  (cost=92.44..94.29 rows=742 width=18) (actual time=1.368..1.369 rows=1 loops=1)
                       Sort Key: job_queue.created_at
                       Sort Method: quicksort  Memory: 113kB
                       ->  Bitmap Heap Scan on job_queue  (cost=17.74..88.73 rows=742 width=18) (actual time=0.194..0.686 rows=1378 loops=1)
                             Recheck Cond: ((status = 'queued'::text) AND (target = 'x86_64-unknown-linux-gnu'::text) AND (benchmark_set = 1))
                             Heap Blocks: exact=22
                             ->  Bitmap Index Scan on foo_bar  (cost=0.00..17.55 rows=742 width=0) (actual time=0.129..0.129 rows=1385 loops=1)
                                   Index Cond: ((status = 'queued'::text) AND (target = 'x86_64-unknown-linux-gnu'::text) AND (benchmark_set = 1))
   CTE updated_queue
     ->  Update on job_queue job_queue_1  (cost=0.28..8.33 rows=1 width=106) (actual time=1.516..1.518 rows=1 loops=1)
           ->  Nested Loop  (cost=0.28..8.33 rows=1 width=106) (actual time=1.410..1.411 rows=1 loops=1)
                 ->  CTE Scan on picked  (cost=0.00..0.02 rows=1 width=32) (actual time=1.390..1.390 rows=1 loops=1)
                 ->  Index Scan using job_queue_pkey on job_queue job_queue_1  (cost=0.28..8.30 rows=1 width=10) (actual time=0.016..0.016 rows=1 loops=1)
                       Index Cond: (id = picked.id)
 Planning Time: 0.497 ms
 Trigger for constraint job_queue_collector: time=0.122 calls=1
 Execution Time: 1.921 ms
```

database/src/pool/postgres.rs
Jamesbarford: I'll remove it, I was running this block of SQL statements multiple times which caused it to error out as I'd already created the column

Jamesbarford: 52bfe1f796c47d852b54a6c7d76f64eaa099740a

database/src/pool/postgres.rs
Kobzol: Ok, let's leave it like this. If we run into perf. issues later, we can always modify the index or the query.

Kobzol (approved):
Thanks! I will continue on top of this and try to join the dequeuing logic with actually running the benchmarks.

PR #2213, opened 2025-07-29
drop dep from tabled

klensy:
This removes optional dep `testing_table` from builded deps (but it still preserved in Cargo.lock).


Kobzol (approved):
Thanks.

PR #2214, opened 2025-07-31
Feat; mark benchmark requests as complete

Jamesbarford:
Related to; https://github.com/rust-lang/rustc-perf/pull/2200 which I've closed as it was becoming fiddly to update.
- Mark `benchmark_request`'s as complete based on;
    - parent being complete if there is a parent
    - all jobs in the `job_queue` being in a terminal state.

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -313,3 +313,36 @@ Columns:
   execute.
 * **is_active** (`boolean NOT NULL`): For controlling whether the collector is
   active for use. Useful for adding/removing collectors.
+
+### job_queue
+
+This table stores ephemeral benchmark jobs, which specifically tell the
+collector which benchmarks it should execute. The jobs will be kept in the
+table for ~30 days after being completed, so that we can quickly figure out
+what master parent jobs we need to backfill when handling try builds.
+
+Columns:
+
+- **id** (`bigint` / `serial`): Primary-key identifier for the job row;
+  auto-increments with each new job.
+- **request_id** (`bigint`): References the parent benchmark request that
+  spawned this job.
+- **target** (`text NOT NULL`): Hardware/ISA the benchmarks must run on
+  (e.g. AArch64, x86_64).
+- **backend** (`text NOT NULL`): Code generation backend the collector should
+  test (e.g. llvm, cranelift).
+- **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
+  execute.
+- **collector_id** (`text`): Id of the collector that claimed the job
+  (populated once the job is started).
+- **created_at** (`timestamptz NOT NULL`): Datetime when the job was queued.
+- **started_at** (`timestamptz`): Datetime when the collector actually began
+  running the benchmarks; NULL until the job is claimed.
+- **completed_at** (`timestampt`): Datetime when the collector finished
+  (successfully or otherwise); used to purge rows after ~30 days.
+- **status** (`text NOT NULL`): Current job state. `queued`, `in_progress`,
+  `success`, or `failure`.
+- **retry** (`int NOT NULL`): Number of times the job has been re-queued after
+  a failure; 0 on the first attempt.
+- **error** (`text`): Optional error message or stack trace from the last
+  failed run; NULL when the job succeeded.
--- a/database/schema.md
+++ b/database/schema.md
@@ -323,26 +323,24 @@ what master parent jobs we need to backfill when handling try builds.
 
 Columns:
 
-- **id** (`bigint` / `serial`): Primary-key identifier for the job row;
-  auto-increments with each new job.
-- **request_id** (`bigint`): References the parent benchmark request that
+* **id** (`bigint` / `serial`): Primary*key identifier for the job row;
+  auto*increments with each new job.
+* **request_tag** (`text`): References the parent benchmark request that
   spawned this job.
-- **target** (`text NOT NULL`): Hardware/ISA the benchmarks must run on
+* **target** (`text NOT NULL`): Hardware/ISA the benchmarks must run on
   (e.g. AArch64, x86_64).
-- **backend** (`text NOT NULL`): Code generation backend the collector should
+* **backend** (`text NOT NULL`): Code generation backend the collector should
   test (e.g. llvm, cranelift).
-- **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
+* **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
   execute.
-- **collector_id** (`text`): Id of the collector that claimed the job
+* **collector_name** (`text`): Name of the collector that claimed the job
   (populated once the job is started).
-- **created_at** (`timestamptz NOT NULL`): Datetime when the job was queued.
-- **started_at** (`timestamptz`): Datetime when the collector actually began
+* **created_at** (`timestamptz NOT NULL`): Datetime when the job was queued.
+* **started_at** (`timestamptz`): Datetime when the collector actually began
   running the benchmarks; NULL until the job is claimed.
-- **completed_at** (`timestampt`): Datetime when the collector finished
+* **completed_at** (`timestampt`): Datetime when the collector finished
   (successfully or otherwise); used to purge rows after ~30 days.
-- **status** (`text NOT NULL`): Current job state. `queued`, `in_progress`,
+* **status** (`text NOT NULL`): Current job state. `queued`, `in_progress`,
   `success`, or `failure`.
-- **retry** (`int NOT NULL`): Number of times the job has been re-queued after
+* **retry** (`int NOT NULL`): Number of times the job has been re*queued after
   a failure; 0 on the first attempt.
-- **error** (`text`): Optional error message or stack trace from the last
-  failed run; NULL when the job succeeded.
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -253,6 +253,13 @@ pub trait Connection: Send + Sync {
         target: &Target,
         benchmark_set: &BenchmarkSet,
     ) -> anyhow::Result<Option<BenchmarkJob>>;
+
+    /// Try and mark the benchmark_request as completed. Will return `true` if
+    /// it has been marked as completed else `false` meaning there was no change
+    async fn mark_benchmark_request_as_completed(
+        &self,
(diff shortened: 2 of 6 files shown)


database/src/pool/postgres.rs
```diff
@@ -1887,6 +1888,82 @@ where
...
+            "Can only mark benchmark request whos status is in_progress as complete"
+        );
+
+        // Find if the benchmark is completed and update it's status to completed
+        // in one SQL block
+        let row = self
+            .conn()
+            .query_opt(
+                "
+                UPDATE
```
Jamesbarford, line 1906: This is a bit of a monster but atomically achieves what we want it to.

The query uses indexes as can be seen here through running `EXPLAIN ANALYSE` below.

It's verbose however the points of interest are all operations use either index scans or bitmap index scans

```
Update on benchmark_request  (cost=4.71..28.37 rows=1 width=52) (actual time=0.047..0.049 rows=0 loops=1)
   ->  Nested Loop Anti Join  (cost=4.71..28.37 rows=1 width=52) (actual time=0.045..0.047 rows=0 loops=1)
         ->  Index Scan using benchmark_request_tag_key on benchmark_request  (cost=0.42..17.14 rows=1 width=15) (actual time=0.045..0.046 rows=0 loops=1)
               Index Cond: (tag = 'foo'::text)
               Filter: ((status <> 'completed'::text) AND ((parent_sha IS NULL) OR (NOT (SubPlan 1))))
               SubPlan 1
                 ->  Nested Loop  (cost=0.70..16.71 rows=2 width=0) (never executed)
                       ->  Index Scan using job_queue_request_tag_idx on job_queue job_queue_1  (cost=0.28..12.25 rows=2 width=8) (never executed)
                             Index Cond: (request_tag = benchmark_request.parent_sha)
                             Filter: (status <> ALL ('{success,failure}'::text[]))
                       ->  Materialize  (cost=0.42..4.44 rows=1 width=9) (never executed)
                             ->  Index Only Scan using benchmark_request_tag_key on benchmark_request parent_request  (cost=0.42..4.44 rows=1 width=9) (never execut
ed)
                                   Index Cond: (tag = benchmark_request.parent_sha)
                                   Heap Fetches: 0
         ->  Bitmap Heap Scan on job_queue  (cost=4.30..11.21 rows=2 width=14) (never executed)
               Recheck Cond: (request_tag = 'foo'::text)
               Filter: (status <> ALL ('{success,failure}'::text[]))
               ->  Bitmap Index Scan on job_queue_request_tag_idx  (cost=0.00..4.29 rows=2 width=0) (never executed)
                     Index Cond: (request_tag = 'foo'::text)
 Planning Time: 2.288 ms
 Execution Time: 0.238 ms
```

Jamesbarford: I'm open to suggestions on how to test this. The obstacles I see are;
- Marking jobs as complete.
- A way of getting a benchmark request (or Vec of them) to see if one was marked as complete.

database/src/pool/postgres.rs
```diff
@@ -1887,6 +1888,82 @@ where
...
+                        benchmark_request.parent_sha IS NULL
+                        OR NOT EXISTS (
+                            SELECT
+                                1
+                            FROM
+                                job_queue
+                            JOIN
+                                benchmark_request AS parent_request 
+                            ON
+                                parent_request.tag = benchmark_request.parent_tag
```
```suggestion

```
Why do we need the `JOIN`? Wouldn't it work the same if we deleted this?

database/src/pool/postgres.rs
```diff
@@ -1887,6 +1888,82 @@ where
...
+                ],
+            )
+            .await
+            .context("Failed to mark benchmark_request as completed")?;
+        // The affected id is returned by the query thus we can use the row's
+        // presence to determine if the request was marked as completed
+        if let Some(row) = row {
+            let completed_at = row.get::<_, DateTime<Utc>>(0);
+            // Also mutate our object
+            benchmark_request.status = BenchmarkRequestStatus::Completed { completed_at };
```
Kobzol: I don't think that this is needed. We don't do any kind of in-memory ORM updates elsewhere in the codebase. We just assume that if you want to reload the latest state from the DB, you have to reload stuff from it.

database/src/pool/postgres.rs
Jamesbarford: We don't need the `JOIN`. I made a bit of a hash of it. We need to say where the tag matches the parent_sha. A bit like in the the where above; 183a7e00d95b373d68ada2f4272605d12080ca08

database/src/pool/postgres.rs
```diff
@@ -1957,6 +1950,44 @@ where
...
+                self.conn()
+                    .execute(
+                        "
+                        UPDATE
+                            job_queue 
+                        SET
+                            status = $1,
+                            completed_at = NOW()
+                        WHERE
+                            request_tag = $2
```
Kobzol: This is not enough, because the jobs will be distinguished also by profile and codegen backend. Since the collector will always dequeue a specific job and know it's ID, we should just mark the jobs as finished based on their unique ID.

database/src/pool/postgres.rs
```diff
@@ -1957,6 +1950,44 @@ where
             Ok(false)
         }
     }
+
+    async fn mark_benchmark_job_as_completed(
+        &self,
+        request_tag: &str,
+        benchmark_set: u32,
+        target: &Target,
+        status: &BenchmarkJobStatus,
```
Kobzol: Maybe create an `enum BenchmarkJobConclusion { Success, Failure }` to avoid the invalid states?

database/src/pool/postgres.rs
Jamesbarford: I see, yes. Exposing the database id in the struct would seem to make sense in this instance

database/src/pool/postgres.rs
Jamesbarford: It's more a developer error I think. Much like the `panic!(...)` I put in `update_benchmark_request_status(...)` to prevent the wrong functions being used in the wrong context. And provided we have good test coverage, or good enough, we shouldn't need anything else. It was easy enough for me to fix the tests I broke.

Jamesbarford: After moving some code around you suggestion makes sense if we have the function prototype as;

```rs
async fn mark_benchmark_job_as_completed(
    &self,
    id: u32,
    benchmark_job_conclusion: &BenchmarkJobConclusion
) -> anyhow::Result<()>;
```

Then it becomes intuitive and is different to my previous comment.

Jamesbarford: cd7ff3d32dd4e2f179bae76e3d903b5f137ed874

database/src/pool.rs
```diff
@@ -388,6 +400,42 @@ mod tests {
...
+            .unwrap();
+
+        assert_eq!(job.request_tag(), request_tag);
+
+        /* Mark the job as complete */
+        db.mark_benchmark_job_as_completed(job.id(), &BenchmarkJobConclusion::Success)
+            .await
+            .unwrap();
+
+        db.mark_benchmark_request_as_completed(request_tag)
```
Kobzol: This should assert that it returned `true`.

database/src/pool.rs
```diff
@@ -837,4 +888,85 @@ mod tests {
...
+                .add_collector_config("collector-1", &Target::X86_64UnknownLinuxGnu, 1, true)
+                .await;
+            assert!(insert_result.is_ok());
+
+            let benchmark_request =
+                BenchmarkRequest::create_master("sha-1", "parent-sha-1", 42, time);
+            db.insert_benchmark_request(&benchmark_request)
+                .await
+                .unwrap();
+            assert!(db
```
Kobzol: Same here, it should assert that it returned `true`.

database/src/pool/postgres.rs
```diff
@@ -1868,25 +1875,101 @@ where
...
+                WHERE
+                    benchmark_request.tag = $2
+                    AND benchmark_request.status != $1
+                    AND NOT EXISTS (
+                        SELECT
+                            1
+                        FROM
+                            job_queue
+                        WHERE
+                            request_tag = benchmark_request.tag
```
Kobzol: Let's use `job_queue.request_tag` and `job_queue.status` here. To be consistent with the subquery below, but mainly to avoid using attributes of the parent query by accident.

site/src/job_queue/mod.rs
```diff
@@ -345,15 +353,51 @@ mod tests {
...
             .unwrap();
+
+        assert_eq!(job.request_tag(), request_tag);
+
+        /* Mark the job as complete */
+        db.mark_benchmark_job_as_completed(job.id(), &BenchmarkJobConclusion::Success)
+            .await
+            .unwrap();
+
+        db.mark_benchmark_request_as_completed(request_tag)
```
Kobzol: Same here.

Btw there's quite a lot of duplication between the DB and website tests around the job queue and requests. I'll go through them and try to create some more high-level test API and unify the tests in one of the packages, otherwise it will become quite cumbersome to modify the tests after code changes.

database/src/pool.rs
Jamesbarford: c9b32be9cefd809e408509e549adfcc73c4984f0

database/src/pool.rs
Jamesbarford: c9b32be9cefd809e408509e549adfcc73c4984f0

database/src/pool/postgres.rs
Jamesbarford: c9b32be9cefd809e408509e549adfcc73c4984f0

site/src/job_queue/mod.rs
Jamesbarford: Yes there is, a separate "queue testing" `package` would be helpful

database/schema.md
```diff
@@ -313,3 +313,34 @@ Columns:
...
+
+This table stores ephemeral benchmark jobs, which specifically tell the
+collector which benchmarks it should execute. The jobs will be kept in the
+table for ~30 days after being completed, so that we can quickly figure out
+what master parent jobs we need to backfill when handling try builds.
+
+Columns:
+
+* **id** (`bigint` / `serial`): Primary*key identifier for the job row;
+  auto*increments with each new job.
```
fbstj, line 327: was this intentional italics?

Kobzol: Probably not :) We can fix it in some follow-up.

PR #2216, opened 2025-08-07
Preparation for collector job queue integration

Kobzol:
This PR contains a bunch of changes that I had to do to prepare the collector for integration with the new job queue system. It doesn't contain all the changes, but it should be self-contained enough to land as a separate PR.

Note: the last commit is global `clippy --fix`, there was a bunch of warnings accrued.

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -1064,9 +1064,7 @@ fn main_result() -> anyhow::Result<i32> {
                         let sha = commit.sha.to_string();
                         let sysroot = rt
                             .block_on(Sysroot::install(sha.clone(), &host_target_tuple, &backends))
-                            .with_context(|| {
-                                format!("failed to install sysroot for {commit:?}")
-                            })?;
+                            .with_context(|| format!("failed to install sysroot for {commit:?}"))?;
 
                         let mut benchmarks = get_compile_benchmarks(
                             &compile_benchmark_dir,
@@ -1928,9 +1926,7 @@ async fn bench_compile(
         );
         let result = measure(&mut processor).await;
         if let Err(s) = result {
-            eprintln!(
-                "collector error: Failed to benchmark '{benchmark_name}', recorded: {s:#}"
-            );
+            eprintln!("collector error: Failed to benchmark '{benchmark_name}', recorded: {s:#}");
             errors.incr();
             tx.conn()
                 .record_error(
--- a/collector/src/compile/execute/mod.rs
+++ b/collector/src/compile/execute/mod.rs
@@ -670,9 +670,7 @@ fn process_stat_output(
             continue;
         }
         if !pct.starts_with("100.") {
-            panic!(
-                "measurement of `{name}` only active for {pct}% of the time"
-            );
+            panic!("measurement of `{name}` only active for {pct}% of the time");
         }
         stats.insert(
             name.to_owned(),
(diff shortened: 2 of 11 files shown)


database/src/lib.rs
```diff
@@ -1093,8 +1093,14 @@ impl fmt::Display for BenchmarkJobStatus {
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
-pub struct BenchmarkSet(pub u32);
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct BenchmarkSet(u32);
```
Jamesbarford, line 1097: Could we possibly even get rid of this struct and use a `u32`? It is cumbersome to use anywhere outside of `database` hence why I think `u32` was made public. We don't have `struct Pr(u32)`, `struct DequeueCount(u32)` or `struct Tag(String)` for instance

database/src/lib.rs
```diff
@@ -1112,32 +1118,32 @@ pub struct BenchmarkJob {
...
 }
 
 impl BenchmarkJob {
     pub fn id(&self) -> u32 {
         self.id
     }
 
-    pub fn target(&self) -> &Target {
-        &self.target
+    pub fn target(&self) -> Target {
```
Jamesbarford, line 1129: Why have you removed all of the references?

database/src/lib.rs
```diff
@@ -1147,6 +1153,11 @@ impl BenchmarkJob {
             | BenchmarkJobStatus::Completed { collector_name, .. } => Some(collector_name),
         }
     }
+
+    /// How many times was the job already dequed?
+    pub fn deque_count(&self) -> u32 {
```
Jamesbarford, line 1158: Maybe we could also add something like;

```rs
pub has_failed(&self) -> bool {
    self.dequeue_count == MAX_DEQUEUE_COUNT && matches!(self.status, BenchmarkJobStatus::InProgress)
}
```

site/src/main.rs
```diff
@@ -43,7 +43,7 @@ async fn main() {
             let artifacts = res.index.load().artifacts().count();
             if commits + artifacts == 0 {
                 eprintln!("Loading complete but no data identified; exiting.");
-                std::process::exit(1);
+                // std::process::exit(1);
```
Jamesbarford: Should this be commented back in?

Kobzol: Good catch! But actually, this behavior is super annoying and I wanted to get rid of it for a long time. https://github.com/rust-lang/rustc-perf/pull/2218

database/src/lib.rs
Kobzol: The database layer just stores the number of dequeues, what is the specific number of dequeues that is considered a failure is determined by the collector, so I wouldn't put it into the database crate.

database/src/lib.rs
Kobzol: All of these are simple `Copy` types, so there's really no need to pass/return them by reference, same as we don't return e.g. the `u32` above as a reference.

database/src/lib.rs
Kobzol: Well, tbh I would argue that we should have at least `Pr(u32)` and `Tag(String)`, as especially the tag can be confused quite easily with other strings :laughing: The `BenchmarkJob` has three `u32`s (ID, benchmark set, dequeue counter), but I agree that it's probably not super important in this place to have it (new)typed. But since it's already in code, and I find it a bit better than raw `u32`, I would keep it.

database/src/lib.rs
Jamesbarford: Agree that it doesn't need to be at the database layer. However it seems odd that the collector decides it fails too as it feels like queue management than a function of what is essentially a worker node.

Kobzol: In theory the website could do it, but it's not very practical; the collector dequeues jobs and it has to know if it should continue with a given job or not.

Jamesbarford (approved):
LGTM, just needs a rebase

PR #2218, opened 2025-08-08
Do not exit website if there is no data

Kobzol:
It was super annoying for local tests. Let's see if CI is fine with it.


lqd:
It’s annoying when testing the UI (e.g. when using the perf.rlo remote endpoint even locally), right?

lqd:
Worst case you also add a `--terminate-if-no-data` flag for CI to avoid the handful of 5 second timeouts.

Jamesbarford:
Would `exit(EXIT_SUCCESS)` help matters? We still terminate but we don't set an error code?

lqd (approved):
It's fine.

PR #2220, opened 2025-08-18
Feat; add a duration column for benchmark_request's

Jamesbarford:
Adds a `duration_ms` column to the `benchmark_request` table. It takes the difference from the latest `completed_at` and the earliest `started_at` from jobs associated with the request. This column is populated when marking a request as completed.

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -814,7 +814,7 @@ pub enum BenchmarkRequestStatus {
     InProgress,
     Completed {
         completed_at: DateTime<Utc>,
-        duration_ms: u32,
+        duration: Duration,
     },
 }
 
@@ -846,9 +846,9 @@ impl BenchmarkRequestStatus {
                 completed_at: completion_date.ok_or_else(|| {
                     anyhow!("No completion date for a completed BenchmarkRequestStatus")
                 })?,
-                duration_ms: duration_ms.ok_or_else(|| {
+                duration: Duration::from_millis(duration_ms.ok_or_else(|| {
                     anyhow!("No completion duration for a completed BenchmarkRequestStatus")
-                })? as u32,
+                })? as u64),
             }),
             _ => Err(anyhow!("Unknown BenchmarkRequestStatus `{text}`")),
         }
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -973,6 +973,9 @@ mod tests {
 
             assert_eq!(job.request_tag(), benchmark_request.tag().unwrap());
 
+            /* Make the job take some amount of time */
+            std::thread::sleep(Duration::from_millis(1000));
+
             /* Mark the job as complete */
             db.mark_benchmark_job_as_completed(job.id(), BenchmarkJobConclusion::Success)
                 .await
@@ -994,10 +997,10 @@ mod tests {
                 req.status(),
                 BenchmarkRequestStatus::Completed { .. }
             ));
-            let BenchmarkRequestStatus::Completed { duration_ms, .. } = req.status() else {
+            let BenchmarkRequestStatus::Completed { duration, .. } = req.status() else {
                 unreachable!();
             };
-            assert!(duration_ms >= 1);
+            assert!(duration >= Duration::from_millis(1000));
 
             let completed_index = db.load_benchmark_request_index().await.unwrap();
             assert!(completed_index.contains_tag("sha-1"));


database/src/pool/postgres.rs
```diff
@@ -1891,7 +1894,14 @@ where
...
                 SET
                     status = $1,
-                    completed_at = NOW()
+                    completed_at = NOW(),
+                    duration_ms = (
+                        SELECT (MAX(EXTRACT('epoch' FROM job_queue.completed_at)) -
+                                MIN(EXTRACT('epoch' FROM job_queue.started_at))) * 1000 AS duration_ms
+                        FROM
+                            job_queue
+                        LEFT JOIN benchmark_request ON job_queue.request_tag = benchmark_request.tag
```
Kobzol: This has to be done with some other way, right? This would just take a look at the largest `completed_at` and oldest `started_at` in the job queue, regardless of the `benchmark_request`(s) that we're completing. Maybe we can do it in two steps? First finish the requests, and then fill in the duration for the steps that were completed, but have no duration set yet.

Jamesbarford: Yes you're right the query is incorrect. I think we can add;

```sql
WHERE benchmark_request.tag = $<the_tag>
```

Then we will only get the jobs associated with the request. Which means we can keep it as one query.

Jamesbarford: 37b346052c41aeedb81d8aeaedb97baec1d73678

Jamesbarford: fceeab8773a72d5ce2d9a42c35d94b10ad072ebe I've simplified it further, all it needs is the `WHERE` not the `LEFT JOIN`

Kobzol: Ah, I forgot that we try to finish a specific tag, I was somehow assuming that we just finish all requests that are done, but that's no longer the case. Yeah, looks good! Could you please add a test?

Jamesbarford: > I forgot that we try to finish a specific tag, I was somehow assuming that we just finish all requests that are done

We are doing this as this function is called inside a loop in the cron job. Just we're not doing it as one query (though we could, with some effort, probably do that in 1 query).

For the test I've modified a pre-existing one; 030948b9036517999f4bb78fc5a41b4e19f2a894. 

Ideally we'd add roughly 3 requests, add 3 jobs per request, mark 2 of the requests as complete, modify the dates for `started_at` and `completed_at` per job belonging to the completed requests and then assert the `duration_ms` is set as expected for the 2 completed requests and that the `duration_ms` is NULL for the `in_progress` job. I can't see how we can do something like this with our current setup however.

database/src/lib.rs
```diff
@@ -812,7 +812,10 @@ pub enum BenchmarkRequestStatus {
     WaitingForArtifacts,
     ArtifactsReady,
     InProgress,
-    Completed { completed_at: DateTime<Utc> },
+    Completed {
+        completed_at: DateTime<Utc>,
+        duration_ms: u32,
```
```suggestion
        duration: Duration,
```
In Rust we can just use a proper duration type.

database/src/pool.rs
```diff
@@ -980,9 +980,28 @@ mod tests {
 
             db.mark_benchmark_request_as_completed(tag).await.unwrap();
 
-            let completed = db.load_benchmark_request_index().await.unwrap();
+            /* From the status page view we can see that the duration has been
+             * updated. Albeit that it will be a very short duration. */
+            let status_page_view = db.get_status_page_data().await.unwrap();
```
Kobzol, line 988: You can add a sleep to the test to wait e.g. at least 1 second.

database/src/pool/postgres.rs
```diff
@@ -1891,7 +1894,14 @@ where
                     benchmark_request
                 SET
                     status = $1,
-                    completed_at = NOW()
+                    completed_at = NOW(),
+                    duration_ms = (
```
Kobzol, line 1898: What happens if there is no job in the job queue attached to the request? You can test e.g. by adding a `WHERE` clause that is always false. Will the query return an error?

Jamesbarford: The query will not error; `duration_ms` would be set as `NULL`.

If the `WHERE` is false however that means that the `request_tag` was not set properly for the jobs. And if that was the case then the rest of the query would return `false` as it scans all jobs matching the `tag` for the job; `WHERE job_queue.request_tag = benchmark_request.tag` thus would not be marked as complete.

The other case would be if there is a divergence of tags in the query which I can't see how that would happen. I don't quite understand the scenario you are describing? Nor see how a divergence of `request_tag` to `tag` may happen which would cause the result of the `SELECT` to return a `NULL`.

database/src/pool.rs
Jamesbarford: 43c0c4b7e8d7e10e6cea24cb92b37d2d86b2edd0 :+1:

database/src/lib.rs
Jamesbarford: 43c0c4b7e8d7e10e6cea24cb92b37d2d86b2edd0 :+1:

Kobzol (approved):

PR #2221, opened 2025-08-18
Feat; get collectors configuration query

Jamesbarford:
Return a database dump of all the collectors


database/src/pool.rs
```diff
@@ -273,6 +273,9 @@ pub trait Connection: Send + Sync {
     ) -> anyhow::Result<()>;
 
     async fn get_status_page_data(&self) -> anyhow::Result<PartialStatusPageData>;
+
+    /// Get all of the configuration for all of the collectors
+    async fn get_collectors_config(&self) -> anyhow::Result<Vec<CollectorConfig>>;
```
```suggestion
    async fn get_collector_configs(&self) -> anyhow::Result<Vec<CollectorConfig>>;
```

Jamesbarford: b2f3fb4d7590ec5615cfae0648171cc3edd9d507 :+1:

PR #2226, opened 2025-08-20
Add job benchmark loop

Kobzol:
This PR adds the main logic required to execute job benchmarks in the collector.

It handles:
- Updating the collector's heartbeat periodically
- Quick loading of already downloaded sysroots, to avoid redownloading them in-between jobs and also collector restarts (useful for local testing)
- Dequeing jobs, including in-progress jobs and expanding benchmark sets
- Distinguishing between transient and permanent job errors, storing job errors into the DB, and marking jobs as failed or successful
- Marking jobs that have been dequeued too many times as failed
- Reconnecting to the DB if a transient I/O/network/DB error happens, to try to refresh the DB connection

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -1528,24 +1528,19 @@ async fn run_benchmark_job(
             Toolchain::from_sysroot(&sysroot, commit.sha.clone())
         }
         ArtifactId::Tag(tag) => {
-            create_toolchain_from_published_version(&tag, job.target().as_str())?
+            create_toolchain_from_published_version(tag, job.target().as_str())?
         }
     };
     log::info!("Sysroot download finished");
 
-    let (compile_config, runtime_config) = create_benchmark_configs(
-        conn,
-        &toolchain,
-        &artifact_id,
-        &job,
-        &all_compile_benchmarks,
-    )
-    .await
-    .map_err(|error| {
-        BenchmarkJobError::Permanent(anyhow::anyhow!(
-            "Cannot prepare benchmark configs: {error:?}"
-        ))
-    })?;
+    let (compile_config, runtime_config) =
+        create_benchmark_configs(conn, &toolchain, &artifact_id, job, all_compile_benchmarks)
+            .await
+            .map_err(|error| {
+                BenchmarkJobError::Permanent(anyhow::anyhow!(
+                    "Cannot prepare benchmark configs: {error:?}"
+                ))
+            })?;
 
     let shared = SharedBenchmarkConfig {
         artifact_id,
@@ -1594,7 +1589,7 @@ async fn create_benchmark_configs(
     let compile_config = if bench_rustc || !bench_compile_benchmarks.is_empty() {
         Some(CompileBenchmarkConfig {
             benchmarks: all_compile_benchmarks
-                .into_iter()
+                .iter()
                 .filter(|b| bench_compile_benchmarks.contains(&b.name))
                 .cloned()
                 .collect(),
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -1152,11 +1152,11 @@ mod tests {
 
             assert_eq!(collector_configs[0].name(), collector_name_one);
             assert_eq!(collector_configs[0].benchmark_set(), benchmark_set_one);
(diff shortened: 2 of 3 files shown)


collector/src/toolchain.rs
```diff
@@ -145,54 +213,54 @@ impl SysrootDownload {
...
         }
 
-        Err(anyhow!(
-            "unable to download sha {} triple {} module {} from any of {:?}",
-            self.rust_sha,
-            self.triple,
-            component,
-            urls
-        ))
+        if !non_404_error {
```
Jamesbarford: I could be mistaken however this reads slightly oddly; if http is not a not found `404` then we produce an error of sha not found? Then we fall through to an IO error if it was a `404` despite the sha actually not being found.

database/src/pool/postgres.rs
```diff
@@ -1776,7 +1776,7 @@ where
         benchmark_set: BenchmarkSet,
     ) -> anyhow::Result<Option<(BenchmarkJob, ArtifactId)>> {
         // We take the oldest job from the job_queue matching the benchmark_set,
-        // target and status of 'queued'
+        // target and status of 'queued' or 'in-progress'
```
Jamesbarford:
```suggestion
        // target and status of 'queued' or 'in_progress'
```

collector/src/toolchain.rs
```diff
@@ -145,54 +213,54 @@ impl SysrootDownload {
...
+            Err(SysrootDownloadError::SysrootShaNotFound)
+        } else {
+            Err(SysrootDownloadError::IO(anyhow!(
```
Jamesbarford, line 262: We can use `resp.error_for_status()` so we can get the actual error message too from the response which could be useful for debugging; https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status

Kobzol: `error_for_status()` is nice, but the reason why we don't provide more context here is simply because we have up to three (potentially different) errors. Now that we detect 404s explicitly, we could just bail out on the first non-404 error, but I'm a bit worried about backwards compatibility, I don't know if "toolchain not found" is always reported with a 404.

collector/src/toolchain.rs
Kobzol: The variable name wasn't great, renamed it and added a comment.

collector/src/toolchain.rs
Jamesbarford: For logging purposes, it would be useful to distinguish between a `500` or other HTTP status codes, since they could indicate different underlying issues. At present, these are all reported as `SysrootDownloadError::SysrootShaNotFound`, which is overly specific and potentially misleading given the range of errors that could occur

Kobzol: Not sure if I follow. `SysrootShaNotFound` is returned if we see *only* 404 errors. If we ever see a different error than 404, we won't return `SysrootShaNotFound`.

Jamesbarford: Perhaps we can invert it;

```rs
if is_404 {
    // The only errors we saw were 404, so we assume that the toolchain is simply not on CI
    Err(SysrootDownloadError::SysrootShaNotFound)
} else {
    Err(SysrootDownloadError::IO(anyhow!("unable to download sha {} triple {} module {component} from any of {urls:?} error: {:?}",
         self.rust_sha,
         self.triple,
         resp.error_for_status() // <- so we can see what the error is
    ))
}
```

I still think it is valuable to be able to see what the error is for the HTTP request, either through logging or making it part of the error so we don't lose information that could be useful.

Kobzol: This wouldn't work most of the time, because we try nightly -> beta -> stable, and 99% of the time we actually care about nightly. If there was an error in nightly, then it would be swallowed anyway, and this would only show the last error from stable.

The invert is also a bit annoying, because it would have to start at `let mut is_404 = true;`, but that's misleading, because at the start there was no 404 error.

Added logging of the error response body, that should be enough.

Jamesbarford: Looks good :+1:

Kobzol:
Ok, let's try.

PR #2227, opened 2025-08-21
Feat; attempt at rendering new status page

Jamesbarford:
Caveat; I'm no UI designer/programmer.

An attempt at rendering something more useful on the new status page;
- queue
- completed requests
- collectors along with what they are working on

I think we may want to do a join from; job -> items in a benchmark set -> pstats. Then we can see what is going on at the job level. Also to give an estimate we'd want to do the same kind of query but take a sample of completed "pstats" to work out the duration per "pstat" that is currently in flight. We could then also estimate how long a job would take from that sample.

### With in_progress requests
Screenshot from 2025-08-21 15-43-11

### Idle collectors
Screenshot from 2025-08-21 15-42-38

### DIFF
--- a/site/frontend/src/pages/status_new/collector.vue
+++ b/site/frontend/src/pages/status_new/collector.vue
@@ -1,0 +1,228 @@
+<script setup lang="ts">
+import {computed} from "vue";
+import {BenchmarkRequest} from "./data";
+
+const props = defineProps<{
+  collector: CollectorConfigAndWork;
+}>();
+
+function statusClass(c: CollectorConfig): string {
+  return c.isActive ? "active" : "inactive";
+}
+</script>
+
+<template>
+  <div class="collector-card">
+    <div class="header">
+      <div class="collector-name">
+        <span>
+          <strong class="collector-sm-padding-right">{{
+            collector.config.name
+          }}</strong>
+          <span
+            class="collector-sm-padding-left-right collector-left-divider"
+            >{{ collector.config.target }}</span
+          >
+          <span
+            class="collector-sm-padding-left-right status"
+            :class="statusClass(collector.config)"
+          >
+            {{ collector.config.isActive ? "Active" : "Inactive" }}
+          </span>
+        </span>
+      </div>
+    </div>
+
+    <div class="meta">
+      <div class="collector-meta">
+        <span class="collector-meta-name">
+          <strong>Benchmark Set:</strong>
+        </span>
+        <span> #{{ collector.config.benchmarkSet }}</span>
+      </div>
+
+      <div class="collector-meta">
+        <span class="collector-meta-name">
+          <strong>Last Heartbeat:</strong>
+        </span>
+        <span>{{ collector.config.lastHeartbeatAt }}</span>
+      </div>
+
+      <div class="collector-meta">
+        <span class="collector-meta-name">
+          <strong>Date Added:</strong>
+        </span>
+        <span>{{ collector.config.dateAdded }}</span>
+      </div>
+    </div>
+
+    <div v-if="collector.request !== null">
+      <div class="table-collector-wrapper">
+        <table class="table-collector">
+          <caption>
+            current benchmark request
+          </caption>
+          <thead>
+            <tr class="table-header-row">
+              <th class="table-header-padding">Type</th>
+              <th class="table-header-padding">Created At</th>
+              <th class="table-header-padding">Sha / Tag</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td class="table-cell-padding">{{ collector.request.type }}</td>
+              <td class="table-cell-padding">
+                {{ collector.request.createdAt }}
+              </td>
+              <td class="table-cell-padding">{{ collector.request.tag }}</td>
+            </tr>
+          </tbody>
+        </table>
+      </div>
+
+      <div class="table-collector-wrapper">
+        <table class="table-collector" style="border-collapse: collapse">
+          <caption>
+            current benchmark jobs
+          </caption>
+          <thead>
+            <tr class="table-header-row">
+              <th class="table-header-padding">State</th>
+              <th class="table-header-padding">Started At</th>
+              <th class="table-header-padding">Backend</th>
+              <th class="table-header-padding">Profile</th>
+              <th class="table-header-padding">Dequeue Counter</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr v-for="j in collector.jobs">
+              <td class="table-cell-padding">{{ j.state }}</td>
+              <td class="table-cell-padding">{{ j.startedAt }}</td>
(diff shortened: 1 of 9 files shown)


site/frontend/src/pages/status_new/collector.vue
```diff
@@ -0,0 +1,231 @@
+<script setup lang="ts">
+import {CollectorConfigAndWork, CollectorConfig} from "./data";
+
+const props = defineProps<{
+  collector: CollectorConfigAndWork;
+}>();
+
+/* trick typescript into thinking props is used. We _have_ to define it else
+ * `collector` in the below code is untyped and thus fails to compile */
+<any>props;
```
Kobzol: Is this some kind of IDE thing? I don't see any warnings/errors without it, and we don't use it anywhere else, so it shouldn't be needed.

site/frontend/src/pages/status_new/data.ts
```diff
@@ -1,42 +1,55 @@
+export const BenchmarkRequestComplete = "completed";
```
Kobzol: I don't think that these constants are needed. This is one of the benefits of TS vs Rust, that it has string value as actual types. So you don't need the constant, you can just use the string literal (it's easier to read `"completed"` than `typeof BenchmarkRequestComplete`), and it will still be type-checked.

site/frontend/src/pages/status_new/page.vue
```diff
@@ -1,68 +1,122 @@
...
 
 const loading = ref(true);
-const dataNew: Ref<StatusResponse | null> = ref(null);
+const dataNew: Ref<
+  (StatusResponse & {collectorJobMap: CollectorJobMap}) | null
+> = ref(null);
 
-function statusClass(c: CollectorConfig): string {
-  return c.isActive ? "active" : "inactive";
+function pullRequestUrlAsHtml(reqType: BenchmarkRequestType): string {
```
Kobzol: We can actually use JSX for these simple components, I pushed a change.

site/frontend/src/pages/status_new/page.vue
```diff
@@ -1,68 +1,122 @@
...
+  if (reqType.type === ReleaseCommit) {
+    return "";
+  }
+  return `<a href="https://github.com/rust-lang/rust/pull/${reqType.pr}">#${reqType.pr}</a>`;
+}
+
+function formatDuration(milliseconds: number): string {
```
Kobzol: We already have this in `pages/status/page.vue`, let's move it to `src/utils/time.ts`.

site/frontend/src/pages/status_new/page.vue
```diff
@@ -1,68 +1,122 @@
...
+            <tbody>
+              <template v-for="req in dataNew.completed">
+                <tr>
+                  <td v-html="pullRequestUrlAsHtml(req.requestType)"></td>
+                  <td>{{ req.requestType.type }}</td>
+                  <td>{{ req.requestType.tag }}</td>
+                  <td>{{ req.status.state }}</td>
+                  <td>{{ req.status.completedAt }}</td>
+                  <td v-html="formatDuration(req.status.duration)"></td>
+                  <td>{{ req.errors.join("\n") }}</td>
```
Kobzol: `\n` won't really work in HTML, unless we wrap it in `<pre>` (which we should do, the old page does the same).

site/frontend/src/pages/status_new/collector.vue
Jamesbarford: If I run `npm run check` I get the following error;
```
> perf-rlo-ui@1.0.0 check
> vue-tsc -p tsconfig.json --noEmit && prettier --check src

src/pages/status_new/collector.vue:4:7 - error TS6133: 'props' is declared but its value is never read.

4 const props = defineProps<{
```

And doing, the following seems to fix the error.
```ts
<any>props;
```

site/frontend/src/pages/status_new/page.vue
Jamesbarford: 6170ded4cde2b4dc30804944040e7a5006e93492 :+1:

site/frontend/src/pages/status_new/page.vue
Jamesbarford: 6170ded4cde2b4dc30804944040e7a5006e93492 :+1:

site/frontend/src/pages/status_new/page.vue
Jamesbarford: I've reverted the JSX change; it wasn't rendering and causing a number of errors in the console (`ReferenceError: h is not defined`, which didn't make any sense).

The shape of the object looked rather odd when I logged it to the console, something like;
```
Proxy {
  target: {
      request: BenchmarkRequestType
  },
  handler: {
      // other properties
  }
}
```

Ill go over the rest of your feedback and come back to this.

site/frontend/src/pages/status_new/data.ts
Jamesbarford: Interesting. I generally avoid using magic strings because modifying them across multiple locations can be tedious and error-prone. I'm happy to update everything to magic strings if that's your preference, though I personally prefer symbolic constants

site/frontend/src/pages/status_new/page.vue
Jamesbarford: I needed to add an option to the `tsconfig.json` and then the problem went away; e8c254556e26bb1d1db268417efb4feb7e44fb8c

Kobzol: Oh, sorry, when using JSX, the source file has to import `h` from `vue` :laughing:

Kobzol:
Hmm, interesting, I don't need the `<any>props;` thing, and the import source change breaks my build. Could you please post your output of `npm list`?
```
perf-rlo-ui@1.0.0 /projects/personal/rust/rustc-perf/site/frontend
├── @babel/types@7.24.5
├── @parcel/compressor-brotli@2.12.0
├── @parcel/transformer-vue@2.12.0
├── @types/msgpack-lite@0.1.11
├── date-fns@2.30.0
├── highcharts@11.4.1
├── msgpack-lite@0.1.26
├── parcel@2.12.0
├── prettier@2.8.8
├── sass@1.77.2
├── typescript@5.4.5
├── uplot@1.6.30
├── vue-tsc@1.8.27
└── vue@3.4.27
```

site/frontend/src/pages/status_new/collector.vue
Jamesbarford: 763dbfc7f275209cbc7bc98edb037592f65d8cf9 used `props` in the `<template>`

klensy:
Something wrong with bench durations on https://perf.rust-lang.org/status.html page, for example cargo:
```
cargo-0.87.1		0m 00.397s	0m 00.391s
```
clearly bench didn't run 0.4 sec.

PR #2229, opened 2025-08-26
Track stdlib artifact size

Kobzol:
Since we stopped linking from librustc_driver to libstd.so dynamically on Linux, we lost the ability to track libstd size, which is a shame. This PR restores that ability. I removed the tracking of libtest though, I don't think that was super useful and it almost never changes anyway.


Jamesbarford (approved):
:+1: LGTM

PR #2230, opened 2025-08-27
Check latest commit SHA in the collector

Kobzol:
In the end I decided to just query GitHub directly, it's the easiest thing to do, and we should only do it only ~once per hour, so should be fine. It was much easier than adding a singleton database table and threading the git information from the website.

It permits some slight desync of the SHA in the collector and the website, but that also happened before and we ~never had issues with it.

I made the periodic check optional to avoid annoying restarts when running the collector locally.


Jamesbarford (approved):
LGTM :+1:

PR #2233, opened 2025-08-28
Feat; status page UI pt2

Jamesbarford:
- Rework SQL queries, for some reason there was a query there that we weren't using so I've deleted that. I also bring back parent requests along with their jobs. However we only parse the jobs if the `parent_active` flag is set. Which will save a bit of time.
- Delete all gnarly frontend transformations and push them to the server side. The remaining transformations are UI specific; namely drawing `----` under the requests that are in progress
- Sort the jobs for the collectors (in_progress, queued, completed)
- Removed `request` from collectors
- Add asterisk for a parent request which is completed but still needs benchmarking due to backfilling.

The ui now looks like this;
image

I've not added a filter for filtering in/out completed. I'd prefer to do that in another PR as that involves interactivity which I've not looked at how to do.

### DIFF
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -1098,54 +1098,54 @@ mod tests {
 
             assert!(status_page_data.in_progress.len() == 1);
             // we should have 2 jobs
-            assert!(status_page_data.in_progress[0].1.len() == 2);
+            assert!(status_page_data.in_progress[0].request.1.len() == 2);
             // the request should be in progress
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
Thanks! I'll merge this now, but after I examined the code in more detail, I think that the DB/backend side has ballooned in complexity a bit, I'll try to refactor it and send a PR for a review.

PR #2234, opened 2025-08-29
Refactor new status page

Kobzol:
I examined the DB/backend/frontend logic of the new status page after the recent improvements, and it seemed to me like enough cruft from various experiments has been accumulated that we should first simplify the logic a bit before we move forward. In particular I found it very hard to understand what the SQL queries were doing anymore, so I tried to make them simpler, and only send data to the frontend that it really currently needs.

In the last N completed requests I removed the array aggregation. We need both `benchmark` and `error` fields from the `error` table, so we would need to somehow post-process the arrays anyway, and it was also causing some issues with nested NULLs when I switched to `RIGHT JOIN` (because with the `LEFT JOIN`, the query was loading essentially every artifact and error from DB and then filtering it).

To test the changes, I implemented some helper test builders. I didn't yet port the existing tests to them, but I started writing new tests with them. The PR should be minus ~300 lines, if I subtract the new test code.

I simplified the frontend data model to match the old status page and avoid the enums, and I didn't yet bother adding back the status/type constant, but I'm fine with putting them back if you want.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -697,7 +697,7 @@ impl PostgresConnection {
                         errors.error
                     FROM completed
                     LEFT JOIN errors ON errors.tag = completed.tag
-                    -- Resort the requests, because the original order may be lost
+                    -- Re-sort the requests, because the original order may be lost
                     ORDER BY completed.completed_at DESC;
                 ")).await.unwrap(),
                 get_jobs_of_in_progress_benchmark_requests: conn.prepare(&format!("


database/src/pool/postgres.rs
```diff
@@ -667,12 +668,70 @@ impl PostgresConnection {
...
+                        RIGHT JOIN artifacts ON error.aid = artifacts.id
+                    )
+                    -- Select request duplicated for each pair of (benchmark, error)
+                    SELECT
+                        completed.*,
+                        errors.benchmark,
+                        errors.error
+                    FROM completed
+                    LEFT JOIN errors ON errors.tag = completed.tag
+                    -- Resort the requests, because the original order may be lost
```
Jamesbarford: Nit; "re-sort"

Kobzol:
The constants are in upper case on the Rust side :sweat_smile: But more generally, using constants won't save us from differences between the frontend and the backend, and the TypeScript type-checker checks that we didn't make a mistake on the frontend side (since CI is green, the constants are fine). But as I said, if you really find it better, I'll add them back :)

Kobzol:
Yeah, TS is relatively unique in this, as it allows modelling string *values* (not just the string type) in the typesystem. So in that case I find it better to not use the constants, as it's both less code and it's easier to read.

database/src/pool/postgres.rs
Kobzol: Updated the comment.

Jamesbarford (approved):
LGTM, nice tidy up :rocket:

PR #2236, opened 2025-09-02
Feat; filter collectors jobs by status

Jamesbarford:
Adds the ability to filter jobs in a collector by status;

filter jobs ui


site/frontend/src/pages/status_new/collector.vue
```diff
@@ -8,6 +9,15 @@ const props = defineProps<{
 function statusClass(c: CollectorConfig): string {
   return c.isActive ? "active" : "inactive";
 }
+
+const FILTERS = ["InProgress", "Queued", "Success", "Failure"];
```
Kobzol:
```suggestion
const FILTERS: BenchmarkJobStatus[] = ["InProgress", "Queued", "Success", "Failure"];
const ACTIVE_FILTERS: Ref<Set<BenchmarkJobStatus>> = ref(new Set(FILTERS));
```
And then make the parameter of `filterJobByStatus` be of type `BenchmarkJobStatus`. And it should show you a type check failure for the "Failure" filter (it should be called "Failed" :)

site/frontend/src/pages/status_new/collector.vue
```diff
@@ -71,22 +101,29 @@ function statusClass(c: CollectorConfig): string {
...
-            </td>
-            <td class="table-cell-padding">{{ job.backend }}</td>
-            <td class="table-cell-padding">{{ job.profile }}</td>
-            <td class="table-cell-padding">
-              {{ job.dequeCounter }}
-            </td>
-          </tr>
+          <template v-for="job in collector.jobs">
+            <tr
+              :key="`${job.requestTag}-${job.status}-${ACTIVE_FILTERS.has(
```
Kobzol: I don't think that you need to include the filter status in the key. In fact, the key should be unique, so ideally we should be using the job ID instead, otherwise two rows can have the same key and that can produce weird UI.

But I would just remove the key, I don't think we need it here.

site/frontend/src/pages/status_new/collector.vue
```diff
@@ -8,6 +9,15 @@ const props = defineProps<{
 function statusClass(c: CollectorConfig): string {
   return c.isActive ? "active" : "inactive";
 }
+
+const FILTERS = ["InProgress", "Queued", "Success", "Failure"];
+const ACTIVE_FILTERS = ref(new Set(FILTERS));
```
Kobzol: Sets don't fully work reactively with `ref`, because the watching is not "deep" (https://github.com/orgs/vuejs/discussions/10834).

I would suggest representing the state in a structurally more straightforward way, as we do elsewhere in the frontend:
```typescript
const filter: Ref<{[index: keyof BenchmarkJobStatus]: boolean}> = ref({
  InProgress: true,
  Queued: true,
  Success: true,
  Failed: false,
});
```
plus something like this:
```typescript
function toggleJobStatusFilter(status: BenchmarkJobStatus) {
  filter.value[status] = !filter.value[status];
}
```

You can then index into `filter[status]` to see if the status value is enabled or not.

Jamesbarford: The set did work, I have changed it to how you have suggested. Though I have filtered out both `Failed` and `Success`

Kobzol: Right, it did indeed track the inner modifications of the `Set`, but it's not ideal to track such inner state in Vue, the reactivity tracking works best when observing simple values and objects, and the set wasn't really needed here, so that's why I suggested removing it.

Btw I would prefer to show `Failed` jobs by default, we should make these highly visible and not hidden by default.

Jamesbarford: e1025b8232ffcd99580c92418cb78ca247f9c817 :+1:

PR #2238, opened 2025-09-03
Fix; queue ordering and errors formatting

Jamesbarford:
- Fixes the ordering of the queue.
- Fix formatting of errors, we now return a hashtable; I've formatted it to be `error_name: error\n`.
- Fix `createdAt` being used in the `Completed At` column (not sure if the column name is the one that needed changing or the value).

From the refactor we have lost the ability to have the parent that will be benchmarked appear "above the waterline".

### DIFF
--- a/site/frontend/src/pages/status_new/page.vue
+++ b/site/frontend/src/pages/status_new/page.vue
@@ -99,13 +99,7 @@ function formatStatus(status: BenchmarkRequestStatus): string {
 }
 
 function hasErrors(errors: Dict<string>) {
-  // This is a bit odd but if the error object has values the loop will be
-  // hit and thus return true.
-  for (const key in errors) {
-    console.log(key);
-    return true;
-  }
-  return false;
+  return Object.keys(errors).length !== 0;
 }
 
 function getErrorsLength(errors: Dict<string>) {


site/src/request_handlers/status_page_new.rs
```diff
@@ -15,13 +15,12 @@ pub async fn handle_status_page_new(ctxt: Arc<SiteCtxt>) -> anyhow::Result<statu
...
 
     // The queue contains any in-progress request(s) and then the following requests in queue order
-    // We reverse so that it starts with the request that will be benchmarked the latest
     let mut queue: Vec<status_new::BenchmarkRequest> = build_queue(&*conn, &index)
         .await?
         .into_iter()
         .map(|req| request_to_ui(&req, HashMap::new()))
         .collect();
-    queue.reverse();
+
```
Kobzol: Why did you remove this? The ordering on the status page should look like this (to match the current status page):
- Queued 3
- Queued 2
- Queued 1 (next to be benchmarked)
- In progress
- Most recently completed
- Second most recently completed

etc.

Kobzol (commented):
While you're at it, could you please copy-paste and adapt the way we show errors on the current status page? There is a clickable button that expands the errors if there are any. Without such a button, the page layout will probably look terrible if we just dump the error content directly into the table.

site/src/request_handlers/status_page_new.rs
Jamesbarford: Oh is it meant to be like this? 
image


I found that really confusing so removed the `reverse()` as I thought it made sense to see what was inprogress in bold with a line and then everything else underneath the line.

I found this more intuitive as it's actually reflective of what is happening;

image

Kobzol: We could change it, but I first wanted to match what is on the current status page. This is the only ordering that makes sense for the page to be called a "Timeline", IMO, as it accurately displays the requests ordered by time of start/completion. You can read it from top to bottom to see the ordering of requests, which I find quite nice.

Jamesbarford: Adding `started at` and `created at` would also be nice touches, and maybe formatting the date to be a bit more readable. But I'm happy to put it back to whatever is expected

Jamesbarford: > We could change it, but I first wanted to match what is on the current status page. This is the only ordering that makes sense for the page to be called a "Timeline", IMO, as it accurately displays the requests ordered by time of start/completion. You can read it from top to bottom to see the ordering of requests, which I find quite nice.

I'm sorry, I don't understand. I'll revert it.

Jamesbarford: e7a337a2ffa39cc5f4e5a76663d815bc5b904409

site/frontend/src/pages/status_new/page.vue
```diff
@@ -97,6 +98,21 @@ function formatStatus(status: BenchmarkRequestStatus): string {
   }
 }
 
+function hasErrors(errors: Dict<string>) {
+  // This is a bit odd but if the error object has values the loop will be
```
Kobzol: You already use `Object.keys(errors).length` in `getErrorsLength`, you can use the same here :)

Jamesbarford: 6913885c263c5febfb6cdb241c7b5a8098a4b681 :+1:

Kobzol (approved):
Thanks!

PR #2240, opened 2025-09-04
Stop hardcoding the master branch of rust-lang/rust

Kobzol:
- https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
- https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

Those were the only two occurrences that I found.

### DIFF
--- a/collector/src/compile/execute/rustc.rs
+++ b/collector/src/compile/execute/rustc.rs
@@ -193,9 +193,9 @@ fn checkout(artifact: &ArtifactId) -> anyhow::Result<()> {
             .current_dir("rust")
             .arg("fetch")
             .arg("origin")
-            .arg("master")
+            .arg("HEAD")
             .status()
-            .context("git fetch origin master")?;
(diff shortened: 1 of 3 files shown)


collector/src/compile/execute/rustc.rs
```diff
@@ -193,9 +193,9 @@ fn checkout(artifact: &ArtifactId) -> anyhow::Result<()> {
             .current_dir("rust")
             .arg("fetch")
             .arg("origin")
-            .arg("master")
+            .arg("HEAD")
```
Mark-Simulacrum, line 196: Does this update the reference that bootstrap is reading? My memory is hazy (and I'm not sure how to test) but I think I've seen git not update e.g. origin/master for this kind of command...

Kobzol: It fetches and hard resets to a specific SHA anyway, so I think that it should be fine ( I don't know why we fetch HEAD at all).

PR #2245, opened 2025-09-15
Fix; error table

Jamesbarford:
- Add Nullable `job_id` to the error table, remove unique constraint.
- Rename columns `benchmark` and `error` to `context` and `message`.
- Pass through `job_id` in collector code.


collector/src/bin/collector.rs
```diff
@@ -1497,6 +1513,7 @@ async fn run_job_queue_benchmarks(
                             artifact_row_id,
                             &format!("job:{}", benchmark_job.id()),
```
Kobzol: We can change the context now, because the job ID will be stored explicitly in the table now. Not sure what the context should be named in this case though, maybe just "unknown" or "general".

collector/src/runtime/mod.rs
```diff
@@ -89,7 +89,12 @@ pub async fn bench_runtime(
         if let Err(error) = result {
             eprintln!("collector error: {error:#}");
             tx.conn()
-                .record_error(collector.artifact_row_id, &step_name, &format!("{error:?}"))
+                .record_error(
+                    collector.artifact_row_id,
+                    &step_name,
+                    &format!("{error:?}"),
+                    None,
```
Kobzol: This should also receive the proper job ID.

database/src/pool/postgres.rs
```diff
@@ -387,6 +387,29 @@ static MIGRATIONS: &[&str] = &[
     r#"
     ALTER TABLE collector_config ADD COLUMN commit_sha TEXT NULL;
     "#,
+    r#"
+    CREATE TABLE IF NOT EXISTS error_new (
```
```suggestion
    CREATE TABLE error_new (
```

database/src/pool/postgres.rs
```diff
@@ -387,6 +387,29 @@ static MIGRATIONS: &[&str] = &[
...
+        id      SERIAL PRIMARY KEY,
+        aid     INTEGER NOT NULL,
```
```suggestion
    aid     INTEGER  NOT NULL REFERENCES artifact(id) ON DELETE CASCADE ON,
```
We want to keep the FK constraint, and also remove all errors when an artifact is purged from the DB.

database/src/pool/postgres.rs
```diff
@@ -387,6 +387,29 @@ static MIGRATIONS: &[&str] = &[
...
+        aid,
+        error,
+        benchmark
+    FROM
+        error;
+
+    DROP TABLE error;
+    ALTER TABLE error_new RENAME TO error;
+
+    CREATE INDEX IF NOT EXISTS error_artifact_idx ON error(aid);
```
```suggestion
    CREATE INDEX error_artifact_idx ON error(aid);
```

Kobzol (commented):
Thanks! Could you please update the docs MD file too?

collector/src/bin/collector.rs
```diff
@@ -1497,6 +1513,7 @@ async fn run_job_queue_benchmarks(
...
                             &format!("Error while benchmarking job {benchmark_job:?}: {error:?}"),
```
Kobzol, line 1515: This could become just `format!("{error:?}")`, as the context can be found out from the job ID. Although, if we do delete jobs, it might not hurt to leave the context also in the string form, and it would allow us not to load job info on the compare page.. Hmm, ok, then let's keep it.

database/schema.md
```diff
@@ -250,14 +250,21 @@ bors_sha    pr  parent_sha  complete  requested    include  exclude  runs  commi
...
 
-```
-sqlite> select * from error limit 1;
-aid         benchmark   error
-----------  ---         -----
-1           syn-1.0.89  Failed to compile...
-```
+Columns:
+
+* **id** (`BIGINT` / `SERIAL`): Primary key identifier for the job row;
```
Kobzol: `job` => `error` (twice in the sentence)

collector/src/bin/collector.rs
Kobzol: Let's use something more general please as the context, retry count exhaustion is not the only thing that can happen, even now we have four different kinds of permanent errors. Maybe just Job failure, to distinguish it from benchmark failures.

Kobzol:
Your updates are fine - I wrote that message *before* you did the MD docs update :sweat_smile: Maybe you just some some outdated GitHub comment? I'm fine with the current state of the PR :)

Tried it locally and it seems to work fine.

PR #2246, opened 2025-09-15
Fix; for query performance

Jamesbarford:
# Context
We wanted to see the `EXPLAIN ANALYSE` for' `completed`, `in_progress` and `dequeue` queries;

I've tested this with the following in the database;
- 91,303 requests
- 70,001 artifacts
- 40,047 errors
- 70,122 jobs

For small numbers of items it is _always_ faster to use a sequence scan as it avoids loading an index. Postgres seems to be smart enough to know when to switch from a sequence scan to an index scan depending on how many entries there are.

The below indexes are created off the basis of wanting to avoid _any_ sequence scan where possible.

## Completed
This needs an index on the column `completed_at` otherwise we get a full sequence scan on the `completed_at` column c799d67d230c73fe64b5102e3bbbf3d2cea2bc31. Similarly the `error` table needs an index on `aid` to prevent a full sequence scan of the error table when we do a join on the `artifact.id` 6a6dadb1d87b94c2b8a82dc70b38a720f3a494a1.

<details>
<summary>BEFORE</summary>

image
</details>

<details>
<summary>AFTER</summary>
image

</details>

## In Progress

This query needs no new indexes. We do a full sequence scan of the job queue table. I'm not sure yet if this is avoidable or acceptable. I'm also not exactly sure what we are wanting from the query. Below is the current `EXPLAIN ANALYSE` output

<details>
<summary>BEFORE - EXPLAIN ANALYSE output</summary>
image
</details>

If the query is to get the jobs of currently `in_progress` requests I have updated the query in this commit; fe911551812af11ae96e668cfa2ece133d228503

Which produces a more optimised;

<details>
<summary>AFTER - EXPLAIN ANALYSE output</summary>

image

</details>

## Dequeue

This query needs no new indexes.

<details>
<summary>EXPLAIN ANALYSE output</summary>
image
</details>

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -749,7 +749,7 @@ impl PostgresConnection {
                         SELECT tag FROM parents
                     )
                     -- Only get the jobs of in_progress requests
-                    SELECT * FROM job_queue WHERE job_queue.request_tag IN (SELECT * FROM requests)
+                    SELECT * FROM job_queue INNER JOIN requests ON job_queue.request_tag = requests.tag
                 ")).await.unwrap(),
             }),
             conn,


database/src/pool/postgres.rs
```diff
@@ -410,6 +410,11 @@ static MIGRATIONS: &[&str] = &[
 
     CREATE INDEX error_artifact_idx ON error(aid);
     "#,
+    // For completed requests we take the last N completed. As the total number
+    // of requests grows to make things fast we need an index on the completed_at
+    r#"
+    CREATE INDEX IF NOT EXISTS benchmark_request_completed_idx ON benchmark_request(completed_at);
```
```suggestion
    CREATE INDEX benchmark_request_completed_idx ON benchmark_request(completed_at);
```

database/src/pool/postgres.rs
```diff
@@ -743,10 +748,8 @@ impl PostgresConnection {
                         UNION
                         SELECT tag FROM parents
                     )
-                    SELECT job_queue.*
-                    FROM requests
-                    -- Only get requests that have some jobs
-                    RIGHT JOIN job_queue on job_queue.request_tag = requests.tag
+                    -- Only get the jobs of in_progress requests
+                    SELECT * FROM job_queue WHERE job_queue.request_tag IN (SELECT * FROM requests)
```
Kobzol: Good find! It is indeed better to select the jobs and filter them by the requests, rather than doing it in the opposite direction. Not sure what I was thinking adding a `RIGHT JOIN` here...

What you wrote is essentially
```sql
SELECT *
FROM job_queue
JOIN requests ON job_queue.request_tag = requests.tag
```
it seems that it generates the same plan. I'd still slightly prefer the `JOIN` version though, as I have more confidence that an index will be used for `JOIN ... ON ...` rather than for `WHERE ... IN ...`.

Kobzol (commented):
Thank you, good finds! Can you clean up the commits a bit please, and apply the nits?

database/src/pool/postgres.rs
Jamesbarford: b9fd3466b56b0266e947f3e16136601589f1b681 :+1:

database/src/pool/postgres.rs
Jamesbarford: b880a243b3eda94ec2749fed7c6b6ab80e298ad6 :+1: (I've used `INNER JOIN` explicitly)

PR #2247, opened 2025-09-15
add triage for 2025-09-15

panstromek:


triage/2025/2025-09-15.md
```diff
@@ -0,0 +1,125 @@
+# 2025-09-15 Triage Log
+
+Difficult week to interpret, because a positive change in [#145910](https://github.com/rust-lang/rust/pull/145910) skewes the results more negative in our benchmarks. Overall result is probably still slightly negative, because there's more work from added features. On the other hand, we also have a nice improvement in reducing the number of query dependencies in compiler's incremental system in [#145186](https://github.com/rust-lang/rust/pull/145186).
```
lqd: do you mean this?

```suggestion
Difficult week to interpret, because a positive change in [#145910](https://github.com/rust-lang/rust/pull/145910) skews the more negative results in our benchmarks. Overall result is probably still slightly negative, because there's more work from added features. On the other hand, we also have a nice improvement in reducing the number of query dependencies in compiler's incremental system in [#145186](https://github.com/rust-lang/rust/pull/145186).
```

triage/2025/2025-09-15.md
```diff
@@ -0,0 +1,125 @@
...
+|:----------------------------------:|:-----:|:--------------:|:-----:|
+| Regressions ❌ <br /> (primary)    | 0.3%  | [0.2%, 0.4%]   | 3     |
+| Regressions ❌ <br /> (secondary)  | 0.1%  | [0.1%, 0.2%]   | 3     |
+| Improvements ✅ <br /> (primary)   | -     | -              | 0     |
+| Improvements ✅ <br /> (secondary) | -0.1% | [-0.1%, -0.1%] | 1     |
+| All ❌✅ (primary)                 | 0.3%  | [0.2%, 0.4%]   | 3     |
+
+Some small regressions. I suspect the primary regressions in `unicode-normalization` come from https://github.com/rust-lang/rust/pull/144765, because it contains a few giant tables with a lot of range patterns. The other candidate would be https://github.com/rust-lang/rust/pull/146178 but I don't see anything that would be related in that benchmark.
+
+Either way, together with the secondary benchmarks, all of those are tiny changes (0.002s) on tiny benchmarks, so I don't think this worth more investigation.
```
lqd:
```suggestion
Either way, together with the secondary benchmarks, all of those are tiny changes (0.002s) on tiny benchmarks, so I don't think this is worth more investigation.
```

triage/2025/2025-09-15.md
```diff
@@ -0,0 +1,125 @@
...
+
+| (instructions:u)                   | mean  | range          | count |
+|:----------------------------------:|:-----:|:--------------:|:-----:|
+| Regressions ❌ <br /> (primary)    | 0.1%  | [0.1%, 0.1%]   | 1     |
+| Regressions ❌ <br /> (secondary)  | 0.1%  | [0.1%, 0.3%]   | 8     |
+| Improvements ✅ <br /> (primary)   | -     | -              | 0     |
+| Improvements ✅ <br /> (secondary) | -0.1% | [-0.2%, -0.1%] | 2     |
+| All ❌✅ (primary)                 | 0.1%  | [0.1%, 0.1%]   | 1     |
+
+Slight regression in doc benchmarks, similar to [pre-merge results](https://github.com/rust-lang/rust/pull/144847#issuecomment-3189779391). Expected because of more const handling (looks like more time is spent in const related queries from brief look at the detailed results of hellp-world: https://perf.rust-lang.org/detailed-query.html?commit=5c11fb842afc9876e985886c30292a227caee632&benchmark=helloworld-doc&scenario=full&base_commit=a171994070dc18c3a32fc1aa2d98cf03ae96b63e)
```
lqd:
```suggestion
Slight regression in doc benchmarks, similar to [pre-merge results](https://github.com/rust-lang/rust/pull/144847#issuecomment-3189779391). Expected because of more const handling (looks like more time is spent in const related queries from a brief look at the detailed results of hello-world: https://perf.rust-lang.org/detailed-query.html?commit=5c11fb842afc9876e985886c30292a227caee632&benchmark=helloworld-doc&scenario=full&base_commit=a171994070dc18c3a32fc1aa2d98cf03ae96b63e)
```

triage/2025/2025-09-15.md
panstromek: Not sure. I wanted to say that the change looks more negative in our benchmarks than it really is in general (outside of the benchmark suite). I thought it'd be correct to say that "the results are skewed," but I guess it sounds awkard in that sentence.

lqd: maybe something like the change "performs a bit worse in our benchmarks than it would in the real world"?

lqd (approved):
r=me with the last intro clarification you’ll prefer, and the commits squashed (no need to have separate typo commits with me as a coauthor)

triage/2025/2025-09-15.md
panstromek: ok, let's do that one, sounds more straightforward. Thanks.

```suggestion
Difficult week to interpret, because a positive change in [#145910](https://github.com/rust-lang/rust/pull/145910) performs a bit worse in our benchmarks than it would in the real world. Overall result is probably still slightly negative, because there's more work from added features. On the other hand, we also have a nice improvement in reducing the number of query dependencies in compiler's incremental system in [#145186](https://github.com/rust-lang/rust/pull/145186).
```

PR #2248, opened 2025-09-17
Fix; Allow postgres to persist data

Jamesbarford:
I've lost countless useful database setups due to me turning off my computer turning it back on again and then losing the data as it's only stored in RAM.

Removing this line has a negligible performance impact on tests.


Kobzol (approved):
Ok, I agree that the perf. gain is not so large for this to be worth it, the data can be indeed useful for long-term testing.

PR #2250, opened 2025-09-19
Estimate when requests will end in the new status page

Kobzol:
Alternative to https://github.com/rust-lang/rustc-perf/pull/2249.


Kobzol:
Should be just four after https://github.com/rust-lang/rustc-perf/pull/2252 ^^ As long as the query count doesn't scale with the number of fetched requests, as it did before, I think it'll be fine.

PR #2251, opened 2025-09-19
Change `./target/release` example commands to `cargo run --release`

Kobzol:
Discussed in https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/run.20with.20broken.20benchmarks.20self-profile/with/540398362.

CC @lcnr


nnethercote (approved):
Seems fine, though I couldn't tell from the Zulip thread what problem the change is solving?

PR #2252, opened 2025-09-20
Do not use the benchmark index when computing benchmark request queue

Kobzol:
This removes one half of the usage of the benchmark index. Now when loading the status page or computing the queue for deciding which requests to enqueue, we only load pending requests from the database, together with the information whether their parent is completed or not, and we no longer have to load all completed requests from the DB.

This leaves only one usage of the benchmark index, when determining whether to add master/release requests into the database.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -5,9 +5,10 @@ use crate::{
     BenchmarkJobConclusion, BenchmarkJobStatus, BenchmarkRequest, BenchmarkRequestIndex,
     BenchmarkRequestStatus, BenchmarkRequestType, BenchmarkRequestWithErrors, BenchmarkSet,
     CodegenBackend, CollectionId, CollectorConfig, Commit, CommitType, CompileBenchmark, Date,
-    Index, Profile, QueuedCommit, Scenario, Target, BENCHMARK_JOB_STATUS_FAILURE_STR,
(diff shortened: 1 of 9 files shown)


database/src/pool/postgres.rs
```diff
@@ -1185,13 +1199,13 @@ where
...
             None => {
                 self.conn()
-            .query_opt("insert into artifact (name, date, type) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING RETURNING id", &[
-                &info.name,
-                &info.date,
-                &info.kind,
-            ])
-            .await
-            .unwrap();
+                    .query_opt("insert into artifact (name, date, type) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING RETURNING id", &[
```
Jamesbarford, line 1205: Nit; capitalisation of SQL keywords. It's vastly more readable and the style we've been using with all of the new queries.
```suggestion
                    .query_opt("INSERT INTO artifact (name, date, type) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING RETURNING id", &[
```

Jamesbarford (changes requested):

database/src/pool/postgres.rs
Kobzol: This is an old query that I didn't modify :sweat_smile: But why not, a nice drive-by cleanup :)

Kobzol: Huh, this was `rustfmt` reformatting some old code, interesting.

PR #2253, opened 2025-09-22
Reload benchmark request and DB index after a request is completed

Kobzol:
And also send the GH comments. This should hopefully conclude the request completion logic in the new job queue system.

### DIFF
--- a/site/src/job_queue/mod.rs
+++ b/site/src/job_queue/mod.rs
@@ -11,7 +11,7 @@ use database::{
 };
 use parking_lot::RwLock;
 use std::{str::FromStr, sync::Arc};
-use tokio::time::{self, Duration};
+use tokio::time::{self, Duration, MissedTickBehavior};
 
 pub fn run_new_queue() -> bool {
     std::env::var("RUN_CRON")
@@ -372,6 +372,8 @@ async fn cron_enqueue_jobs(ctxt: &SiteCtxt) -> anyhow::Result<()> {
 /// Entry point for the cron job that manages the benchmark request and job queue.
 pub async fn cron_main(site_ctxt: Arc<RwLock<Option<Arc<SiteCtxt>>>>, run_interval: Duration) {
     let mut interval = time::interval(run_interval);
+    interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
     let ctxt = site_ctxt.clone();
 
     loop {


site/src/job_queue/mod.rs
```diff
@@ -278,28 +291,89 @@ async fn process_benchmark_requests(
...
-    let index = conn.load_benchmark_request_index().await?;
+    let index = ctxt.known_benchmark_requests.load();
 
+    let mut requests_inserted = false;
     // Put the master commits into the `benchmark_requests` queue
-    create_benchmark_request_master_commits(site_ctxt, &*conn, &index).await?;
+    requests_inserted |= create_benchmark_request_master_commits(ctxt, &*conn, &index).await?;
     // Put the releases into the `benchmark_requests` queue
-    create_benchmark_request_releases(&*conn, &index).await?;
+    requests_inserted |= create_benchmark_request_releases(&*conn, &index).await?;
```
Jamesbarford, line 307: If we have created `master` commits using `|=` will mean we will not create release commits meaning we potentially miss adding elements to the queue. If I'm reading it correctly;
```rs
requests_inserted |= create_benchmark_request_master_commits(ctxt, &*conn, &index).await?;
```
Evaluates to `true`. Then as `requests_inserted` is now true it will short circuit and not call `create_benchmark_request_releases`?

Though we are not wanting to insert release commits if we insert master this comment can be ignored.

Jamesbarford (changes requested):

site/src/job_queue/mod.rs
Kobzol: `|` (bit or) is not short-circuiting in Rust, same as in C (https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=25f0365af521a2da542fe47bb1e07be1) :)

Jamesbarford: It's a logical OR but with Bit OR-like assignment :melting_face: https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.arith-logic

I was thinking it would expand to;

```rs
requests_inserted = requests_inserted || create_benchmark_request_releases(&*conn, &index).await?;
```

But it does not!

Kobzol: Yeah. I think it's easier to consider it to be a bit or, rather than a logical or, in the end. If you treat the bool as u1, it will work that way.

PR #2254, opened 2025-09-22
Fix; `job_queue` insert only if tag exists in `benchmark_request`

Jamesbarford:
Make the parent sha null if it does not refer to a tag in the `benchmark_request` table. This prevents a foreign key violation when subsequently trying to create jobs for back filling.

### DIFF
--- a/site/src/job_queue/mod.rs
+++ b/site/src/job_queue/mod.rs
@@ -334,9 +334,7 @@ async fn cron_enqueue_jobs(ctxt: &SiteCtxt) -> anyhow::Result<()> {
                     false,
                     *pr,
                     sha.clone().expect("Completed try commit without a SHA"),
-                    parent_sha
-                        .clone()
-                        .expect("Completed try commit without a parent SHA"),
+                    parent_sha.clone(),
                 ),
                 BenchmarkRequestType::Master {
                     pr,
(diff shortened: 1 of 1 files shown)


database/src/pool/postgres.rs
```diff
@@ -1580,6 +1580,22 @@ where
...
+        // @TODO delete this when the new system has been running. This code
+        // prevents a foreign key violation in the job_queue table. This can
+        // happen as the parent_sha may not be in the range of initial commits
+        // we load into the database when the job queue first starts and inserts
+        // it's first batch of requests.
+        let row = self
+            .conn()
+            .query_one(
+                "SELECT tag FROM benchmark_request WHERE tag = $1",
+                &[&benchmark_request.parent_sha()],
```
Kobzol: If the parent SHA is NULL (which will happen for try builds), `tag = NULL` will return `NULL`, which will be interpreted as `false`, right? Then the `query_one` call will fail.

Kobzol: The same would happen if the parent SHA was actually missing in the DB. We'll need to use e.g. `query`, not `query_one`.

Jamesbarford: Judging by a raft of tests failing :sweat:. Yes.

Jamesbarford: I don't know if I like what I have done here but it does achieve what we want and should only be temporary; 68d0d8710e3793cca15fab5084b1e954ebd05aad

Jamesbarford: This will most likely mean that all `parent_sha`'s will be null when initially starting the CRON. At least this is what I have observed locally.

database/src/pool.rs
```diff
@@ -585,6 +585,10 @@ mod tests {
         run_postgres_test(|ctx| async {
             let db = ctx.db();
 
+            // Releases don't need parent shas
+            ctx.insert_release_request("parent-sha-1").await;
```
Kobzol: We don't currently check against it, but a release request cannot be the parent of anything. So I'd rather not have this situation in tests.

PR #2257, opened 2025-09-24
Fix; enqueuing a job that has already been enqueued

Jamesbarford:
Fixes the error message below which happens when trying to enqueue a job that has already been enqueued. We only use the return value from this query in tests.

```
[2025-09-24T08:51:57Z ERROR site::job_queue] Cron job failed to execute: failed to insert benchmark_job

    Caused by:
        query returned an unexpected number of rows
```

### DIFF
--- a/database/src/pool.rs
+++ b/database/src/pool.rs
@@ -224,7 +224,7 @@ pub trait Connection: Send + Sync {
         commit_date: DateTime<Utc>,
     ) -> anyhow::Result<()>;
 
-    /// Add a benchmark job to the job queue and return its ID.
+    /// Add a benchmark job to the job queue.
     async fn enqueue_benchmark_job(
         &self,
         request_tag: &str,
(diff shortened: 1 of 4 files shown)


Kobzol (approved):

database/src/pool.rs
```diff
@@ -224,15 +224,15 @@ pub trait Connection: Send + Sync {
         commit_date: DateTime<Utc>,
     ) -> anyhow::Result<()>;
 
-    /// Add a benchmark job to the job queue and return its ID.
+    /// Add a benchmark job to the job queue.
```
```suggestion
    /// Add a benchmark job to the job queue and returns its ID, if it was not already in the DB previously.
```
Actually, this is still nice to document.

Kobzol (approved):

database/src/pool.rs
Jamesbarford: 8a547ff922a82eb722f33839598ed002268db930 :+1:

PR #2263, opened 2025-10-03
Fix; collector offline status

Jamesbarford:
Adds an `Offline` tag if the collector has been inactive to 1 hour (arbitrarily picked).

Screenshot from 2025-10-03 11-47-43

### DIFF
--- a/site/frontend/src/pages/status_new/collector.vue
+++ b/site/frontend/src/pages/status_new/collector.vue
@@ -8,10 +8,6 @@ const props = defineProps<{
   collector: CollectorConfig;
 }>();
 
-function statusClass(c: CollectorConfig): string {
-  return c.isActive ? "active" : "inactive";
-}
-
 const FILTERS: BenchmarkJobStatus[] = [
   "InProgress",
   "Queued",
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/status_new/collector.vue
```diff
@@ -1,5 +1,6 @@
-<script setup lang="ts">
-import {ref, Ref} from "vue";
+<script setup lang="tsx">
+import {h, ref, Ref} from "vue";
+import {parseISO, differenceInHours} from "date-fns";
 import {formatISODate} from "../../utils/formatting";
 import {CollectorConfig, BenchmarkJobStatus} from "./data";
```
Kobzol, line 6: The `statusClass` function is unused now.

site/frontend/src/pages/status_new/collector.vue
```diff
@@ -40,6 +41,37 @@ function formatJobStatus(status: BenchmarkJobStatus): string {
...
+        statusText = "Offline";
+        statusClass = "inactive";
+      } else {
+        statusText = "Active";
+        statusClass = "active";
+      }
+      break;
+    case false:
+      statusText = "Inactive";
+      statusClass = "inactive";
```
Kobzol, line 61: Not that we should see this status often, but I'd use a grey color for "inactive", and a red color for "offline".

Kobzol (commented):
One hour is probably fine, because the individual jobs should now be much faster than one hour.

site/frontend/src/pages/status_new/collector.vue
Jamesbarford: image

Kobzol: Yeah, something like that, looks good :)

Jamesbarford: 33704e279e304a2825a22598a3b28a409a321580 :+1:

site/frontend/src/pages/status_new/collector.vue
Jamesbarford: 33704e279e304a2825a22598a3b28a409a321580 :+1:

PR #2264, opened 2025-10-03
Fix; Runtime & Rustc benchmarks to run once

Jamesbarford:
Adds a `kind` column to the `job_queue` for compiletime, runtime and rustc then removes those as BenchmarkSetMembers.

This is to fix running them multiple times for when jobs are created through a Cartesian product.

### DIFF
--- a/site/src/job_queue/mod.rs
+++ b/site/src/job_queue/mod.rs
@@ -6,8 +6,8 @@ use crate::load::{partition_in_place, SiteCtxt};
 use chrono::Utc;
 use collector::benchmark_set::benchmark_set_count;
 use database::{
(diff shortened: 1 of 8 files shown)


site/src/job_queue/mod.rs
```diff
@@ -234,6 +234,30 @@ pub async fn enqueue_benchmark_request(
 
     // Target x benchmark_set x backend x profile -> BenchmarkJob
     for target in Target::all() {
+        // enqueue Runtime job and Rustc job, for each backend
+        for &backend in backends.iter() {
```
Kobzol: Let's just do rustc only for LLVM + x64 for now, and runtime benchmarks only for LLVM. We don't support multiple codegen backends for either anyway at this moment.

Jamesbarford: e335e3657f3c881f93c31d065fd1e13cc9046869 :+1:

PR #2265, opened 2025-10-06
Remove benchmark request and jobs when an artifact is purged

Kobzol:
Another item from the long tail list of items to be modified for the new system. Purging happens when we manually want to remove an artifact from the DB, usually because its artifacts are missing on CI.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1582,12 +1582,19 @@ where
         // thanks to ON DELETE CASCADE.
         let info = aid.info();
         self.conn()
-            .execute("delete from artifact where name = $1", &[&info.name])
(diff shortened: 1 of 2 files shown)


database/src/pool/postgres.rs
```diff
@@ -1592,6 +1592,13 @@ where
             )
             .await
             .unwrap();
+        self.conn()
+            .execute(
+                "delete from benchmark_request where tag = $1",
```
Jamesbarford: Nit;
```suggestion
                "DELETE FROM benchmark_request WHERE tag = $1",
```

(although  this is amongst some older SQL without keywords in caps to up to you)

Kobzol: Changed the casing in the whole function.

Jamesbarford (approved):
LGTM :+1:

PR #2266, opened 2025-10-06
Fix job queue migration

Kobzol:
Migrations that add a new `NOT NULL` column should always contain a `DEFAULT` value.


Jamesbarford (approved):
LGTM :+1:

PR #2269, opened 2025-10-07
Fix; 12 chars for a tag and a more responsive collectors grid

Jamesbarford:
This should make the page a bit more user friendly;
- Stop the table bleeding out of the collector card element
- Set the tag length to 12 characters

### DIFF
--- a/site/frontend/src/pages/status_new/collector.vue
+++ b/site/frontend/src/pages/status_new/collector.vue
@@ -149,7 +149,7 @@ function ActiveStatus({collector}: {collector: CollectorConfig}) {
           <template v-for="job in collector.jobs">
             <tr v-if="ACTIVE_FILTERS[job.status]">
               <td class="table-cell-padding">
-                {{ job.requestTag }}
+                {{ job.requestTag.slice(0, 12) }}
               </td>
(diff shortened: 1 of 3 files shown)


site/frontend/src/pages/status_new/collector.vue
```diff
@@ -149,7 +149,7 @@ function ActiveStatus({collector}: {collector: CollectorConfig}) {
           <template v-for="job in collector.jobs">
             <tr v-if="ACTIVE_FILTERS[job.status]">
               <td class="table-cell-padding">
-                {{ job.requestTag }}
+                {{ job.requestTag.slice(0, 12) }}
```
Kobzol: This won't really work for beta releases, which look something like `1.91.0-beta.3` and just barely won't make 12 chars. Let's create a shared function for rendering the tag that will only shorten it if it's a 40-length string (a commit SHA).

Jamesbarford: 4df6c63b4a885e492d2ac546f23a01ef99e75333 :+1:

Kobzol (approved):
Thanks!

PR #2271, opened 2025-10-07
Allow running the website with SQLite

Kobzol:


site/src/job_queue/mod.rs
```diff
@@ -366,6 +366,9 @@ async fn process_benchmark_requests(
 /// finishes completed benchmark requests.
 async fn perform_queue_tick(ctxt: &SiteCtxt) -> anyhow::Result<()> {
     let mut conn = ctxt.conn().await;
+    if !conn.supports_job_queue() {
```
Jamesbarford, line 369: I'm not sure of the use case you are thinking of? Given my lack of context I feel a warning message saying that the queue tick is running but not doing anything is needed

Kobzol: If you run the website locally with SQLite, it's great if it wouldn't crash :laughing: I don't think we need a warning, SQLite just fundamentally doesn't support the job queue right now (but it can be added later). For local usage the job queue is not needed anyway.

Jamesbarford: I was thinking a warning might help us catch a situation where we deploy with the wrong database?

Kobzol: I have no idea how that would happen, tbh. I wanted to avoid spam in the console when you run the website locally, which is why I didn't add any warning. A better solution would be to avoid running the cron job altogether, but that's a bit annoying to detect in `main`.

PR #2191, opened 2025-07-06
use stable extract_if: since 1.87

klensy:
https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.extract_if

PR #2273, opened 2025-10-07
Bump MSRV to 1.88.0

Kobzol:
Unblocks https://github.com/rust-lang/rustc-perf/pull/2191.


Kobzol:
It is related, because I bumped rustc on CI, which triggered new lints :) Already fixed them.

PR #2276, opened 2025-10-10
Get host tuple from rustc used for benchmarking/profiling, if possible

Kobzol:
Another attempt to fix https://github.com/rust-lang/rust/pull/147515.

r? @lqd

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -794,16 +794,54 @@ fn main_result() -> anyhow::Result<i32> {
         runtime: &runtime_benchmark_dir,
     };
 
-    let host_target_tuple =
-        String::from_utf8(command_output(Command::new("rustc").arg("--print=host-tuple"))?.stdout)?
-            .trim()
-            .to_string();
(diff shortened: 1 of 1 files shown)


collector/src/bin/collector.rs
```diff
@@ -794,16 +794,54 @@ fn main_result() -> anyhow::Result<i32> {
...
+        | Commands::AddCollector { .. }
+        | Commands::BenchmarkJobQueue { .. } => None,
+    };
+
+    let host_target_tuple = match used_rustc {
+        Some(rustc) => get_host_tuple_from_rustc(&rustc),
+        None => get_host_tuple_from_rustc("rustc"),
+    };
+    // We only unwrap the host tuple in places where we actually need it, to avoid panicking if it
+    // is missing, but we don't really need it.
```
lqd, line 833: I guess technically we wouldn't have needed to run the command in the first place, not just avoid checking its result. I guess it only will make 1 rustc call per rustc-perf command so it's not that bad, and we'll most likely actually need that 1 invocation.

PR #2282, opened 2025-10-14
new statuses for collector

Jamesbarford:
"inactive" does have a capital "I". I tried adding emojis but it looked a bit odd with a colourful background. 

Screenshot from 2025-10-14 09-45-59
Screenshot from 2025-10-14 09-44-31
Screenshot from 2025-10-14 09-44-06
Screenshot from 2025-10-14 09-44-52

### DIFF
--- a/site/frontend/src/pages/status_new/collector.vue
+++ b/site/frontend/src/pages/status_new/collector.vue
@@ -59,7 +59,7 @@ function ActiveStatus({collector}: {collector: CollectorConfig}) {
         const allJobsComplete = collector.jobs.every(
           (job) => job.status === "Failed" || job.status === "Success"
         );
-        if (allJobsComplete || collector.jobs.length === 0) {
+        if (allJobsComplete) {
           statusText = "Waiting";
           statusClass = "waiting";
         } else {


site/frontend/src/pages/status_new/collector.vue
```diff
@@ -43,23 +43,31 @@ function ActiveStatus({collector}: {collector: CollectorConfig}) {
...
       if (hourDiff >= maxInactivityHours) {
         statusText = "Offline";
         statusClass = "offline";
       } else {
-        statusText = "Active";
-        statusClass = "active";
+        const allJobsComplete = collector.jobs.every(
+          (job) => job.status === "Failed" || job.status === "Success"
+        );
+        if (allJobsComplete || collector.jobs.length === 0) {
```
Kobzol:
```suggestion
        if (allJobsComplete) {
```
The second pat of the condition is not necessary.

Jamesbarford: I was under the impression that without it, it could be in the "Benchmarking" state which would be inaccurate?

Kobzol: `.every` on an empty array returns true (try evaluating e.g. `[].every(x => x)` in the browser console).

Jamesbarford: cde492a7e9bc8505de020e941e9eeaff758f66a1 :exploding_head:

Kobzol: It's essentially a set operation - for all elements in the array, the property has to hold. If the set is empty, this is automatically always true, becase there is simply no element for which it could not hold :)

PR #2283, opened 2025-10-14
Job completion progress in the timeline

Jamesbarford:
Adds a job completion progress counter to the timeline table.

image

### DIFF
--- a/site/frontend/src/pages/status_new/page.vue
+++ b/site/frontend/src/pages/status_new/page.vue
@@ -13,9 +13,9 @@ import {useExpandedStore} from "../../utils/expansion";
 import {
   BenchmarkRequest,
   BenchmarkRequestStatus,
-  BenchmarkJob,
   CollectorConfig,
   StatusResponse,
+  isJobComplete,
 } from "./data";
 import Collector from "./collector.vue";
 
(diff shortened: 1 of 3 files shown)


site/frontend/src/pages/status_new/page.vue
```diff
@@ -122,6 +123,32 @@ function PullRequestLink({request}: {request: BenchmarkRequest}) {
   );
 }
 
+// This works off the assumption that all of the collectors are working on the
+// same request.
+function getJobCompletion(
+  req: BenchmarkRequest,
+  collectors: CollectorConfig[]
+) {
+  const sampleJob = collectors?.[0]?.jobs?.[0];
```
Kobzol: I don't think that we need such assumption, it might break if we had backfilled jobs, and the first collector can be offline/inactive/not relevant. What about something like this:
```typescript
const jobs = collectors.flatMap(c => c.jobs).filter(j => j.requestTag === req.tag);
if (jobs.length === 0) {
    return "";
}
const completed = jobs.reduce((acc, job) => {
    if (job.status === "Failed" || job.status === "Success") {
      acc += 1;
    }
    return acc;
  }, 0);
return `${completed} / ${jobs.length}`;
```

Also, we already have the same check for "Failed" / "Success" when determining the collector status. It would be nice to separate that logic into a shared function, something like `function isJobCompleted(job: BenchmarkJob): boolean`.

Jamesbarford: 9f40a841f06c0bf5c09d66353d4fd23bdc01a47f :+1:

Kobzol (approved):
The comment on top of the function can be removed now.

Kobzol (approved):

PR #2285, opened 2025-10-14
Add target to runtime benchmarks

Kobzol:
So that we can run runtime benchmarks on ARM machines.

### DIFF
--- a/database/src/bin/postgres-to-sqlite.rs
+++ b/database/src/bin/postgres-to-sqlite.rs
@@ -397,11 +397,11 @@ impl Table for RuntimePstatSeries {
     }
 
     fn postgres_select_statement(&self, _since_weeks_ago: Option<u32>) -> String {
-        "select id, benchmark, metric from ".to_string() + self.name()
+        "SELECT id, benchmark, target, metric FROM ".to_string() + self.name()
     }
 
     fn sqlite_insert_statement(&self) -> &'static str {
-        "insert into runtime_pstat_series (id, benchmark, metric) VALUES (?, ?, ?)"
+        "INSERT INTO runtime_pstat_series (id, benchmark, target, metric) VALUES (?, ?, ?, ?)"
     }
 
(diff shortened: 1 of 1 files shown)


database/src/pool/postgres.rs
```diff
@@ -660,8 +666,8 @@ impl PostgresConnection {
                     select name, category
                     from benchmark
                 ").await.unwrap(),
-                insert_runtime_pstat_series: conn.prepare("insert into runtime_pstat_series (benchmark, metric) VALUES ($1, $2) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
-                select_runtime_pstat_series: conn.prepare("select id from runtime_pstat_series where benchmark = $1 and metric = $2").await.unwrap(),
+                insert_runtime_pstat_series: conn.prepare("insert into runtime_pstat_series (benchmark, target, metric) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
```
Jamesbarford: Nit;
```suggestion
                insert_runtime_pstat_series: conn.prepare("INSERT INTO runtime_pstat_series (benchmark, target, metric) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
```

database/src/pool/postgres.rs
```diff
@@ -660,8 +666,8 @@ impl PostgresConnection {
...
+                select_runtime_pstat_series: conn.prepare("select id from runtime_pstat_series where benchmark = $1 and target = $2 and metric = $3").await.unwrap(),
```
Jamesbarford: Nit;
```suggestion
                select_runtime_pstat_series: conn.prepare("SELECT id FROM runtime_pstat_series WHERE benchmark = $1 AND target = $2 AND metric = $3").await.unwrap(),
```

Jamesbarford (approved):
LGTM :+1:

PR #2293, opened 2025-10-20
Update job queue loop to avoid early exit

Kobzol:
This will cause the collector to run forever and wait for new jobs.

This PR also adds a bash script to run the collector with the new job queue.

### DIFF
--- a/collector/collect-job-queue.sh
+++ b/collector/collect-job-queue.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# This script expects DATABASE and COLLECTOR_NAME to be defined in the environment
+
+set -u -o pipefail
+
+echo "Running job queue collector"
+
+export RUST_LOG=collector=trace,collector::sysroot=debug
+export PATH="/home/collector/.cargo/bin:$PATH"
+
+while : ; do
+        # Update and rebuild the collector.
+        git pull
+        git reset --hard @{upstream}
+
+        # Make sure we have a recent build, so that we can successfully build
+        # the collector.
+        rustup update stable
+        cargo +stable build --release -p collector
+
+        CURRENT_SHA=`git rev-parse HEAD`
+
+        target/release/collector benchmark_job_queue \
+          --db "${DATABASE}" \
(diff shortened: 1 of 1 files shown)


collector/src/bin/collector.rs
```diff
@@ -1504,96 +1507,113 @@ async fn run_job_queue_benchmarks(
...
-        // Here we check if we should update our commit SHA, if rustc-perf has been updated.
-        // We only check for updates when we switch *benchmark requests*, not *benchmark jobs*,
-        // to avoid changing code in the middle of benchmarking the same request.
-        // Note that if an update happens, the job that we have just dequeued will have its deque
-        // counter increased. But since updates are relatively rare, that shouldn't be a big deal,
-        // it will be dequeued again when the collector starts again.
-        if check_git_sha && is_new_request && needs_git_update(collector) {
+    // Outer loop - wait for some jobs to appear
+    loop {
+        if check_git_sha && needs_git_update(collector) {
```
Jamesbarford, line 1512: Will this smash `get_latest_sha`, which in turn makes a network request thus possibly hitting rate limits?

Kobzol: In case there are no jobs to be found, it would send a network request to GitHub every 30 seconds, yes. I wouldn't exactly call that smashing though. If you think that 30s is too much, I can add some guard to do this check only e.g. every few minutes or so. It will be done before starting a job anyway.

Jamesbarford: Ignore me, it's not a network request. It's a command line call

Kobzol: (Which does a network request :sweat_smile:) We have to make a network request to find if git has been updated.

Jamesbarford: Ahhhhh :shaking_face:

Jamesbarford: 30 seconds seems fine, I was thinking that `loop` would spin with no sleep. But that is me misreading it

PR #2295, opened 2025-10-20
Fix stable benchmarks CI

Kobzol:
Revert back to 22.04, due to some issues with `perf` being unavailable for newer kernels.


Jamesbarford (approved):
LGTM :+1:

PR #2296, opened 2025-10-21
Ensure that deploys are not concurrent

Kobzol:
We have a merge queue now, but we still deploy from master - make sure that we always deploy at most one version at a time.


Jamesbarford (approved):
LGTM :+1:

PR #2298, opened 2025-10-22
Add multi-collector documentation

Jamesbarford:
It is effectively a consolidation of the hack.md's

### DIFF
--- a/docs/multiple-collectors.md
+++ b/docs/multiple-collectors.md
@@ -34,6 +34,8 @@ The table below details a set of keywords, or a glossary of terms, that appear t
 | **MAX_JOB_FAILS** | Maximum number of failures before a job is marked as a failed. |
 | **Assigning a job** | The act of allocating one or more *jobs* to a collector. |
 | **website** | A standalone server responsible for inserting work into the queue. |
+| **backfilling** | Occurs when a commit's parent_sha does not have the same configuration as the request currently being enqueued. In this case, jobs with the requested configuration are added so that the commit can be benchmarked against its parent under matching conditions. |
+| **benchmark index** | A set off shas and release tags which have completed benchmark requests. Saves database lookups. |
 
 ## Programs that need to be available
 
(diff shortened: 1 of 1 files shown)


docs/multiple-collectors.md
```diff
@@ -0,0 +1,166 @@
...
+| **collector_id** | A unique identifier of a *collector* (hard-coded at first for simplicity). |
+| **job** | High-level "work item" that defines a set of *test cases* that should be benchmarked on a specific collector. |
+| **job_queue** | Queue of *jobs*. |
+| **MAX_JOB_FAILS** | Maximum number of failures before a job is marked as a failed. |
+| **Assigning a job** | The act of allocating one or more *jobs* to a collector. |
+| **website** | A standalone server responsible for inserting work into the queue. |
+
+### Master artifacts
+
+The website will go through all recent master commits, and check if they are done by looking at the `sha` and `status` column in the `benchmark_request` table.
```
Kobzol: This is not really what happens right now, because we just load the known SHAs into the benchmark request index.

docs/multiple-collectors.md
```diff
@@ -0,0 +1,166 @@
...
+- If the request is `waiting for parent` commit benchmark to be completed, nothing happens.
+- If the request is missing, we will recursively find a set of parent master commits that are missing data (by looking at their status in `benchmark_request`).
+    - If the set is non-empty, these commits will be handled recursively with the same logic as this commit.
+    - If the set is empty, the request will be *enqueued*.
+
+### Try artifacts
+
+> The logic for try artifacts can either happen both in cron and in the GH webhook listener (that receives `@rust-timer queue/build` notifications), or only in cron.
+
+The website will go through all try artifacts in `benchmark_request` that are not yet marked as `completed`.
```
Kobzol: In general, the text here is very pseudo-codey and doesn't correspond to how rustc-perf operates. It would be better to describe what the cron job does more literaly, I think.

docs/multiple-collectors.md
Jamesbarford: I've corrected this now :+1:

docs/multiple-collectors.md
Jamesbarford: I feel I've done both, as attempt to explain the CRON more literally. I think this is quite useful, I've moved it to after the explanation of the CRON which feels like it flows a bit better now

docs/multiple-collectors.md
```diff
@@ -115,6 +89,29 @@ Step 4:
...
+    - If the set is non-empty, these commits will be handled recursively with the same logic as this commit.
+    - If the set is empty, the request will be *enqueued*.
+
+### Try artifacts
+
+The website will go through all try artifacts in `benchmark_request` that are not yet marked as `completed`.
+
+- If the request is `waiting for artifacts`, do nothing (sometime later a GH notification will switch the status to `waiting for parent` once the artifacts are ready).
+- If the request is `waiting for parent`:
+    - Recursively find a set of **grandparent** master commits that are missing data (by looking at their status in `benchmark_request`). This could happen on the edge switch from `waiting for artifacts` to `waiting for parent` in the GH webhook handler, or it could happen in each cron invocation.
```
Kobzol: We don't do this recursive lookup.

docs/multiple-collectors.md
```diff
@@ -115,6 +89,29 @@ Step 4:
...
+    - If that set is empty, generate all necessary **parent** jobs and check if they are all completed in the `job_queue`.
```
Kobzol: We don't check the parent jobs.

docs/multiple-collectors.md
```diff
@@ -115,6 +89,29 @@ Step 4:
...
 The collectors health is monitored by updating a heartbeat column in the `collector_config` table. The UI will indicate the collector as offline if it is inactive for a specified period of time. This should be caught either by error logs or someone viewing the page and subsequently reporting the collector as offline in Zulip.
 
+### Master artifacts
+
+The website maintains a set of all master commits that have been completed in memory so it is able to quickly determine if the commit needs benchmarking.
+
+- If the request's sha is in the benchmark index, nothing happens.
+- If the request is `in_progress`, check [request completion](#Checking-request-completion).
+- If the request is `waiting_for_parent` commit benchmark to be completed, nothing happens.
+- If the request is missing, we will recursively find a set of parent master commits that are missing data (by looking at their status in `benchmark_request`).
```
Kobzol: Same here, we don't recursively check for missing parents.

Kobzol (commented):
I think that it might be better to describe the functionality around what happens in the code in `master` currently. The HackMD still describes conceptual ideas (like recursive lookup of parents and checking if a job is in a job queue or not), which don't actually happen in the code, so it would be confusing to describe it as such in the document.

So something like:
- When a cron job runs, it scans master commits and releases from the web, if they are not in the benchmark request index, it adds them to the DB.
- Then it tries to finish in-progress request(s).
- Then it build the request queue, and tries to enqueue the first request, if there is nothing in progress.
- During enqueuing, we insert jobs both for the request and its parent; if the parent jobs already existed, nothing new will be inserted.

docs/multiple-collectors.md
Jamesbarford: I've removed these two blocks as they don't really add anything, I've added to the key terms and improved the CRON write up; bda5b90b7f6a8b476a5437d6b2a63a7a2ee37a07

PR #2301, opened 2025-10-25
Make rustc-perf example Cargo commands more bulletproof

Kobzol:
Hopefully this should resolve issues like https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/collector.20can't.20rustc.20--print.20sysroot/with/547044191 for good.

### DIFF
--- a/site/frontend/src/utils/cargo.ts
+++ b/site/frontend/src/utils/cargo.ts
@@ -1,0 +1,9 @@
+/**
+ * Return a Cargo command that builds rustc-perf properly and then executes
+ * the collector.
+ * The returned command is meant to be followed by a `collector` command and its
+ * arguments.
+ */
+export function cargo_collector_command(): string {
+  return "cargo build --release -p collector && ./target/release/collector";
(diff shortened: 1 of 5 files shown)


site/frontend/src/utils/cargo.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Return a Cargo command that builds rustc-perf properly and then executes
+ * the collector.
+ * The returned command is meant to be followed by a `collector` command and its
+ * arguments.
+ */
+export function cargo_collector_command(): string {
+  return "cargo build --release -p collector && ./target/release/collector";
```
lqd, line 8: So this builds rustc-fake while --bin collector doesn’t?!

Kobzol: Exactly. Cargo has no bin dependencies yet, so there's no way to tell that `--bin collector` actually depends on `--bin rustc-fake`.

PR #2306, opened 2025-11-03
Show collector jobs by default

Kobzol:
I just ended up clicking on "Show jobs" after every page load... :D


Jamesbarford (approved):
Makes sense to me, when I was checking in on the page I was also clicking that button. And there aren't so many jobs on a collector that it becomes visual noise. :+1:

PR #2309, opened 2025-11-03
Use the new job queue for request duration estimation

Kobzol:
Currently, all GH PR comments show that there is nothing in the queue. This PR switches that estimation to the new job queue.


Kobzol:
This is called once per queued try build, which happens maybe 5-10 times per day, so I don't think that's a problem.

PR #2319, opened 2025-11-05
Show at the top of the timeline when the current benchmark is expected to finish

Jamesbarford:
Display a message saying what is currently in progress and when it is expected to finish. Or nothing if nothing is in progress.

image

### DIFF
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -124,15 +124,30 @@ function ExpectedCurrentRequestCompletion() {
   const diffSeconds = differenceInSeconds(estimatedCompleted, now);
   const prettyDisplay = formatSecondsAsDuration(diffSeconds);
 
-  return (
-    <span>
-      Current Benchmark for{" "}
-      <strong>
-        <CommitSha tag={req.tag}></CommitSha>
-      </strong>{" "}
-      expected to end in approximately {prettyDisplay}
-    </span>
-  );
+  if (req.requestType === "Release") {
+    return (
+      <span>
+        Current Benchmark for{" "}
+        <strong>
+          <CommitSha tag={req.tag}></CommitSha>
+        </strong>{" "}
+        expected to end in approximately {prettyDisplay}
+      </span>
+    );
+  } else {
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/status/page.vue
```diff
@@ -104,6 +105,23 @@ function getErrorsLength(errors: Dict<string>) {
...
+  if (!req.endEstimated) return "";
+  if (!req.completedAt) return "";
+  const formattedDate = formatISODate(req.completedAt);
+  return (
+    <span>
+      Current Benchmark for{" "}
+      <strong>
+        <CommitSha tag={req.tag}></CommitSha>
+      </strong>{" "}
+      expected to end at {formattedDate}{" "}
```
Kobzol: We already show the absolute duration in the timeline table, I thought that the added value of this would be to show relative time ("is expected to end in 5 minutes"). We already have code for this in the old status page, so I'd just copy-paste it from there.

site/frontend/src/pages/status/commit-sha.vue
```diff
@@ -1,19 +1,21 @@
...
+const looksLikeSha = computed(() => tagLooksLikeSha(props.tag));
 </script>
 
 <template>
   <a
     v-if="looksLikeSha"
     :href="'https://github.com/rust-lang/rust/commit/' + tag"
   >
-    {{ tag.substring(0, 13) }}
+    {{ truncate ? tag.substring(0, 13) : tag }}
```
Kobzol, line 19: I'd always truncate the tag if it looks like a SHA, otherwise the message will be annoyingly wide. https://github.com/rust-lang/rustc-perf/pull/2318 solves showing the full SHA (but people just click on the link anyway and then copy-paste from the URL, I think).

Jamesbarford: It's not that massive and it avoids _any_ ambiguity which felt like some of the feedback was aimed at; 
image

site/frontend/src/pages/status/page.vue
Jamesbarford: a29b1ec6eee7655aab2d003dbfa6245e7ddc2c2b

Kobzol (approved):

PR #2320, opened 2025-11-05
Correctly pass backend when computing benchmark detail graph

Kobzol:
I tried to do a backfill of Cranelift results, which stored some Cranelift data in the production database. This is now starting to uncover some weak spots in the querying logic where we were not selecting the right backend correctly.

This PR fixes the compile-time detailed graph endpoint.

Found [here](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/30.20day.20history.20graph.20endpoint.20.20returns.20502/with/553860741).

### DIFF
--- a/database/src/selector.rs
+++ b/database/src/selector.rs
@@ -212,6 +212,11 @@ impl CompileBenchmarkQuery {
         self
     }
 
+    pub fn backend(mut self, selector: Selector<CodegenBackend>) -> Self {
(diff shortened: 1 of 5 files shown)


lqd:
The "fuck it we'll do it live" way. This is the way.

lqd (approved):

PR #2321, opened 2025-11-10
Fix benchmarking of stable/beta releases

Kobzol:
Those should only run the stable benchmarks and should not run the bootstrap benchmark.

Noted in https://perf.rust-lang.org/status.html (the beta release had errors and ran for too long).


Jamesbarford (approved):
LGTM :+1:

PR #2325, opened 2025-11-11
Add stable benchmarks to the compile-time benchmark set

Kobzol:
https://github.com/rust-lang/rustc-perf/pull/2321 wasn't enough, the compile-time benchmark set has to contain *all* benchmarks, otherwise release benchmarks won't work.


Jamesbarford (commented):
LGTM :+1:

PR #2326, opened 2025-11-17
Fix - return an error if `@rust-timer` is given no arguments

Jamesbarford:
Should handle the case where `@rust-timer` has no arguments.

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -158,6 +158,11 @@ async fn handle_rust_timer(
             Err(error) => errors.push_str(&format!("Cannot parse build command: {error}\n")),
         }
     }
+
+    if valid_build_cmds.is_empty() {
+        errors.push_str("@rust-timer command cannot be empty\n");
+    }
+
     if !errors.is_empty() {
         main_client.post_comment(issue.number, errors).await;
         return Ok(github::Response);


site/src/request_handlers/github.rs
```diff
@@ -158,6 +158,11 @@ async fn handle_rust_timer(
             Err(error) => errors.push_str(&format!("Cannot parse build command: {error}\n")),
         }
     }
+
+    if valid_build_cmds.is_empty() {
+        errors.push_str("@rust-timer command cannot be empty\n".to_string());
```
lqd:
```suggestion
        errors.push_str("@rust-timer command cannot be empty\n");
```

r=me when green

site/src/request_handlers/github.rs
```diff
@@ -158,6 +158,11 @@ async fn handle_rust_timer(
             Err(error) => errors.push_str(&format!("Cannot parse build command: {error}\n")),
         }
     }
+
+    if valid_build_cmds.is_empty() {
```
Kobzol, line 162: This should also have `&& errors.is_empty()`, otherwise it would say "command cannot be empty" when it was not empty, but there was some command parser failure.

site/src/request_handlers/github.rs
```diff
@@ -158,6 +158,11 @@ async fn handle_rust_timer(
...
+        errors.push_str("@rust-timer command cannot be empty\n");
```
Kobzol, line 163: Oops, there's a bigger problem with this. Because the bot used `@rust-timer` here, it triggers an endless sequence of bot comments 😆 https://github.com/rust-lang/rust/pull/112049#issuecomment-3542567996

lqd: 😱

PR #2331, opened 2025-11-18
Post more precise message if a Try is queued multiple times

Jamesbarford:
This will give a more precise message if a Try is enqueued multiple times

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -108,7 +108,7 @@ async fn record_try_benchmark_request_without_artifacts(
         Ok(BenchmarkRequestInsertResult::Inserted) => get_awaiting_on_bors_message(),
         Err(e) => {
             log::error!("Failed to insert release benchmark request: {e}");
-            "Something went wrong! This is most likely an internal failure, please let us know on [Zulip]( https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra)".to_string()
+            "Something went wrong! This is most likely an internal failure, please let us know on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra)".to_string()
         }
     }
 }


site/src/request_handlers/github.rs
```diff
@@ -76,19 +76,48 @@ async fn handle_issue(
...
-) {
+) -> ServerResult<github::Response> {
     let try_request = BenchmarkRequest::create_try_without_artifacts(pr, backends, "");
     log::info!("Inserting try benchmark request {try_request:?}");
-    if let Err(e) = conn.insert_benchmark_request(&try_request).await {
-        log::error!("Failed to insert try benchmark request: {}", e);
+
+    match conn.insert_benchmark_request(&try_request).await {
+        Ok(BenchmarkRequestInsertResult::AlreadyQueued) => {
+            log::error!(
```
```suggestion
            log::info!(
```
This is not really an error, just the user did an operation that had no effect.

site/src/request_handlers/github.rs
```diff
@@ -76,19 +76,48 @@ async fn handle_issue(
...
+                "Failed to insert try benchmark request, a request for PR`#{pr}` already exists"
+            );
+            client.post_comment(pr, format!("Failed to insert try benchmark request, a request for PR`#{pr}` already exists")).await;
```
```suggestion
            client.post_comment(pr, format!("This pull request was already queued before and is awaiting a try build to finish.")).await;
```
Otherwise it sounds like there was some failure in the bot itself.

site/src/request_handlers/github.rs
```diff
@@ -76,19 +76,48 @@ async fn handle_issue(
...
+        }
+        Ok(BenchmarkRequestInsertResult::Queued) => {
+            client
+                .post_comment(pr, get_awaiting_on_bors_message())
+                .await;
+        }
+        Err(e) => {
+            log::error!("Failed to insert release benchmark request: {e}");
+            client
+                .post_comment(pr, "Something went wrong! This is most likely an internal failure, please message on Zulip".to_string())
```
```suggestion
                .post_comment(pr, "Something went wrong! This is most likely an internal failure, please let us know on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance)".to_string())
```

site/src/request_handlers/github.rs
```diff
@@ -76,19 +76,48 @@ async fn handle_issue(
...
+
+{COMMENT_MARK_TEMPORARY}"
+    )
+}
+
 /// The try request does not have a `sha` or a `parent_sha` but we need to keep a record
 /// of this commit existing. The DB ensures that there is only one non-completed
 /// try benchmark request per `pr`.
 async fn record_try_benchmark_request_without_artifacts(
+    client: &client::Client,
```
Kobzol: Seems like this function could just return a String with the comment contents that should be posted? So that it doesn't have to know about the GH client and the HTTP response.

site/src/request_handlers/github.rs
```diff
@@ -123,7 +152,8 @@ async fn handle_rust_timer(
                 let conn = ctxt.conn().await;
 
                 if should_use_job_queue(issue.number) {
-                    record_try_benchmark_request_without_artifacts(
+                    return record_try_benchmark_request_without_artifacts(
```
Kobzol: And then here we don't need to have `return`, and can just use the previous code flow.

site/src/request_handlers/github.rs
```diff
@@ -177,7 +201,8 @@ async fn handle_rust_timer(
         let conn = ctxt.conn().await;
         for command in &valid_build_cmds {
             if should_use_job_queue(issue.number) {
-                record_try_benchmark_request_without_artifacts(
+                return record_try_benchmark_request_without_artifacts(
```
Kobzol: The code in theory handles multiple commands, so it's weird that it uses `return` here. Let's just post all the comments generated by `record_try_benchmark_request_without_artifacts` here and don't return early.

database/src/pool/postgres.rs
```diff
@@ -1648,7 +1649,9 @@ where
                     profiles,
                     commit_date
                 )
-                VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9);
+                VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
+                ON CONFLICT DO NOTHING
+                RETURNING id;
```
Kobzol: Do we need the `RETURNING` part for anything?

database/src/lib.rs
```diff
@@ -1039,6 +1039,12 @@ impl BenchmarkRequest {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum BenchmarkRequestInsertResult {
+    Queued,
```
Kobzol: I understand what do you mean by this, but without context, those two names are a bit.. confusing 😆 I'd either:
- Rename to `Inserted`/`Ok` and `NothingWasInserted`, or
- Add a doc comment that described what do those variants mean

Kobzol (commented):
Thanks! Left some comments.

site/src/request_handlers/github.rs
lqd: Why not on t-infra's stream?

database/src/pool/postgres.rs
Jamesbarford: So we get a row back, otherwise nothing will be returned

site/src/request_handlers/github.rs
Kobzol: I mean, it might as well be my DMs xD But yeah, maybe t-infra is better. https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra

database/src/pool/postgres.rs
Kobzol: Just keep using `execute` as before (instead of `query`), it returns the number of affected rows (https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Client.html#method.execute).

site/src/request_handlers/github.rs
```diff
@@ -201,8 +192,7 @@ async fn handle_rust_timer(
         let conn = ctxt.conn().await;
         for command in &valid_build_cmds {
             if should_use_job_queue(issue.number) {
-                return record_try_benchmark_request_without_artifacts(
-                    main_client,
+                record_try_benchmark_request_without_artifacts(
```
Kobzol, line 196: Ah, I realized that this is a bit tricky. Because here we should actually post the comment returned by `record_try_benchmark_request_without_artifacts`, but *only* if there was either an error or it was already queued before. So `record_try_benchmark_request_without_artifacts` would return e.g. an enum, and then the code would only post a comment from the outside based on its value.

That being said, this situation should be super rare, let's just ignore it here to avoid complicating it further. We also didn't deal with it in the old system.

site/src/request_handlers/github.rs
```diff
@@ -76,18 +76,40 @@ async fn handle_issue(
...
+            log::info!(
+                "Failed to insert try benchmark request, a request for PR`#{pr}` already exists"
+            );
+            "This pull request was already queued before and is awaiting a try build to finish."
+                .to_string()
+        }
+        Ok(BenchmarkRequestInsertResult::Inserted) => get_awaiting_on_bors_message(),
+        Err(e) => {
+            log::error!("Failed to insert release benchmark request: {e}");
+            "Something went wrong! This is most likely an internal failure, please let us know on [Zulip]( https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra)".to_string()
```
```suggestion
            "Something went wrong! This is most likely an internal failure, please let us know on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra)".to_string()
```

Kobzol (approved):
Looks good, thank you!

PR #2338, opened 2025-12-05
Improve queue ordering when bootstrapping the system

Kobzol:
If the new system boots up with an empty DB (in local experiments), it should at least try to order the master requests semi-reasonably. This change improves that.


Jamesbarford (approved):
LGTM :+1:

PR #2339, opened 2025-12-05
Modify UI of the status page

Kobzol:
Some small improvements to the status page, based on papercuts I found while testing multiple collectors. Most of the commits compress the size of the job table, or add more information to it.

### DIFF
--- a/site/frontend/src/pages/status/collector.vue
+++ b/site/frontend/src/pages/status/collector.vue
@@ -136,6 +136,16 @@ function formatTime(dateString: string | null): string {
   }
   return format(date, "HH:mm:ss");
 }
+
+function jobDuration(job: BenchmarkJob): string {
+  if (!isJobComplete(job)) {
+    return "";
+  }
+  const start = parseDateIsoStringOrNull(job.startedAt);
+  const end = parseDateIsoStringOrNull(job.completedAt);
+  const diff = differenceInSeconds(end, start);
+  return `Job took ${formatSecondsAsDuration(diff)}`;
+}
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/status/collector.vue
```diff
@@ -101,6 +117,35 @@ function formatProfile(job: BenchmarkJob): string {
     return "";
   }
 }
+function timeSince(timestamp: string): string {
```
Jamesbarford, line 120: We have a `site/frontend/src/utils/formatting.ts` file, I think this should live there

site/frontend/src/pages/status/collector.vue
```diff
@@ -101,6 +117,35 @@ function formatProfile(job: BenchmarkJob): string {
...
+    return "";
+  }
+  const now = new Date();
+  const diffSeconds = differenceInSeconds(now, date);
+  return formatSecondsAsDuration(diffSeconds);
+}
+
+// Takes a date like `2025-09-10T08:22:47.161348Z` and shows just the time
+// portion (`08:22:47`).
+function formatTime(dateString: string | null): string {
```
Jamesbarford, line 132: Think this should also live in; `site/frontend/src/utils/formatting.ts`

site/frontend/src/pages/status/collector.vue
Kobzol: I usually only put functions into a shared module if more than a single file needs them. Those two functions (especially `formatTime`) are quite specific for this file, and I don't think that they belong to `formatting.ts` without being generalized somehow.

Jamesbarford: Makes sense :+1:

PR #2342, opened 2025-12-08
Allow for optional jobs not interfering which benchmark completion

Jamesbarford:
EDIT: Original comment cleared up below.

I see a problem with this;

If a benchmark request (req1) is marked as `complete` while it still has unfinished optional jobs, and a subsequent benchmark request (req2) is marked as `in_progress`, then any optional jobs belonging to req1 will be skipped.

If this were implemented using a generic `is_optional` flag, those jobs would also be skipped, because the request would be considered `complete` before the optional jobs were ever picked up.

The way I see to rectify this is to change how the dequeue query logic works for the ordering. Moreover `is_optional` is perhaps misleading as a term though I'm struggling to think of a better name. However I'd be interested to know your thoughts before doing so.

---

For the ui should we just filter optional jobs out?

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1990,8 +1990,7 @@ where
                         AND benchmark_set = $3
                     ORDER BY
                         -- Prefer in-progress jobs that have not been finished
-                        -- previously, so that we can finish them. Also prefer
-                        -- mandatory jobs over optional ones.
+                        -- previously, so that we can finish them.
                         CASE
                             WHEN status = $5 THEN 0
                             WHEN status = $1 THEN 1


database/src/pool/postgres.rs
```diff
@@ -1985,11 +1989,12 @@ where
...
                     ORDER BY
-                        -- Prefer in-progress jobs that have not been finished previously, so that
-                        -- we can finish them.
+                        -- Prefer in-progress jobs that have not been finished
+                        -- previously, so that we can finish them. Also prefer
+                        -- mandatory jobs over optional ones.
                         CASE
-                            WHEN status = $5 THEN 0
-                            WHEN status = $1 THEN 1
+                            WHEN (status = $5 AND is_optional = FALSE) THEN 0
```
Kobzol: I don't think that we have to prefer non-optional jobs. It shouldn't even be an issue in practice, as we expect to use optional jobs for different targets/collectors, so a single collector should ideally only ever do required or only ever do optional jobs, and not mix-and-match.

database/src/tests/builder.rs
```diff
@@ -116,13 +117,19 @@ pub struct JobBuilder {
...
+    is_optional: bool,
 }
 
 impl JobBuilder {
     pub fn profile(mut self, profile: Profile) -> Self {
         self.profile = profile;
         self
     }
+
+    pub fn is_optional(mut self, is_optional: bool) -> Self {
```
Kobzol:
```suggestion
    pub fn make_optional(mut self) -> Self {
```
Otherwise the name was confusing, it was a setter that begins with `is_`.

database/src/pool.rs
```diff
@@ -1286,4 +1292,98 @@ mod tests {
...
+            let (job, _) = db
+                .dequeue_benchmark_job(collector_name, target, benchmark_set)
+                .await
+                .unwrap()
+                .unwrap();
+
+            assert_eq!(job.request_tag(), benchmark_request.tag().unwrap());
+
+            /* Make the job take some amount of time */
+            std::thread::sleep(Duration::from_millis(1000));
```
```suggestion
            tokio::time::sleep(Duration::from_millis(1000)).await;
```
Although I'm not sure if we even need to wait?

Kobzol (commented):
Thanks, looks fine. Left some comments.

database/src/tests/builder.rs
Jamesbarford: :+1: 35a3aebd028e738a01ad6e267d7af1ced556811d

database/src/pool.rs
Jamesbarford: :+1: 35a3aebd028e738a01ad6e267d7af1ced556811d (we need the `await`)

database/src/pool/postgres.rs
Jamesbarford: :+1: 35a3aebd028e738a01ad6e267d7af1ced556811d, yup that makes sense

database/src/pool/postgres.rs
```diff
@@ -1985,8 +1989,9 @@ where
                         AND target = $2
                         AND benchmark_set = $3
                     ORDER BY
-                        -- Prefer in-progress jobs that have not been finished previously, so that
-                        -- we can finish them.
+                        -- Prefer in-progress jobs that have not been finished
+                        -- previously, so that we can finish them. Also prefer
+                        -- mandatory jobs over optional ones.
```
```suggestion
                        -- previously, so that we can finish them.
```

Jamesbarford: 2cd7234e783e7d62570d5dea80e2bd251eb27d17 :+1:

PR #2351, opened 2025-12-09
Update job queue documentation

Kobzol:
This PR unifies our documentation a little bit, and rewrites the job queue documentation to reflect the latest state, and remove duplicated things that were already documented in the glossary and the DB schema.

I'd like to request a review also from someone who hasn't interacted with the new system at all (e.g. @marcoieni or @Mark-Simulacrum), to see if it's comprehensible to other people (mainly the `job-queue.md` document).


docs/job-queue.md
```diff
@@ -0,0 +1,154 @@
...
+Their information is stored in the `collector_config` table. Each collector has its assigned benchmark set and target, which are used to determine which jobs the collector will handle.
+
+To register a new collector, you can run the following command:
+
+```bash
+cargo run --bin collector add_collector \
+  --collector_name "<name>" \
+  --target <target> \
+  --benchmark_set "<benchmark-set>" \
+  --is_active
```
Mark-Simulacrum, line 119: Presumably this needs to be run with credentials - maybe on the instance itself?

Kobzol: It doesn't need to be on the collector machine, but it needs access to the database where the collector will be stored, yeah.

docs/job-queue.md
```diff
@@ -0,0 +1,154 @@
+# Job queue
```
Jamesbarford, line 1: I think we could do something like this to make it obvious that we are talking about the deployed system;

```md
> [!WARNING]
> The following is for the queue used for the deployed version of rustcperf. For running locally please see <doc-link>
```

Or we could do something like;

```
docs/
├── README.md            (links to things)
├── local/
│   ├── README.md
│   └── running_locally.md
└── deployed/
    ├── README.md
    └── job_queue.md
```

What do you think?

Kobzol: So, I have been thinking about the same thing. But it's not just the `docs` directory, we have docs in the `site` directory, in the `collector` directory, in the `database` directory, it's a bit all over the place. I don't think it makes sense to do incremental changes to the *docs structure* itself, if we wanted to make it more comprehensive, I'd just go make it a normal mdbook or something (but I think that is still overkill at this point).

I wouldn't necessarily go for a full hierarchy here. A small improvements that we could do is to have an "index" of all documentation in `docs/README.md`, which will link to the existing docs. And in fact I already did that in this PR.

Jamesbarford: Sounds good to me

PR #2352, opened 2025-12-11
Add `target` for GitHub commands

Jamesbarford:
Creates `targets=<...>` functionality, defaulting to `x86_64-unknown-linux-gnu` in the database.

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -441,7 +441,7 @@ static MIGRATIONS: &[&str] = &[
     ALTER TABLE runtime_pstat_series ADD CONSTRAINT runtime_test_case UNIQUE(benchmark, target, metric);
     "#,
     r#"ALTER TABLE job_queue ADD COLUMN is_optional BOOLEAN NOT NULL DEFAULT FALSE"#,
-    r#"ALTER TABLE benchmark_request ADD COLUMN targets TEXT NOT NULL DEFAULT 'x86_64-unknown-linux-gnu'"#,
+    r#"ALTER TABLE benchmark_request ADD COLUMN targets TEXT NOT NULL DEFAULT ''"#,
 ];
 
 #[async_trait::async_trait]


database/src/lib.rs
```diff
@@ -913,6 +921,7 @@ pub struct BenchmarkRequest {
     status: BenchmarkRequestStatus,
     backends: String,
     profiles: String,
+    targets: Option<String>,
```
Kobzol: So, I wanted to actually refactor this so that we can distinguish between "default values" for the parameter vs "overridden values" from the outside. But I think that with the `Target::is_optional` method, we don't need that at the moment.

Therefore, any reason why is this not `String`? It would be nice to make it consistent with how backends and profiles are handled, as it's pretty much the same thing.

database/src/lib.rs
```diff
@@ -1056,18 +1085,26 @@ pub enum BenchmarkRequestInsertResult {
     NothingInserted,
 }
 
-pub fn parse_backends(backends: &str) -> Result<Vec<CodegenBackend>, String> {
-    backends
+fn parse_raw_string<T>(raw_string: &str, name: &str) -> Result<Vec<T>, String>
```
```suggestion
fn parse_comma_separated<T>(raw_string: &str, name: &str) -> Result<Vec<T>, String>
```

database/schema.md
```diff
@@ -276,6 +276,7 @@ Columns:
   * `completed`: Completed request.
 * **backends** (`text NOT NULL`): Comma-separated list of codegen backends to benchmark. If empty, the default set of codegen backends will be benchmarked.
 * **profiles** (`text NOT NULL`): Comma-separated list of profiles to benchmark. If empty, the default set of profiles will be benchmarked.
+* **targets** (`text NOT NULL`): Comma-separated list of targets to benchmark. If empty, the default `x86_64-unknown-linux-gnu` be benchmarked.
```
Kobzol:
```suggestion
* **targets** (`text NOT NULL`): Comma-separated list of targets to benchmark. If empty, the default `x86_64-unknown-linux-gnu` will be benchmarked.
```

database/src/pool/postgres.rs
```diff
@@ -441,6 +441,7 @@ static MIGRATIONS: &[&str] = &[
     ALTER TABLE runtime_pstat_series ADD CONSTRAINT runtime_test_case UNIQUE(benchmark, target, metric);
     "#,
     r#"ALTER TABLE job_queue ADD COLUMN is_optional BOOLEAN NOT NULL DEFAULT FALSE"#,
+    r#"ALTER TABLE benchmark_request ADD COLUMN targets TEXT NOT NULL DEFAULT 'x86_64-unknown-linux-gnu'"#,
```
```suggestion
    r#"ALTER TABLE benchmark_request ADD COLUMN targets TEXT NOT NULL DEFAULT ''"#,
```
Let's make the default empty string, because that is also what we use as the default when creating new requests.

Kobzol (approved):
Thanks! Let's try it.

PR #2354, opened 2025-12-16
Toggle status page date format between 24hr and 12hr

Jamesbarford:
A little checkbox button allowing changing the date format between 24hr and 12hr;

## 24hr

24hr format

## 12hr

12hr format

Closes https://github.com/rust-lang/rustc-perf/issues/2347

### DIFF
--- a/site/frontend/src/pages/status/page.vue
+++ b/site/frontend/src/pages/status/page.vue
@@ -9,10 +9,9 @@ import {
   formatISODate,
   formatSecondsAsDuration,
   parseDateIsoStringOrNull,
-  setDateFmt,
-  getDateFmt,
   DATE_FMT_12HR,
   DATE_FMT_24HR,
+  preferredDateTimeFormat,
 } from "../../utils/formatting";
 import {useExpandedStore} from "../../utils/expansion";
 import {
@@ -33,11 +32,6 @@ const data: Ref<{
   queueLength: number;
   collectors: CollectorConfig[];
 } | null> = ref(null);
-
-// This state exists so the UI immediately update when a user changes their
-// date format preference. As only setting Local Storage does not trigger a
-// re-render.
-const dateFmt: Ref<string> = ref(getDateFmt());
 
 type BenchmarkRequestRow = BenchmarkRequest & {
   isLastInProgress: boolean;
@@ -216,13 +210,12 @@ const tableWidth = 8;
 
 function toggleDate() {
   let nextDateFmt: string;
-  if (dateFmt.value === DATE_FMT_24HR) {
+  if (preferredDateTimeFormat.value === DATE_FMT_24HR) {
     nextDateFmt = DATE_FMT_12HR;
   } else {
     nextDateFmt = DATE_FMT_24HR;
   }
-  setDateFmt(nextDateFmt);
-  dateFmt.value = nextDateFmt;
+  preferredDateTimeFormat.value = nextDateFmt;
 }
 
 loadStatusData(loading);
@@ -235,7 +228,10 @@ loadStatusData(loading);
         <h1>Timeline</h1>
         <span class="local-time-message small-padding-bottom">
           <strong>Times are local.</strong>
-          <DateFmtPicker :toggleDate="toggleDate" :dateFmt="dateFmt" />
+          <DateFmtPicker
+            :toggleDate="toggleDate"
+            :dateFmt="preferredDateTimeFormat"
+          />
         </span>
         <span class="small-padding-bottom">
           <ExpectedCurrentRequestCompletion />
--- a/site/frontend/src/utils/formatting.ts
+++ b/site/frontend/src/utils/formatting.ts
@@ -1,4 +1,6 @@
 import {parseISO, format} from "date-fns";
+import {createPersistedRef} from "../storage";
+import {PREF_DATETIME_FORMAT} from "../pages/compare/prefs";
 
 // `time` has to be in seconds
 export function formatSecondsAsDuration(time: number): string {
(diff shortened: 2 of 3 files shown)


site/frontend/src/utils/formatting.ts
```diff
@@ -20,10 +20,23 @@ export function formatSecondsAsDuration(time: number): string {
...
+
+export function setDateFmt(dateFmt: string) {
+  window.localStorage.setItem(DATE_FMT_KEY, dateFmt);
+}
+
+export function getDateFmt() {
+  return window.localStorage.getItem(DATE_FMT_KEY) ?? DATE_FMT_24HR;
+}
+
 // Takes a date like `2025-09-10T08:22:47.161348Z` -> `"2025-09-10 08:22:47"`
```
Kobzol:
```suggestion
// Takes a date like `2025-09-10T08:22:47.161348Z` and formats it according to the user preference stored in local storage (either 12 hour or 24 hour format).
```

site/frontend/src/utils/formatting.ts
```diff
@@ -20,10 +20,23 @@ export function formatSecondsAsDuration(time: number): string {
   return s;
 }
 
+export const DATE_FMT_KEY = "__rustc-perf-user-date-fmt-preference__";
+export const DATE_FMT_24HR = "yyyy-MM-dd HH:mm:ss";
+export const DATE_FMT_12HR = "yyyy-MM-dd pp";
```
Kobzol:
```suggestion
export const DATE_FMT_12HR = "yyyy-MM-dd hh:mm:ss a";
```
Let's make it explicit, I don't know what "localized" could mean. Also with `pp` it wasn't zero-padded in your screenshot, which made the width uneven.

site/frontend/src/utils/formatting.ts
```diff
@@ -20,10 +20,23 @@ export function formatSecondsAsDuration(time: number): string {
   return s;
 }
 
+export const DATE_FMT_KEY = "__rustc-perf-user-date-fmt-preference__";
+export const DATE_FMT_24HR = "yyyy-MM-dd HH:mm:ss";
```
Kobzol, line 26:
```suggestion
// Date formats taken from https://date-fns.org/v4.1.0/docs/format
export const DATE_FMT_24HR = "yyyy-MM-dd HH:mm:ss";
```

site/frontend/src/utils/formatting.ts
```diff
@@ -20,10 +20,23 @@ export function formatSecondsAsDuration(time: number): string {
...
+  window.localStorage.setItem(DATE_FMT_KEY, dateFmt);
+}
+
+export function getDateFmt() {
+  return window.localStorage.getItem(DATE_FMT_KEY) ?? DATE_FMT_24HR;
+}
+
 // Takes a date like `2025-09-10T08:22:47.161348Z` -> `"2025-09-10 08:22:47"`
-export function formatISODate(dateString?: string): string {
+export function formatISODate(dateString?: string, fmt?: string): string {
```
Kobzol:
```suggestion
export function formatISODate(dateString?: string): string {
```
Let's just always load the state from local storage.

site/frontend/src/pages/status/page.vue
```diff
@@ -20,13 +24,15 @@ import {
...
 import CommitSha from "./commit-sha.vue";
+import DateFmtPicker from "./date-format-selection.vue";
 
 const loading = ref(true);
 
 const data: Ref<{
   timeline: BenchmarkRequestRow[];
   queueLength: number;
   collectors: CollectorConfig[];
+  dateFmt: string;
```
Kobzol: The `dateFmt` doesn't belong here, this is data loaded from the server. But we don't need to remember it and duplicate state, it's enough to just load the data from localstorage always (see my other comment).

Kobzol (commented):
Thanks! Left some comments.

site/frontend/src/pages/status/page.vue
Jamesbarford: Gotcha.

I think we need it in state to trick VUE.js into re-rendering the screen. Though I could be wrong, if we do need this wizardy I'll use another `Ref<...>` type as like you say I'm mixing some ui thing with something from the server.

site/frontend/src/utils/formatting.ts
Jamesbarford: 98d672d9dce4282db5374c2061e3501a46b4b716 :+1:

site/frontend/src/utils/formatting.ts
Jamesbarford: 98d672d9dce4282db5374c2061e3501a46b4b716 :+1:

site/frontend/src/utils/formatting.ts
Jamesbarford: 98d672d9dce4282db5374c2061e3501a46b4b716 :+1:

site/frontend/src/utils/formatting.ts
Jamesbarford: 98d672d9dce4282db5374c2061e3501a46b4b716 :+1:

site/frontend/src/pages/status/page.vue
Jamesbarford: c9432107f87f83b54b4cc05a6a76fc04e0fbca7c I tried and it didn't work. So I split it out with a comment as to why

Kobzol: Hmm, it worked for me when I tried it locally before. The component should re-render when the event is triggered, I think. I'll check.

PR #2356, opened 2025-12-18
ARM hybrid CPU support

nnethercote:
I have an ASUS GX10 Ascent which has an ARM CPU with 10 performance cores and 10 efficiency cores. This PR is enough to get rustc-perf working on it.

### DIFF
--- a/collector/src/compile/benchmark/target.rs
+++ b/collector/src/compile/benchmark/target.rs
@@ -5,6 +5,7 @@ use std::{fmt, str::FromStr};
 /// https://doc.rust-lang.org/nightly/rustc/platform-support.html
 ///
 /// Presently we only support x86_64
+/// FIXME: we actually support Windows and aarch64, but that isn't captured here.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, serde::Deserialize)]
 pub enum Target {
     /// `x86_64-unknown-linux-gnu`
--- a/collector/src/compile/execute/mod.rs
+++ b/collector/src/compile/execute/mod.rs
@@ -135,81 +135,136 @@ pub struct CargoProcess<'a> {
     pub workspace_package: Option<String>,
 }
 
-/// Returns an optional list of P-cores, if the system has P-cores and E-cores.
-/// This list *should* be in a format suitable for the `taskset` command.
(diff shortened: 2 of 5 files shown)


nnethercote:
Note: I haven't touched the `Target` enum, which currently has a single variant `X86_64UnknownLinuxGnu`. Presumably this should be added to, but I wasn't sure because that type doesn't seem to be meaningfully used at the moment?

collector/src/compile/execute/mod.rs
```diff
@@ -654,24 +729,44 @@ fn process_stat_output(
...
+        // Map P-core event name to normal event names.
+        cfg_if::cfg_if! {
+            if #[cfg(all(target_os = "linux", target_arch = "x86_64"))] {
+                if name == "cpu_core/instructions:u/" {
+                    name = "instructions:u".to_string();
+                }
+            } else if #[cfg(all(target_os = "linux", target_arch = "aarch64"))] {
+                // ARM P-core events have names like `armv8_pmuv3_0/instructions:u/` and
+                // `armv8_pmuv3_1/branche-misses/`.
+                let re = regex::Regex::new(r"armv[0-9]_pmuv[0-9]_[0-9]/([^/]*)/").unwrap();
```
Kobzol: Might be worth caching the regex, to avoid creating it anew for each parsed metric for each performed benchmark etc.?

nnethercote: :+1:

nnethercote:
I moved the regex and updated the comment about the ASUS GX10 to mention the L3 cache.

Kobzol (approved):
Couldn't approve the PR for the same reason 🤦

collector/src/compile/benchmark/target.rs
```diff
@@ -5,6 +5,7 @@ use std::{fmt, str::FromStr};
 /// https://doc.rust-lang.org/nightly/rustc/platform-support.html
 ///
 /// Presently we only support x86_64
+/// FIXME: we actually support Windows and aarch64, but that isn't captured here.
```
Kobzol, line 8: This Target enum is not exactly "what does rustc-perf support benchmarks on", but "what targets we support/run in production".

PR #2360, opened 2026-01-05
Remove most of the old queue system

Kobzol:
The new job queue system has been running in production, even with two parallel machines, for several weeks now, without any major issues. I don't expect that we would want to switch to the old system now; if some issues come up, we will just deal with them in the new system.

This PR thus removes most of the old job queue system from rustc-perf. There are still a few pieces left that are still used even with the new system, I plan to remove those later.

CC @Mark-Simulacrum

### DIFF
--- a/database/schema.md
+++ b/database/schema.md
@@ -82,27 +82,6 @@ Columns:
 * **id** (`integer`): Unique identifier
 * **perf_commit** (`text`): Commit sha / tag
 
-### collector_progress
-
-Keeps track of the collector's start and finish time as well as which step it's currently on.
-
-Columns:
-
-* **aid** (`integer`): Artifact id, references the id in the artifact table
-* **step** (`text`): The step the collector is currently benchmarking
-* **start** (`timestamptz`): When the collector started
-* **end** (`timestamptz`): When the collector finished
-
-### artifact_collection_duration
-
-Records how long benchmarking takes in seconds.
-
-Columns:
-
-* **aid** (`integer`): Artifact id, references the id in the artifact table
-* **date_recorded** (`timestamptz`): When this was recorded
-* **duration** (`integer`): How long the benchmarking took in seconds
-
 ### benchmark
 
 The different types of compile-time benchmarks that are run. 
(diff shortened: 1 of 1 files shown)


database/src/pool.rs
```diff
@@ -170,12 +150,7 @@ pub trait Connection: Send + Sync {
 
     async fn collector_remove_step(&self, aid: ArtifactIdNumber, step: &str);
 
-    async fn in_progress_artifacts(&self) -> Vec<ArtifactId>;
-
-    async fn in_progress_steps(&self, aid: &ArtifactId) -> Vec<Step>;
-
-    async fn last_n_artifact_collections(&self, n: u32) -> Vec<ArtifactCollection>;
-
+    // TODO: the following two functions do not work anymore and should not be used!
```
Mark-Simulacrum, line 153: What blocks removal?

site/src/request_handlers/github.rs
```diff
@@ -383,7 +357,9 @@ struct BuildCommand<'a> {
 
 #[derive(Debug)]
 struct BenchmarkParameters<'a> {
+    #[allow(unused)]
     include: Option<&'a str>,
+    #[allow(unused)]
     exclude: Option<&'a str>,
```
Mark-Simulacrum, line 363: Presumably we stopped supporting these - do we intend to re-add support? Or should this be FIXME + deleted in the future?

Mark-Simulacrum (approved):

site/src/request_handlers/github.rs
Kobzol: I didn't ever see anyone using them, and now that the whole benchmark takes just 40 minutes, and we can select subprofiles to make that even shorter, I don't think it's needed. So yeah, I wanted to remove them in the future, but I also wanted to mostly avoid functional changes in this PR, and just delete stuff. I'll do it in a follow-up.

database/src/pool.rs
Kobzol: The new system still uses those two functions to do some lookups if something exists in the old system. It should be removed, but I wanted to avoid changing the new system in this PR, to keep it almost strictly non-functional, as it's already quite big.

Kobzol:
Ok, let's try if this doesn't break anything. I'll remove the rest in a follow-up.

site/src/request_handlers/github.rs
Kobzol: https://github.com/rust-lang/rustc-perf/pull/2362

database/src/pool.rs
Kobzol: https://github.com/rust-lang/rustc-perf/pull/2363

PR #2362, opened 2026-01-06
Remove unused try benchmark request parameters

Kobzol:
I haven't ever seen anyone use runs, include or exclude on try build perf. runs, and the new system doesn't support them anyway. We now have better ways of selecting a subset of benchmarks I think, with targets, backends and profiles.


lqd:
Maybe do a github search for rust-timer comments? I've used this before, and I think I remember seeing one in the last few months -- but maybe it was just one of the profiles.

It's not used a lot, that's for sure.

PR #2365, opened 2026-01-06
Correctly pass backend and target to self-profile query

Kobzol:
This was causing 502 errors when non-default data was present in one of the examined commits.

Also turns the assertion into a user-visible error, to make this easier to debug.

### DIFF
--- a/site/src/request_handlers/self_profile.rs
+++ b/site/src/request_handlers/self_profile.rs
@@ -461,8 +461,16 @@ pub async fn handle_self_profile(
         .map_err(|e| format!("invalid run name: {e:?}"))?;
     let index = ctxt.index.load();
 
-    let backend: CodegenBackend = body.backend.parse()?;
-    let target: Target = body.target.parse()?;
+    let backend: CodegenBackend = if let Some(backend) = body.backend {
+        backend.parse()?
+    } else {
+        CodegenBackend::Llvm
+    };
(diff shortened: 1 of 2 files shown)


lqd:
Does the link that used to fail work for you with this PR? (Can we even test this locally with the measureme files being on the server/S3?)

site/src/request_handlers/self_profile.rs
```diff
@@ -461,10 +461,15 @@ pub async fn handle_self_profile(
         .map_err(|e| format!("invalid run name: {e:?}"))?;
     let index = ctxt.index.load();
 
+    let backend: CodegenBackend = body.backend.parse()?;
+    let target: Target = body.target.parse()?;
```
lqd: What will happen on older links that don't have the new url parameters, will this bail and return nothing?

Kobzol: This doesn't preserve backwards compatibility. Do you think that's important? The PR comments don't link to detailed results, so the new link will be generated correctly from the compare page.

lqd: I guess we can try and see, or just have the historical defaults correctly filled in, since these new params are now mandatory. We sometimes share addresses so it’s unfortunate if we break anything but no big deal I guess.

Kobzol: Ok, I made them optional :)

PR #2407, opened 2026-02-12
collector: Bump deps

heiher:
This PR bumps the collector dependency to the minimal version required for LoongArch support.

- libc: 0.2.125
- jemallocator: 0.5.4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1532,9 +1532,9 @@ dependencies = [
 
 [[package]]
 name = "jemallocator"
-version = "0.5.0"
+version = "0.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Kobzol:
Why do you need to bump deps of stable benchmarks? We try to touch those as minimally as we can. For Loongaarch it should be enough to run the normal benchmarks, right?

Kobzol:
The problem is that changing these benchmarks might perturb our own x64 dashboard and its data. Since we don't modify the stable benchmarks very often, can't you just keep this as a downstream patch? Same as with some of the other loongaarch changes.

Kobzol:
Hi, is this PR still useful to you? Sorry, it fell through the cracks.

heiher:
Thanks! I'll close this PR for the time being.

PR #2421, opened 2026-03-23
-Zthreads=N support implemented as an additional dimension 'Parallel'

azhogin:
-Zthreads=N support implemented as an additional dimension 'Parallel'.
Default values for parallels: `[1,4]`.

Aggregations sample:
Aggregations

Primary benchmarks sample:
PrimaryBenchs

Secondary benchmarks sample:
SecondaryBenchs

Graphs sample (graphs have "scenario_parallel" names like "full_par1 / full_par4"):
Graphs

Filters with par1/par4 added:
Filters

LocalProfilingCmd

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -36,6 +36,7 @@ use collector::benchmark_set::{get_benchmark_set, BenchmarkSetId, BenchmarkSetMe
 use collector::codegen::{codegen_diff, CodegenType};
 use collector::compile::benchmark::category::Category;
 use collector::compile::benchmark::codegen_backend::CodegenBackend;
+use collector::compile::benchmark::parallel::Parallel;
 use collector::compile::benchmark::profile::Profile;
 use collector::compile::benchmark::scenario::Scenario;
 use collector::compile::benchmark::target::Target;
@@ -116,6 +117,7 @@ struct CompileBenchmarkConfig {
     self_profile_storage: Option<Box<dyn SelfProfileStorage>>,
     bench_rustc: bool,
     targets: Vec<Target>,
+    parallels: Vec<Parallel>,
 }
 
 struct RuntimeBenchmarkConfig {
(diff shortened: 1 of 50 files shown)


nnethercote:
The problem with this is that parallelism is an entirely separate dimension: parallel `Full`, parallel `IncrFull`, parallel `IncrUnchanged`, parallel `IncrPatched` are all possible. And it's not even obvious from the name `ParallelFrontend` which of those scenarios is the one being adapted.

azhogin:
> So I think that 2) is the way forward. It's not completely trivial to add a new axis to the compilation benchmarks, but it's also not super hard, it "just" needs some boilerplate code that propagates the new config option throughout the collector, the database, the website backend, and then also the frontend. We have a bunch of PRs from last year that added the `target` compilation benchmark axis, so it would be mostly a copy-paste of that.

'Parallel' implemented as an additional axis with default values `[1, 4]`. There are screenshots submitted with site pages how it looks with this change. @Kobzol, any comments about it?

Kobzol:
Sorry, I didn't have time for most reviews in the past few months. Starting from next week, it should be better, I promise to take a look then.

collector/src/bin/collector.rs
```diff
@@ -116,6 +117,7 @@ struct CompileBenchmarkConfig {
     self_profile_storage: Option<Box<dyn SelfProfileStorage>>,
     bench_rustc: bool,
     targets: Vec<Target>,
+    parallels: Vec<Parallel>,
```
Kobzol, line 120:
```suggestion
    frontend_threads: Vec<FrontendParallelism>,
```
That should describe more accurately what we are talking about.

collector/src/bin/collector.rs
```diff
@@ -411,6 +419,10 @@ struct CompileTimeOptions {
     /// It should be a path to the `clippy-driver` binary.
     #[arg(long)]
     clippy: Option<PathBuf>,
+
+    /// Parallel frontend options in comma-separated list
+    #[arg(long, value_delimiter = ',', default_value = "1,4")]
+    parallels: Vec<u32>,
```
Kobzol, line 425:
```suggestion
    /// Parallel frontend thread values in comma-separated list
    #[arg(long, value_delimiter = ',', default_value = "1")]
    frontend_threads: Vec<u32>,
```

collector/src/bin/collector.rs
```diff
@@ -2135,6 +2153,11 @@ async fn bench_published_artifact(
     } else {
         Scenario::all_non_incr()
     };
+    let parallels = if collector::version_supports_parallel_frontend(&toolchain.id) {
```
Kobzol, line 2156: I think that for now it would be enough to just use `1` for the stable benchmarks.

collector/src/compile/benchmark/mod.rs
```diff
@@ -297,32 +303,36 @@ impl Benchmark {
...
-                                target,
-                                already_computed,
-                            )
-                        })
-                        .copied()
-                        .collect::<Vec<Scenario>>();
-                    if remaining_scenarios.is_empty() {
-                        continue;
-                    }
+                    for parallel in parallels {
```
Kobzol, line 306:
```suggestion
                    for thread_count in frontend_threads {
```

collector/src/compile/benchmark/parallel.rs
```diff
@@ -0,0 +1,42 @@
+use std::fmt;
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct Parallel(pub u32);
```
Kobzol, line 4:
```suggestion
pub struct FrontendThreads(pub u32);
```
Or `FrontendParallelism`, I think that both works. We should clarify that it is related to the frontend though, `Parallel` is ambigous.

collector/src/compile/benchmark/parallel.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl From<Parallel> for database::Parallel {
+    fn from(value: Parallel) -> Self {
+        database::Parallel(value.0)
+    }
+}
+
+impl fmt::Display for Parallel {
```
Kobzol, line 38: I don't think that the display implementation should be showing the compiler flag. I'd rather remove the `Display` impl completely or just let it print the inner value directly.

collector/src/compile/execute/mod.rs
```diff
@@ -406,6 +409,9 @@ impl<'a> CargoProcess<'a> {
 
             let mut cmd = self.base_command(self.cwd, cargo_subcommand);
             cmd.arg("-p").arg(self.get_pkgid(self.cwd)?);
+            if self.parallel.0 != 1 {
+                cmd.env("RUSTC_THREAD_COUNT", self.parallel.0.to_string());
```
Kobzol, line 413: What is this environment variable? Shouldn't we be using `-Zthreads`?

database/src/lib.rs
```diff
@@ -21,6 +21,32 @@ intern!(pub struct Metric);
 intern!(pub struct Benchmark);
 intern!(pub struct TargetName);
 
+#[derive(Serialize, Deserialize, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub struct Parallel(pub u32);
+
+impl Parallel {
+    pub fn par_n(self) -> String {
```
Kobzol, line 28: What is this function for?

database/src/lib.rs
```diff
@@ -21,6 +21,32 @@ intern!(pub struct Metric);
...
+        format!("par{}", self.0)
+    }
+    // Default parallel frontend options
+    pub fn default_opts() -> Vec<Parallel> {
```
Kobzol, line 32: The default for now should just be 1 thread.

Kobzol (commented):
Left an initial review, sorry that this took so long, I didn't have much time for reviews, or Rust for that matter, in the past few months.

I'd like for the website changes to be split into a separate PR, and land just the collector + DB side first, as the PR is quite big already.

collector/src/compile/execute/mod.rs
heinwol: Hi, @Kobzol! As far as I understand, I'll be picking this pr up, supervised by @petrochenkov, as per https://github.com/rust-lang/compiler-team/issues/1005.

This is the first thing I want to address. The only mention of `RUSTC_THREAD_COUNT` i found across both rustc-perf and rust is this:

https://github.com/rust-lang/rustc-perf/blob/ef7a3cee15959874690ad6a312a938498c96edff/collector/src/bin/rustc-fake.rs#L68-L73

Since it's related entirely to the fake compiler, it's no use mentioning the variable elsewhere at all. I think I'll just add `.arg(format!(...,self.frontend_threads.0))`. However, this would produce an inconsistent behavior in the case the variable is provided to the fake compiler. Whether it will make any difference is a question I can't answer for now as I don't know much about rustc-perf and what fake compiler is needed for. I'll look into this later

heinwol:
see my branch with edits: https://github.com/heinwol/rustc-perf/tree/par-front-scenario

collector/src/compile/execute/mod.rs
Kobzol: Ah, interesting, I didn't know that this is a thing. Ok, why not, we can use this environment variable, I suppose.

Kobzol:
Hi! :wave: Can you please open a new PR with your modified branch? Thanks!

Kobzol:
The first part of this was implemented in https://github.com/rust-lang/rustc-perf/pull/2491, based on this PR. Thanks!

PR #2151, opened 2025-06-05
Make some of rust-timer's GitHub messages more succinct

miikkas:
Some initial feedback would be helpful. :)

Fixes: https://github.com/rust-lang/rustc-perf/issues/2149

Issue #2149, opened 2025-06-03
More succinct GitHub comments for rust-timer


miikkas:
Some of @rust-timer's comments are, arguably, a bit verbose, taking up vertical space in GitHub PR threads.

Would it be ok to condense the output, keeping its current information content and (user-)friendliness? I could give this a shot, though I'm not a native English speaker.

Kobzol:
Well, this is an eternal struggle 😆 It used to be shorter, but then people were complaining that it doesn't provide enough context for people on what to do. Usually when we make a change, people want it to be very explicit, but then everyone kind of gets used to that text and it becomes an annoyance. [Stroustrup's rule](https://mjtsai.com/blog/2016/12/20/stroustrups-rule-and-layering-over-time/) in practice :) Feel free to suggest modifications in a PR though :)

PR #2424, opened 2026-04-05
Shorten rust-timer's GitHub comments even more

miikkas:
This PR continues the optimization I started in #2151. (Relates to #2149.)

<details>
<summary>Mock comparison on PC, before and after</summary>

rust-timer-before rust-timer-after

</details>

<details>
<summary>Mock comparison on mobile, before and after</summary>

rust-timer-mobile-before rust-timer-mobile-after

</details>

### DIFF
--- a/site/src/github/comparison_summary.rs
+++ b/site/src/github/comparison_summary.rs
@@ -157,7 +157,7 @@ async fn summarize_run(
         if has_broken_benchmarks {
             " - BENCHMARK(S) FAILED"
         } else if is_regression {
-            " - please read the text below"
+            " - please read:"
         } else {
             " - no action needed"
         },
@@ -262,8 +262,7 @@ fn write_metric_summary(
     message: &mut String,
 ) {
     if !primary.is_relevant() && !secondary.is_relevant() {
-        message
-            .push_str("This benchmark run did not return any relevant results for this metric.\n");
+        message.push_str("This perf run didn't have relevant results for this metric.\n");
     } else {
         match visibility {
             DefaultMetricVisibility::Shown => {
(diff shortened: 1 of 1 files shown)


miikkas:
I have some further ideas on how to continue this, like

- moving some of the explanation and advice to <https://perf.rust-lang.org/help.html> and linking that page directly from rust-timer's GitHub comments, and
- hacking on the tables summarizing the perf run results (possibly with abbreviations with added explanation on the help page linked above etc.)

Would such changes be welcome?

nnethercote:
Needs a rustfmt run.

The word changes seem fine, it's a little shorter without losing any information.

As for further changes, the table seems the best place. If the labels in the left hand column could be made to fit on a single line, that would save some vertical space. Other than that I would be careful about removing information... we need a balance between enough explanation for newcomers vs. what works for experts. It doesn't feel like we can drastically shrink it.

miikkas:
Thanks for the review! I have amended the commit with rustfmt'd code.

> As for further changes, the table seems the best place. If the labels in the left hand column could be made to fit on a single line, that would save some vertical space.

Agreed. I will try working on this, and will open a separate PR if I succeed in saving space here. :)

Kobzol:
I think we're running into diminishing returns here, and I'm not sure if it's worth micro-optimizing further. At the end of the day it's a wall of text that new people might read, but people who see it every day will just ignore it, because they know what it says.

I like the removal of "the text below" though, that's nice compression :laughing:

miikkas:
> At the end of the day it's a wall of text that new people might read, but people who see it every day will just ignore it, because they know what it says.

This is the dilemma. How to best serve both seasoned and new contributors, each with different priorities.

The compression here and in the previous PR improve the SNR, I think, which benefits both user groups. But I agree we're probably nearing the limits within the current overall design.

nnethercote (approved):

nnethercote:
Worth merging, IMO, but I agree there is not much more juice to be squeezed here.

PR #2428, opened 2026-04-17
github-actions: fix all zizmor pedantic findings and add zizmor CI workflow

Dorcas-BD:
Fixes most findings reported by `zizmor --pedantic` across all workflow files.

Changes:
- Pin all action references to commit SHAs to fix `unpinned-uses`
- Add `persist-credentials: false` to all `actions/checkout` steps to fix `artipacked`
- Move `${{ toJson(needs) }}` to an env variable in the `conclusion` job to fix `template-injection`
- Fix `template-injection` in `nightly.yml` by using `${RUSTC_PERF_VERSION}` instead of `${{ env.RUSTC_PERF_VERSION }}` in shell
- Remove `cache-from`/`cache-to` and add `no-cache: true` to Docker build steps to fix `cache-poisoning`
- Add `name:` to anonymous `docker` and `conclusion` jobs to fix `anonymous-definition`
- Add explanatory comment to `contents: write` permission to fix `undocumented-permissions`
- Add `.github/zizmor.yml` to suppress findings that will be addressed in follow-up PRs
- Add `.github/workflows/zizmor.yml` to run zizmor on every PR and prevent regressions. Passes `GITHUB_TOKEN` for full audit coverage.

Pending findings (to be fixed in follow-up PRs):
- `unpinned-images`: postgres images need SHA256 digest pins, which require looking up the exact digest for each image version
- `concurrency-limits`: adding concurrency settings needs careful consideration per workflow to avoid cancelling jobs unintentionally
- `archived-uses`: `actions-rs/cargo` is archived and needs replacing with a direct `cargo build`, but this needs testing to ensure it doesn't break the nightly release
- `misfeature`: replacing `shell: cmd` on Windows needs testing to ensure the toolchain install commands still work
- `stale-action-refs`: simpleinfra and `actions-rs/cargo` refs need updating, but are tied to the above changes
- `superfluous-actions`: replacing `ncipollo/release-action` with `gh release` is a non-trivial change that needs careful testing

### DIFF
--- a/.github/workflows/beta-stable-benchmarks.yml
+++ b/.github/workflows/beta-stable-benchmarks.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           persist-credentials: false
 
@@ -22,7 +22,7 @@ jobs:
           rustup install beta
           rustup default beta
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
         with:
           key: beta
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           - 5432:5432
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@v4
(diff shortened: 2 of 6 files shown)


Kobzol:
Hi, could you please split this into multiple PRs?
- Things that don't change CI behavior.
- Things that change CI behavior (like the cache change - but I don't want to lose the caching...)
- The zizmor workflow

Thank you.

Dorcas-BD:
> Hi, could you please split this into multiple PRs?
> 
> * Things that don't change CI behavior.
> * Things that change CI behavior (like the cache change - but I don't want to lose the caching...)
> * The zizmor workflow
> 
> Thank you.

Thanks for the feedback! I'll split this into separate PRs as suggested and close this one.

PR #2430, opened 2026-04-23
Don't add `rollup=never` when perf does not deserve attention

JonathanBrouwer:


site/src/github/comparison_summary.rs
```diff
@@ -342,6 +343,13 @@ fn try_run_body(is_regression: bool) -> String {
         ""
     };
 
+    // We mark PRs as rollup=never if perf deserves attention, so we don't need to debug the perf in the rollup
+    // We don't remove `rollup=never` if perf does not deserve attention, as sometimes there are multiple perf. runs on the same PR,
+    // and it's not always the case that the latest version that gets approved is the one for which we ran perf.
+    let rollup_never = deserves_attention
+        .then_some("@bors rollup=never\n")
```
Kobzol:
```suggestion
        .then_some("@bors rollup=never")
```
The newline shouldn't be needed, as there is one already in the string template after `never}`.

Kobzol (commented):
Thanks! I think this makes sense.

site/src/github/comparison_summary.rs
JonathanBrouwer: Right, fixed

Kobzol (approved):
Thank you!

PR #2432, opened 2026-04-25
github-actions: pin actions to commit SHAs and fix zizmor security findings

Dorcas-BD:
Follow-up to #2428  (split as requested by @Kobzol ).

Fixes zizmor security findings that don't change CI behaviour.

Changes:
- Pin all action references to commit SHAs to fix `unpinned-uses`
- Add `persist-credentials: false` to all `actions/checkout` steps to fix `artipacked`
- Move `${{ toJson(needs) }}` to an env variable in the `conclusion` job to fix `template-injection`
- Fix `template-injection` in `nightly.yml` by using `${RUSTC_PERF_VERSION}` instead of `${{ env.RUSTC_PERF_VERSION }}` in shell
- Add `name:` to anonymous `docker` and `conclusion` jobs to fix `anonymous-definition`
- Add explanatory comment to `contents: write` permission in `nightly.yml` to fix `undocumented-permissions`

Note: `cache-poisoning` findings are not addressed here and will be handled in a separate PR.

### DIFF
--- a/.github/workflows/beta-stable-benchmarks.yml
+++ b/.github/workflows/beta-stable-benchmarks.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout the source code
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           persist-credentials: false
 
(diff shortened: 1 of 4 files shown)


Kobzol:
@marcoieni Could you please configure renovatebot here to manage GHA versions, and automatically pin the current versions, to ensure that the commit SHAs are okay? Like we did in bors (IIRC).

marcoieni:
Done:
 
- https://github.com/rust-lang/team/pull/2450 
- https://github.com/rust-lang/rustc-perf/pull/2437

PR #2441, opened 2026-05-01
enable `diesel`, `wg-grammar`, `typenum` and `deeply-nested-multi` benchmarks for the new solver

jdonszelmann:
These were shown to be relevant benchmarks in perf runs done in the [past for `-Znext-solver=globally`](https://perf.rust-lang.org/compare.html?start=fed46ffd5059e11669df1bd9406b02914c3fb73f&end=d58c64674341fa7eaac5e571c8d5b45efd427811&stat=instructions:u). This is data from October 2025, but the same benchmarks showed up to be relevant in my local tests.

We talked previously about making all benchmarks next-solver-enabled in [#t-infra > Perf runs with the next solver](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Perf.20runs.20with.20the.20next.20solver/with/591312764). Instead of doing so, I think that only enabling a few relevant ones seems wiser. That way, we don't spend time recompiling benchmarks that aren't likely to be affected by the next solver anyway.

r? @Kobzol 

I hope this is correct, I'm copying a bunch of directories for this... It seems that's how it was done in the past. The only really relevant files are the new `perf-config.json` files.

cc @lcnr


Kobzol:
This should be fine. You will also need to add the new benchmarks in https://github.com/rust-lang/rustc-perf/blob/f32786cd235c7be41dd2f364ebf68671c6e3ae0c/collector/src/benchmark_set/mod.rs#L67 and https://github.com/rust-lang/rustc-perf/blob/f32786cd235c7be41dd2f364ebf68671c6e3ae0c/collector/src/benchmark_set/mod.rs#L148 (please add them to set 1, as it is now faster than set 0).

Kobzol (approved):
Thank you! Let's try it.

PR #2457, opened 2026-05-05
renovate: don't update benchmarks

marcoieni:
Prevent renovate to raise PRs like https://github.com/rust-lang/rustc-perf/pull/2456


lqd (approved):
Thanks!

lqd:
We may also need the same for `collector/runtime-benchmarks` because it's also trying to update them AFAICT?

PR #2471, opened 2026-05-06
collector: scope precise_crabgrind feature to Linux

djc:
crabgrind does not seem to compile on macOS.

### DIFF
--- a/collector/benchlib/Cargo.toml
+++ b/collector/benchlib/Cargo.toml
@@ -16,9 +16,9 @@ serde_json = { workspace = true }
 
 libc = "0.2"
 flate2 = { version = "1", optional = true }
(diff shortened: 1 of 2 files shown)


lqd:
valgrind doesn’t work on macOS IIRC, are the runtime benchmarks profiled with cachegrind and perf only @Kobzol? Do they work on macs?

Kobzol:
No I'm pretty sure that they don't work on macOS at all :) They are indeed profiled either by Valgrind/Cachegrind or by `perf` on Linux.

Kobzol:
Well, time based profiling should still work, I guess.

Kobzol (approved):
So merging this might be a small improvement for non-Linux users.

PR #2475, opened 2026-05-12
Upgrade to hyper 1 and reqwest 0.13

djc:
This was triggered initially by [#t-release > Producing a smaller alternative to &#96;rustc-{version}-src&#96;](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/Producing.20a.20smaller.20alternative.20to.20.60rustc-.7Bversion.7D-src.60/with/593069002) discussing that there were many copies of OpenSSL included, which led me to wonder whether we could use rustls for some of these. reqwest 0.13 defaults to using rustls for TLS by default.

I also generally think it's better to keep dependencies up to date, including (but not only) because it limits duplicate dependencies.


djc:
This fails to build aws-lc-sys:

```
#21 13.84   --- stderr
#21 13.84 
#21 13.84   thread 'main' (1124) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-sys-0.40.0/builder/cc_builder.rs:731:9:
#21 13.84   ### COMPILER BUG DETECTED ###
#21 13.84   Your compiler (cc) is not supported due to a memcmp related bug reported in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189. We strongly recommend against using this compiler. 
```

`Dockerfile` is building on top of `ubuntu:20.04` which is fairly old. See

- https://github.com/rust-lang/rustc-perf/pull/2476

djc:
This unfortunately (still) results in duplicate reqwest versions, due to

- https://github.com/apache/arrow-rs-object-store/pull/707

Kobzol (commented):
The changes look simple enough, but at the moment this increases the dependency count from 464 to 488. I'm not sure if it's worth merging it at least until the object_store crate releases a new version.

djc:
> The changes look simple enough, but at the moment this increases the dependency count from 464 to 488. I'm not sure if it's worth merging it at least until the object_store crate releases a new version.

Added a commit that upgrades object_store to 0.14... but that further increases the dependency count by 15 (in the lockfile). It does at least get rid of reqwest 0.12.

Sources:

- object_store depends on rand 0.10, but there are still a few edges on rand 0.9 (including via termwiz, which hasn't had a release in quite a while)
- object_store depends on nix 0.31, but there are still edges on nix 0.29 (all via termwiz AFAICT)
- object_store depends on itertools 0.15 but the site still depends on itertools 0.10

djc:
The docker job failure looks like a fluke.

Kobzol:
Ok, I guess that we can go all in and just update everything to latest. I did that, and removed `termwiz` from ratatui (it shouldn't have been used anyway), which brought total deps to 467, and collector deps to 368. Before it was 457/365, so not that big of a regression.

Fun fact: the only remaining dependency edge on rand 0.9 now is `quinn-proto` :grin:

djc:
> Ok, I guess that we can go all in and just update everything to latest. I did that, and removed `termwiz` from ratatui (it shouldn't have been used anyway), which brought total deps to 467, and collector deps to 368. Before it was 457/365, so not that big of a regression.

Nice work!

Unfortunately rusqlite doesn't do MSRV so you'll need to downgrade it or bump the Rust version to 1.95 for `cfg_select!` in CI.

> Fun fact: the only remaining dependency edge on rand 0.9 now is `quinn-proto` 😁

Hopefully fixing this:

- https://github.com/quinn-rs/quinn/pull/2707

djc:
> Hopefully fixing this:
> 
> * [0.11.x: upgrade dependencies quinn-rs/quinn#2707](https://github.com/quinn-rs/quinn/pull/2707)

This has been published now.

Kobzol:
Thanks! It didn't help dependency count, but at least we got rid of old `rand`. Rebased over `master` to fix the OpenSSL downgrade that I had to do in the meantime.

Kobzol (approved):
Thank you, let's see if this works in production! :)

As a next step, we could remove
```toml
postgres-native-tls = "0.5"
native-tls = "0.2"
```
from the `database` crate (and replace by something like https://crates.io/crates/tokio-postgres-rustls ? not sure what's the canonical solution for Postgres + RustTLS these days), and that would allow us to get rid of OpenSSL completely 🎉

PR #2476, opened 2026-05-12
Upgrade Docker base to Ubuntu 24.04

djc:


Kobzol:
It is definitely good to get rid of 20.04, but 26.04 is a bit.. fresh. Would 24.04 also work?

Could you please also add a bit of context about why you want to land https://github.com/rust-lang/rustc-perf/pull/2475? Thanks!

djc:
> It is definitely good to get rid of 20.04, but 26.04 is a bit.. fresh. Would 24.04 also work?

Trying now, though I'm not clear on what you're worried about with 26.04 being fresh -- the file was first introduced in June 2020 with 20.04. Seems like newer is mostly likely to be better in most ways, and given that this was just never upgraded it seems better to me to get ahead of the curve.

> Could you please also add a bit of context about why you want to land #2475? Thanks!

Will add that in the PR description.

Kobzol:
I usually just defer updating to very new OS or package versions for services for which we don't have good tests. Let's try if 24.04 works on prod.

Kobzol (approved):

PR #2477, opened 2026-05-13
Update ratatui and some other terminal related crates

bjorn3:
This is required to get rid of outdated window-sys dependencies.


Kobzol:
This adds over 20 new dependencies :crying_cat_face: It's better (~13) with https://github.com/rust-lang/rustc-perf/pull/2478 though. Let's hope that it will be worth it.

Kobzol (approved):

PR #2478, opened 2026-05-13
Update dependencies to almost remove windows-targets dependency

bjorn3:
The last remaining dependencies preventing the removal of windows-targets
are ring and reqwest.

Builds on top of https://github.com/rust-lang/rustc-perf/pull/2477
Part of https://github.com/rust-lang/rust/issues/156459


Kobzol (approved):
Thanks!

PR #2487, opened 2026-06-27
Sort perf builds in the same order as the PR description

JonathanBrouwer:
I want this so the `iffy` jobs are at the top of the list, just like in the PR description. 
This is not currently the case, for example see https://github.com/rust-lang/rust/pull/158455

I have not tested this change, and I'm not even sure that removing the `.rev()` is correct.
Is there any way to test this other than in prod?
r? @Kobzol

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -40,13 +40,8 @@ pub async fn unroll_rollup(
 
     // Sort rolled up commits by their PR number in ascending order, so that they have the
     // same ordering as in the rollup PR description.
-    let mut unrolled_builds: Vec<UnrolledCommit> =
-        enqueue_unrolled_try_builds(&gh_client, rollup_merges, previous_master).await?;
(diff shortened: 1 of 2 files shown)


site/src/github.rs
```diff
@@ -40,13 +40,8 @@ pub async fn unroll_rollup(
 
     // Sort rolled up commits by their PR number in ascending order, so that they have the
```
Kobzol: The comment is now out of date.

Kobzol (commented):
There's not really any other way to test than on production :(

FWIW, this whole logic will hopefully be ported to bors (this year?), see https://github.com/rust-lang/bors/issues/285. But in the meantime this sounds like a reasonable hotfix.

site/src/github.rs
JonathanBrouwer: Fixed

Kobzol (approved):
Thanks, let's try.

PR #2491, opened 2026-07-01
Parallel frontend compiler support (perf backend only)

heinwol:
Supersedes https://github.com/rust-lang/rustc-perf/pull/2421, but only backend (profiler/bencher/database). As said in parent pr, parallel frontend support is implemented as additional dimension with corresponding row added to the database

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -160,6 +160,9 @@ The following options alter the behaviour of the `bench_local` subcommand.
   `IncrUnchanged`, `IncrPatched`, and `All`. The default is `All`. Note that
   `IncrFull` is always run if either of `IncrUnchanged` or `IncrPatched` are
   run (even if not requested).
+- `--frontend-threads <FRONTEND_THREADS>`: comma-separated list of thread counts for
+  the parallel frontend. Maps directly to the `-Zthreads` rustc option.
+  The default is `1`.
 - `--backends <BACKENDS>`: the codegen backends to be benchmarked. The possible
   choices are one or more (comma-separated) of `Llvm`, `Cranelift`. The default
   is `Llvm`.
@@ -204,7 +207,7 @@ and discover all benchmarks within them. If you only want to run benchmark(s) fr
 you can use this to speed up the runtime benchmarking or profiling commands.
 
 The `bench_runtime_local` command also shares some options with the `bench_local` command, notably
-`--id`, `--db`, `--cargo`, `--cargo-config`, `--include`, `--exclude` and `--iterations`. 
+`--id`, `--db`, `--cargo`, `--cargo-config`, `--include`, `--exclude` and `--iterations`.
 
 ### How to view the measurements on your own machine
 
@@ -489,6 +492,7 @@ The following options alter the behaviour of the `profile_local` subcommand.
   diff files will also be produced.
 - `--rustdoc <RUSTDOC>` as for `bench_local`.
 - `--scenarios <SCENARIOS>`: as for `bench_local`.
+- `--frontend-threads <FRONTEND_THREADS>`: as for `bench_local`.
 - `--backends <BACKENDS>`: as for `bench_local`.
 - `--jobs <JOB-COUNT>`: execute `<JOB-COUNT>` benchmarks in parallel. This is only allowed for certain
 profilers whose results are not affected by system noise (e.g. `callgrind` or `eprintln`).
@@ -511,7 +515,7 @@ build directory (at least Valgrind 3.22 is required), like this:
 
 ```
 DEP_VALGRIND=<path-to-valgrind-install>/include cargo run --release --bin collector \
-  --features precise-cachegrind profile_runtime cachegrind <RUSTC> <BENCHMARK_NAME> 
+  --features precise-cachegrind profile_runtime cachegrind <RUSTC> <BENCHMARK_NAME>
 ```
 
 ## Codegen diff
@@ -534,7 +538,7 @@ binary artifacts (executables, libraries). You can compare the binary statistics
       [--profile <Debug|Opt>] \
       [--backend <Llvm|Cranelift>]
   ```
-  
+
   You can also compare (diff) the size statistics between two compilers:
   ```bash
   ./target/release/collector binary_stats compile `<rustc>` --include <benchmark name> --rustc2 <rustc2>
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -36,6 +36,7 @@ use collector::benchmark_set::{get_benchmark_set, BenchmarkSetId, BenchmarkSetMe
 use collector::codegen::{codegen_diff, CodegenType};
 use collector::compile::benchmark::category::Category;
 use collector::compile::benchmark::codegen_backend::CodegenBackend;
+use collector::compile::benchmark::parallel_frontend::FrontendThreads;
 use collector::compile::benchmark::profile::Profile;
 use collector::compile::benchmark::scenario::Scenario;
 use collector::compile::benchmark::target::Target;
@@ -116,6 +117,7 @@ struct CompileBenchmarkConfig {
     self_profile_storage: Option<Box<dyn SelfProfileStorage>>,
     bench_rustc: bool,
     targets: Vec<Target>,
+    frontend_threads_counts: Vec<FrontendThreads>,
 }
 
 struct RuntimeBenchmarkConfig {
@@ -166,50 +168,54 @@ fn generate_diffs(
     benchmarks: &[Benchmark],
     profiles: &[Profile],
     scenarios: &[Scenario],
+    frontend_threads_counts: &[FrontendThreads],
     errors: &mut BenchmarkErrors,
     profiler: &Profiler,
 ) -> Vec<PathBuf> {
     let mut annotated_diffs = Vec::new();
     for benchmark in benchmarks {
         for &profile in profiles {
-            for scenario in scenarios.iter().flat_map(|scenario| {
-                if profile.is_doc() && scenario.is_incr() {
-                    return vec![];
-                }
-                match scenario {
-                    Scenario::Full | Scenario::IncrFull | Scenario::IncrUnchanged => {
-                        vec![format!("{:?}", scenario)]
+            for &frontend_threads in frontend_threads_counts {
+                for scenario in scenarios.iter().flat_map(|scenario| {
+                    if profile.is_doc() && scenario.is_incr() {
+                        return vec![];
                     }
-                    Scenario::IncrPatched => (0..benchmark.patches.len())
-                        .map(|i| format!("{scenario:?}{i}"))
-                        .collect::<Vec<_>>(),
-                }
-            }) {
-                let filename = |prefix, id| {
-                    format!(
-                        "{}-{}-{}-{:?}-{}{}",
-                        prefix,
-                        id,
-                        benchmark.name,
-                        profile,
-                        scenario,
-                        profiler.postfix()
-                    )
-                };
-                let id_diff = format!("{id1}-{id2}");
-                let prefix = profiler.prefix();
-                let prefix2 = profiler.prefix2();
-                let left = out_dir.join(filename(prefix, id1));
-                let right = out_dir.join(filename(prefix, id2));
-                let output = out_dir.join(filename(&format!("{prefix2}-diff"), &id_diff));
+                    match scenario {
+                        Scenario::Full | Scenario::IncrFull | Scenario::IncrUnchanged => {
+                            vec![format!("{:?}", scenario)]
+                        }
+                        Scenario::IncrPatched => (0..benchmark.patches.len())
+                            .map(|i| format!("{scenario:?}{i}"))
+                            .collect::<Vec<_>>(),
+                    }
+                }) {
+                    let filename = |prefix, id| {
+                        format!(
+                            "{}-{}-{}-{:?}-{}-frontend_threads_{}{}",
+                            prefix,
+                            id,
+                            benchmark.name,
+                            profile,
+                            scenario,
+                            frontend_threads.get(),
+                            profiler.postfix()
+                        )
+                    };
+                    let id_diff = format!("{id1}-{id2}");
+                    let prefix = profiler.prefix();
+                    let prefix2 = profiler.prefix2();
+                    let left = out_dir.join(filename(prefix, id1));
+                    let right = out_dir.join(filename(prefix, id2));
+                    let output = out_dir.join(filename(&format!("{prefix2}-diff"), &id_diff));
 
-                if let Err(e) = profiler.diff(&left, &right, &output) {
-                    errors.incr();
-                    eprintln!("collector error: {e:?}");
-                    continue;
-                }
+                    if let Err(e) = profiler.diff(&left, &right, &output) {
+                        errors.incr();
+                        eprintln!("collector error: {e:?}");
+                        continue;
+                    }
 
-                annotated_diffs.push(output);
+                    annotated_diffs.push(output);
+                }
             }
         }
     }
@@ -227,6 +233,7 @@ fn profile_compile(
     backends: &[CodegenBackend],
     errors: &mut BenchmarkErrors,
     targets: &[Target],
+    frontend_threads_counts: &[FrontendThreads],
 ) {
     eprintln!("Profiling {} with {:?}", toolchain.id, profiler);
     if let Profiler::SelfProfile = profiler {
@@ -248,6 +255,7 @@ fn profile_compile(
                 toolchain,
                 Some(1),
                 targets,
+                frontend_threads_counts,
                 // We always want to profile everything
                 &hashbrown::HashSet::new(),
             ));
@@ -411,6 +419,27 @@ struct CompileTimeOptions {
     /// It should be a path to the `clippy-driver` binary.
     #[arg(long)]
     clippy: Option<PathBuf>,
+
+    /// Parallel frontend thread count in comma-separated list
+    #[arg(long = "frontend-threads", value_delimiter = ',')]
+    frontend_threads_counts: Vec<u32>,
+}
+
+impl CompileTimeOptions {
+    fn validate_and_normalize_frontend_threads(&self) -> Vec<FrontendThreads> {
+        if self.frontend_threads_counts.is_empty() {
+            return FrontendThreads::default_threads_counts();
+        }
+
+        let mut threads = self
+            .frontend_threads_counts
+            .iter()
+            .map(|&x| FrontendThreads::new(x))
+            .collect::<Vec<_>>();
+        threads.sort();
+        threads.dedup();
+        threads
+    }
 }
 
 #[derive(Debug, clap::Args)]
@@ -992,6 +1021,7 @@ fn main_result() -> anyhow::Result<i32> {
             purge,
         } => {
             log_db(&db);
+            let frontend_threads_counts = opts.validate_and_normalize_frontend_threads();
             let profiles = opts.profiles.0;
             let scenarios = opts.scenarios.0;
             let backends = opts.codegen_backends.0;
@@ -1044,6 +1074,7 @@ fn main_result() -> anyhow::Result<i32> {
                 },
                 bench_rustc: bench_rustc.bench_rustc,
                 targets: vec![Target::host()],
+                frontend_threads_counts,
             };
 
             rt.block_on(run_benchmarks(conn.as_mut(), shared, Some(config), None))?;
@@ -1084,6 +1115,7 @@ fn main_result() -> anyhow::Result<i32> {
 
             let profiles = &opts.profiles.0;
             let scenarios = &opts.scenarios.0;
+            let frontend_threads_counts = opts.validate_and_normalize_frontend_threads();
             let backends = &opts.codegen_backends.0;
 
             let mut benchmarks = get_compile_benchmarks(&compile_benchmark_dir, (&local).into())?;
@@ -1123,6 +1155,7 @@ fn main_result() -> anyhow::Result<i32> {
                         backends,
                         &mut errors,
                         &[Target::host()],
+                        &frontend_threads_counts,
                     );
                     Ok(id)
                 };
@@ -1141,6 +1174,7 @@ fn main_result() -> anyhow::Result<i32> {
                     &benchmarks,
                     profiles,
                     scenarios,
+                    &frontend_threads_counts,
                     &mut errors,
                     &profiler,
                 );
@@ -1680,6 +1714,7 @@ async fn create_benchmark_configs(
             },
             bench_rustc,
             targets: vec![job.target().into()],
+            frontend_threads_counts: FrontendThreads::default_threads_counts(),
         })
     } else {
         None
@@ -2135,6 +2170,11 @@ async fn bench_published_artifact(
     } else {
         Scenario::all_non_incr()
     };
+    let frontend_threads_counts = if collector::version_supports_parallel_frontend(&toolchain.id) {
+        FrontendThreads::default_threads_counts()
+    } else {
+        vec![FrontendThreads::new(1)]
+    };
 
     // Exclude benchmarks that don't work with a stable compiler.
     let mut compile_benchmarks = get_compile_benchmarks(dirs.compile, CompileBenchmarkFilter::All)?;
@@ -2168,6 +2208,7 @@ async fn bench_published_artifact(
             self_profile_storage: None,
             bench_rustc: false,
             targets: vec![Target::host()],
+            frontend_threads_counts,
         }),
         Some(RuntimeBenchmarkConfig::new(
             runtime_suite,
@@ -2274,6 +2315,7 @@ async fn bench_compile(
                     &shared.toolchain,
                     config.iterations,
                     &config.targets,
+                    &config.frontend_threads_counts,
                     &collector.measured_compile_test_cases,
                 ))
                 .await
--- a/collector/src/compile/benchmark/mod.rs
+++ b/collector/src/compile/benchmark/mod.rs
@@ -1,5 +1,6 @@
 use crate::compile::benchmark::category::Category;
 use crate::compile::benchmark::codegen_backend::CodegenBackend;
+use crate::compile::benchmark::parallel_frontend::FrontendThreads;
 use crate::compile::benchmark::patch::Patch;
 use crate::compile::benchmark::profile::Profile;
 use crate::compile::benchmark::scenario::Scenario;
@@ -19,6 +20,7 @@ use tempfile::TempDir;
 
 pub mod category;
 pub mod codegen_backend;
+pub mod parallel_frontend;
 pub(crate) mod patch;
 pub mod profile;
 pub mod scenario;
@@ -195,6 +197,7 @@ impl Benchmark {
         profile: Profile,
         backend: CodegenBackend,
         target: Target,
+        frontend_threads: FrontendThreads,
     ) -> CargoProcess<'a> {
         let mut cargo_args = self
             .config
@@ -236,6 +239,7 @@ impl Benchmark {
             touch_file: self.config.touch_file.clone(),
             jobserver: None,
             target,
+            frontend_threads,
             workspace_package: self.config.package.clone(),
         }
     }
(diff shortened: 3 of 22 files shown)


Kobzol:
Thanks for the PR! Let's revert the last commit; the CI failure shows why we need to have a custom environment variable, so that it will also work with a stable Rust compiler.

heinwol:
I'm not sure I quite like a separate environment variable for that; don't we have a better mechanism? Also, maybe a collector run shouldn't silently revert to single-threaded scenario in an unsupported version, which it currently does, for my understanding? However, it seems like it's a common behavior in `collector/src/lib.rs::async fn bench_published_artifact`:
```rust
// ...
    let profiles = if collector::version_supports_doc(&toolchain.id) {
        vec![Profile::Check, Profile::Debug, Profile::Doc, Profile::Opt]
    } else {
        vec![Profile::Check, Profile::Debug, Profile::Opt]
    };
    let scenarios = if collector::version_supports_incremental(&toolchain.id) {
        Scenario::all()
    } else {
        Scenario::all_non_incr()
    };
    let frontend_threads_counts = if collector::version_supports_parallel_frontend(&toolchain.id) {
        FrontendThreads::default_opts()
    } else {
        vec![FrontendThreads(1)]
    };
// ...
```
So it might be i'm looking too hard into it...

Kobzol:
`bench_published_artifact` is only used for benchmarking stable versions of the compiler, which is ~never done locally, and we need to keep compatibility with Rust going all the way back to 1.26.0, so I wouldn't worry about that.

collector/src/bin/collector.rs
```diff
@@ -166,50 +168,54 @@ fn generate_diffs(
...
+                }) {
+                    let filename = |prefix, id| {
+                        format!(
+                            "{}-{}-{}-{:?}-{}-{}{}",
+                            prefix,
+                            id,
+                            benchmark.name,
+                            profile,
+                            scenario,
+                            frontend_threads.par_n(),
```
Kobzol: Let's remove the `par_n` function, it contains a hardcoded string prefix that is then used in several places. That shouldn't belong in the database module.

collector/src/bin/collector.rs
```diff
@@ -411,6 +419,14 @@ struct CompileTimeOptions {
     /// It should be a path to the `clippy-driver` binary.
     #[arg(long)]
     clippy: Option<PathBuf>,
+
+    /// Parallel frontend thread count in comma-separated list
+    #[arg(
+        long = "frontend-threads",
+        value_delimiter = ',',
+        default_value = "1,4"
```
Kobzol:
```suggestion
        default_value = "1"
```
The default should just be one for now, we don't need to test multiple values by default, it would just make local runs unnecessarily slower.

collector/src/bin/collector.rs
```diff
@@ -1680,6 +1705,7 @@ async fn create_benchmark_configs(
             },
             bench_rustc,
             targets: vec![job.target().into()],
+            frontend_threads: FrontendThreads::default_opts(),
```
Kobzol:
```suggestion
            frontend_threads: FrontendThreads::default_value(),
```
And also change the default to just `[1]`.

collector/src/bin/collector.rs
```diff
@@ -2135,6 +2161,11 @@ async fn bench_published_artifact(
     } else {
         Scenario::all_non_incr()
     };
+    let frontend_threads_counts = if collector::version_supports_parallel_frontend(&toolchain.id) {
+        FrontendThreads::default_opts()
+    } else {
+        vec![FrontendThreads(1)]
```
Kobzol: Hmm, this is tricky, because for the old version that doesn't support the flag, we would either:
- Not pass the `RUSTC_THREAD_COUNT` at all (which happens now). That works for the old compiler, but if we change the default thread count in the future, then setting `--frontend-threads=1` will actually use the default number of threads, which is misleading.
- Still try to pass `-Zthreads=1`. That will ensure that we are using the right number of threads, but it will break benchmarking on Rust <1.33.0.

I would suggest changing [this](https://github.com/rust-lang/rustc-perf/blob/4f11ea4b7081b5f4501dc48cdc8da819816baf96/collector/src/compile/execute/mod.rs#L412) to *always* pass the flag. It is better to break benchmarking on Rust <1.33.0, which likely won't be needed in the near future, than to use an unexpected thread count by mistake.

collector/src/compile/execute/mod.rs
```diff
@@ -406,6 +409,9 @@ impl<'a> CargoProcess<'a> {
 
             let mut cmd = self.base_command(self.cwd, cargo_subcommand);
             cmd.arg("-p").arg(self.get_pkgid(self.cwd)?);
+            if self.frontend_threads.0 != 1 {
+                cmd.env("RUSTC_THREAD_COUNT", self.frontend_threads.0.to_string());
+            }
```
Kobzol:
```suggestion
                cmd.env("RUSTC_THREAD_COUNT", self.frontend_threads.0.to_string());
```

collector/src/compile/execute/profiler.rs
```diff
@@ -131,11 +131,16 @@ impl Processor for ProfileProcessor<'_> {
...
                 format!(
-                    "{}-{}-{}-{:?}-{}",
-                    prefix, self.id, data.name, data.profile, data.scenario_str
+                    "{}-{}-{}-{:?}-{}-{}",
+                    prefix,
+                    self.id,
+                    data.name,
+                    data.profile,
+                    data.scenario_str,
+                    data.frontend_threads.par_n()
```
Kobzol: Same here, just add `par-` here manually.

collector/src/self_profile.rs
```diff
@@ -74,6 +77,7 @@ impl SelfProfileId {
                     .join(codegen_backend.to_string())
                     .join(profile.to_string())
                     .join(scenario.to_id())
+                    .join(frontend_threads.par_n())
```
Kobzol: And here, I'd suggest naming it `frontend-threads-{n}`

collector/src/self_profile.rs
```diff
@@ -53,13 +55,14 @@ impl SelfProfileId {
                 .join(profile.to_string())
                 .join(scenario.to_id())
                 .join(format!("self-profile-{collection}.mm_profdata.sz")),
-            //  self-profile/<artifact id>/<benchmark>/<target>/<backend>/<profile>/<scenario>
+            //  self-profile/<artifact id>/<benchmark>/<target>/<backend>/<profile>/<scenario>/<frontend_threads>
```
Kobzol: Hmm, if we change this path, then existing profiles will not be loaded correctly. It's not such a big deal, because we usually only care about the most recent self-profiles anyway, but still, it's annoying that once we'd merge this PR, all self-profile links would suddenly stop working.

It would be great if we could implement some fallback, where we first try if a path without the frontend thread suffix exists, and if yes, use that. If not, then try the path with the parallel frontend threads suffix.

And always store new profiles under the new path.

collector/README.md
```diff
@@ -160,6 +160,9 @@ The following options alter the behaviour of the `bench_local` subcommand.
   `IncrUnchanged`, `IncrPatched`, and `All`. The default is `All`. Note that
   `IncrFull` is always run if either of `IncrUnchanged` or `IncrPatched` are
   run (even if not requested).
+- `--frontend-threads <FRONTEND_THREADS>`: comma-separated list of thread counts for
+  the parallel frontend. Maps directly to the `-Zthreads` rustc option.
+  The default is `1,4`.
```
Kobzol:
```suggestion
  The default is `1`.
```

database/src/bin/postgres-to-sqlite.rs
```diff
@@ -189,12 +189,13 @@ impl Table for PstatSeries {
     }
 
     fn postgres_select_statement(&self, _since_weeks_ago: Option<u32>) -> String {
-        "select id, crate, profile, scenario, backend, target, metric from ".to_string()
+        "select id, crate, profile, scenario, backend, target, metric, frontend_threads from "
```
Kobzol:
```suggestion
        "select id, crate, profile, scenario, backend, target, frontend_threads, metric from "
```
The metric is the output, while everything else (apart from ID) is the "input", so I'd like to keep the ordering consistent.

database/src/bin/postgres-to-sqlite.rs
```diff
@@ -189,12 +189,13 @@ impl Table for PstatSeries {
...
+            .to_string()
             + self.name()
     }
 
     fn sqlite_insert_statement(&self) -> &'static str {
-        "insert into pstat_series (id, crate, profile, scenario, backend, target, metric) VALUES (?, ?, ?, ?, ?, ?, ?)"
+        "insert into pstat_series (id, crate, profile, scenario, backend, target, metric, frontend_threads) VALUES (?, ?, ?, ?, ?, ?, ?, ?)"
```
Kobzol:
```suggestion
        "insert into pstat_series (id, crate, profile, scenario, backend, target, frontend_threads, metric) VALUES (?, ?, ?, ?, ?, ?, ?, ?)"
```

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -251,11 +252,11 @@ impl Table for PstatSeries {
     }
 
     fn sqlite_attributes() -> &'static str {
-        "id, crate, profile, scenario, backend, target, metric"
+        "id, crate, profile, scenario, backend, target, metric, frontend_threads"
```
Kobzol:
```suggestion
        "id, crate, profile, scenario, backend, target, frontend_threads, metric"
```

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -251,11 +252,11 @@ impl Table for PstatSeries {
...
     }
 
     fn postgres_attributes() -> &'static str {
-        "id, crate, profile, scenario, backend, target, metric"
+        "id, crate, profile, scenario, backend, target, metric, frontend_threads"
```
Kobzol:
```suggestion
        "id, crate, profile, scenario, backend, target, frontend_threads, metric"
```

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -243,6 +243,7 @@ struct PstatSeriesRow<'a> {
     backend: &'a str,
     target: &'a str,
     metric: &'a str,
+    frontend_threads: u32,
```
Kobzol:
```suggestion
    frontend_threads: u32,
    metric: &'a str,
```

database/src/bin/sqlite-to-postgres.rs
```diff
@@ -272,6 +273,7 @@ impl Table for PstatSeries {
                 backend: row.get_ref(4).unwrap().as_str().unwrap(),
                 target: row.get_ref(5).unwrap().as_str().unwrap(),
                 metric: row.get_ref(6).unwrap().as_str().unwrap(),
+                frontend_threads: row.get(7).unwrap(),
```
Kobzol:
```suggestion
                frontend_threads: row.get(6).unwrap(),
                metric: row.get_ref(7).unwrap().as_str().unwrap(),
```

database/src/pool/postgres.rs
```diff
@@ -441,6 +441,12 @@ static MIGRATIONS: &[&str] = &[
     "#,
     r#"ALTER TABLE job_queue ADD COLUMN is_optional BOOLEAN NOT NULL DEFAULT FALSE"#,
     r#"ALTER TABLE benchmark_request ADD COLUMN targets TEXT NOT NULL DEFAULT ''"#,
+    // Add frontend_threads to pstat_series
+    r#"
+    alter table pstat_series add frontend_threads int not null default 1;
+    alter table pstat_series drop constraint test_case;
+    alter table pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, metric, frontend_threads);
```
```suggestion
    ALTER TABLE pstat_series add frontend_threads int not null default 1;
    ALTER TABLE pstat_series drop constraint test_case;
    ALTER TABLE pstat_series add constraint test_case UNIQUE(crate, profile, scenario, backend, target, frontend_threads, metric);
```

database/src/pool/postgres.rs
```diff
@@ -633,8 +639,8 @@ impl PostgresConnection {
                     .await
                     .unwrap(),
                 get_error: conn.prepare("select context, message from error where aid = $1").await.unwrap(),
-                insert_pstat_series: conn.prepare("insert into pstat_series (crate, profile, scenario, backend, target, metric) VALUES ($1, $2, $3, $4, $5, $6) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
-                select_pstat_series: conn.prepare("select id from pstat_series where crate = $1 and profile = $2 and scenario = $3 and backend = $4 and target = $5 and metric = $6").await.unwrap(),
+                insert_pstat_series: conn.prepare("insert into pstat_series (crate, profile, scenario, backend, target, metric, frontend_threads) VALUES ($1, $2, $3, $4, $5, $6, $7) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
```
```suggestion
                insert_pstat_series: conn.prepare("insert into pstat_series (crate, profile, scenario, backend, target, frontend_threads, metric) VALUES ($1, $2, $3, $4, $5, $6, $7) ON CONFLICT DO NOTHING RETURNING id").await.unwrap(),
```

database/src/pool/postgres.rs
```diff
@@ -633,8 +639,8 @@ impl PostgresConnection {
...
+                select_pstat_series: conn.prepare("select id from pstat_series where crate = $1 and profile = $2 and scenario = $3 and backend = $4 and target = $5 and metric = $6 and frontend_threads = $7").await.unwrap(),
```
```suggestion
                select_pstat_series: conn.prepare("select id from pstat_series where crate = $1 and profile = $2 and scenario = $3 and backend = $4 and target = $5 and frontend_threads = $6 and metric = $7 ").await.unwrap(),
```

database/src/pool/postgres.rs
```diff
@@ -885,7 +891,7 @@ where
             pstat_series: self
                 .conn()
                 .query(
-                    "select id, crate, profile, scenario, backend, target, metric from pstat_series;",
+                    "select id, crate, profile, scenario, backend, target, metric, frontend_threads from pstat_series;",
```
```suggestion
                    "select id, crate, profile, scenario, backend, target, frontend_threads, metric from pstat_series;",
```

database/src/pool/sqlite.rs
```diff
@@ -448,6 +449,25 @@ static MIGRATIONS: &[Migration] = &[
...
+            scenario text not null,
+            backend text not null,
+            target text not null default 'x86_64-unknown-linux-gnu',
+            metric text not null,
+            frontend_threads integer not null default 1,
+            UNIQUE(crate, profile, scenario, backend, target, metric, frontend_threads)
+        );
+        insert into pstat_series_with_frontend_threads select id, crate, profile, scenario, backend, 'x86_64-unknown-linux-gnu', metric, 1 from pstat_series;
+        drop table pstat_series;
+        alter table pstat_series_with_frontend_threads rename to pstat_series;
```
Kobzol:
```suggestion
        CREATE TABLE pstat_series_with_frontend_threads(
            id integer primary key not null,
            crate text not null references benchmark(name) on delete cascade on update cascade,
            profile text not null,
            scenario text not null,
            backend text not null,
            target text not null,
            frontend_threads integer not null,
            metric text not null,
            UNIQUE(crate, profile, scenario, backend, target, frontend_threads, metric)
        );
        INSERT INTO pstat_series_with_frontend_threads select id, crate, profile, scenario, backend, target, 1, metric from pstat_series;
        DROP TABLE pstat_series;
        ALTER TABLE pstat_series_with_frontend_threads rename to pstat_series;
```

database/src/pool/sqlite.rs
```diff
@@ -575,7 +595,7 @@ impl Connection for SqliteConnection {
         let pstat_series = self
             .raw()
             .prepare(
-                "select id, crate, profile, scenario, backend, target, metric from pstat_series;",
+                "select id, crate, profile, scenario, backend, target, metric, frontend_threads from pstat_series;",
```
Kobzol:
```suggestion
                "select id, crate, profile, scenario, backend, target, frontend_threads, metric from pstat_series;",
```
And everywhere else.

database/src/lib.rs
```diff
@@ -562,7 +588,15 @@ pub struct Index {
...
     /// For legacy reasons called `pstat_series` in the database, and so the name is kept here.
-    pstat_series: Indexed<(Benchmark, Profile, Scenario, CodegenBackend, Target, Metric)>,
+    pstat_series: Indexed<(
+        Benchmark,
+        Profile,
+        Scenario,
+        CodegenBackend,
+        Target,
+        Metric,
+        FrontendThreads,
```
```suggestion
        FrontendThreads,
        Metric,
```

database/src/selector.rs
```diff
@@ -193,6 +193,7 @@ pub struct CompileBenchmarkQuery {
     profile: Selector<Profile>,
     backend: Selector<CodegenBackend>,
     metric: Selector<crate::Metric>,
+    frontend_threads: Selector<FrontendThreads>,
     target: Selector<Target>,
```
Kobzol:
```suggestion
    target: Selector<Target>,
    frontend_threads: Selector<FrontendThreads>,
    metric: Selector<crate::Metric>,
```

database/schema.md
```diff
@@ -16,27 +16,28 @@ Below are some diagrams showing the basic layout of the database schema for thes
...
+│ │ pstat_series      │  │ pstat    │   │                  │
+│ ├───────────────────┤  ├──────────┤   │                  │
+│ │ id *              │◄┐│ id *     │   │                  │
+└─┤ crate             │ └┤ series   │   │                  │
+  │ profile           │  │ aid      ├───┼──────────────────┘
+  │ scenario          │  │ cid      │   │
+  │ backend           │  │ value    ├───┘
+  │ metric            │  └──────────┘
+  │ target            |
+  │ frontend_threads  |
```
Kobzol:
```suggestion
  │ frontend_threads  │  └──────────┘
  │ target            |
  │ metric  |
```

database/schema.md
```diff
@@ -110,6 +111,7 @@ Columns:
 * **scenario** (`text`): Describes how much of the incremental cache is full. An empty incremental cache means that the compiler must do a full build.
 * **backend** (`text`): Codegen backend used for compilation, for example 'llvm'
 * **metric** (`text`): the type of metric being collected.
+* **frontend_threads** (`text`): Parallel frontend thread count ('-Zthreads=N').
```
Kobzol:
```suggestion
* **frontend_threads** (`text`): Parallel frontend thread count ('-Zthreads=N').
* **metric** (`text`): the type of metric being collected.
```
Seems like we are missing the `target` here (pre-existing), could you please add it? Thanks!

Kobzol (commented):
Thanks for separating the PR! Did a first review pass. Adding a new benchmark axis is always cumbersome, as it requires changing many places, and this one is also tricky, because it doesn't have a closed set of values, but is just a number. But in general it looks good, just

Please revert all the `site` changes, the integration with the frontend and also the job queuing system should be done in a follow-up PR. For now everything on production will just assume frontend count 1.

heinwol:
Thanks for the fast yet sizable review! I'm sorry for calling out for you so early; i'm not sure this pr is quite review-ready. I'll try to resolve what you commented for now and then dive more into how this works! As until now all i did is brainless-ish renaming, lol

Kobzol:
To be fair, adding a new benchmark axis mostly involves a lot of brainless-ish renaming :) I think that the PR is mostly good to go, though it will require a couple of follow-up PRs to:
1. Add integration in the website backend and frontend
2. Modify the job queue to be able to distinguish that different jobs can have different frontend thread counts

collector/src/self_profile.rs
heinwol: yup, i think a fallback path is a good idea

heinwol:
The dumb part is more or less done, now i'll be dealing with threads argument propagation and fallback legacy paths

EDIT: oh, and the `par_n` one

Kobzol:
> also, a question: should we include frontend_threads into glossary?

Yes, I think it would be nice to mention it there!

collector/src/compile/benchmark/mod.rs
```diff
@@ -297,32 +303,36 @@ impl Benchmark {
...
-                                target,
-                                already_computed,
-                            )
-                        })
-                        .copied()
-                        .collect::<Vec<Scenario>>();
-                    if remaining_scenarios.is_empty() {
-                        continue;
-                    }
+                    for frontend_threads in frontend_threads_counts {
```
Kobzol: I realized that this is not needed, in theory, because `-Zthreads` shouldn't affect the result of the compilation or the target directory (again, *in theory*). So we could just reuse the already prepared target directories for multiple values of `-Zthreads`, and thus considerably save time here.

But that can be done as a follow-up, this PR is already large as-is.

collector/src/compile/execute/mod.rs
Kobzol:
```suggestion
                cmd.env("RUSTC_THREAD_COUNT", self.frontend_threads.0.to_string());
```
Because we cannot detect the default number of threads used by the compiler

database/schema.md
```diff
@@ -101,14 +102,16 @@ Columns:
...
-of a crate, profile, scenario and the metric being collected.
+of a crate, profile, scenario, metric and parallel frontend thread count being collected.
 
 Columns:
 
 * **crate** (`text`) (aka `benchmark`): The benchmarked crate which might be a crate from crates.io or a crate made specifically to stress some part of the compiler.
 * **profile** (`text`): What type of compilation is happening - check build, optimized build (a.k.a. release build), debug build, or doc build.
 * **scenario** (`text`): Describes how much of the incremental cache is full. An empty incremental cache means that the compiler must do a full build.
 * **backend** (`text`): Codegen backend used for compilation, for example 'llvm'
+* **target** (`text NOT NULL`): target triple of the artifact, defaults to `x86_64-unknown-linux-gnu`
```
Kobzol:
```suggestion
* **target** (`text`): target triple of the artifact, defaults to `x86_64-unknown-linux-gnu`
```
For consistency with the other descriptions.

Kobzol (commented):
Thanks! Did another review pass and tried it locally. I think that the only thing required to fix before we can merge this is the backwards compatibility with the self-profile URLs.

database/schema.md
heinwol: Oh, seems like i screwed it a little, thanks. On the other hand, i think descriptions in this document aren't consistent between sections, e.g. in `benchmark_request` we specify `NOT NULL`.

On a side note, i tried putting the schema diagram into mermaid.js, but it appeared to be very cumbersome and affected readability negatively.
<details>
<summary>Unfinished attempt:</summary>

```mermaid
graph LR
    subgraph benchmark["benchmark"]
        B1["name *"]
        B2["stabilized"]
    end

    subgraph collection["collection"]
        C["`id *
        perf_commit`"]
    end

    subgraph artifact["artifact"]
        A["`id *
        name
        date
        type`"]
    end

    subgraph pstat_series["pstat_series"]
        PS["`id *`"]
        CRATE["crate"]
        PS2["`profile
        scenario
        backend
        target
        frontend_threads
        metric`"]
    end

    subgraph pstat["pstat"]
      direction TB
        P1["id *"]
        P2["series"]
        P3["aid"]
        P4["cid"]
        P5["value"]
    end

    CRATE --> benchmark
    P2 --> pstat_series
    P3 --> artifact
    P4 --> collection
```

</details>

Maybe we could write a proper ER diagram here, lol. Not so good for readability, but good for maintenance...

collector/src/bin/collector.rs
heinwol: Actually, could you explain, please, why can't we get information about rustc version in rustc-fake? Or at least we could pass a variable alongside `RUSTC_THREAD_COUNT` (from `CargoProcess`, it seems to be the only entrypoint) -- as we already resort to env variables -- to indicate whether `-Zthreads` flag can be included at all? Feels like i'm missing something

**EDIT**

Or I just made up a nonexistent problem, since everything can be done at the collector level, not rustc-fake. For now, I can't understand what prohibits us to both preserve correctness in forward-compatibility and not to halt benchmarks with legacy compiler.

What I suggest:
- We set default option once (maybe just in `default_opts`, period)
- We check for compatibility in `CargoProcess::run_rustc`
  - If we're compatible, pass the option down
  - If not, emit a warning and ignore the option
- `rustc-fake` picks up the option if it's set, or doesn't include `-Zthreads` at all

So yeah, i suggest just emitting a warning in legacy benchmarks

**EDIT2**

In this scheme it appears we're future-proofed: whenever we change the default counts set (e.g. `1` -> `1,4`)  the legacy path stays the same and the new default is used. If we pass the option explicitly, it would apply whatsoever

Kobzol: The future proofing is related to rustc's defaults, not rustc-perf's default. Let's say that in a few months, rustc will start using 4 threads *by default* (if you don't pass `-Zthreads`), rather than 1 (which it defaults to today). Then if we don't overwrite `-Zthreads` to `1` with `--frontend-threads=1`, then `--frontend-threads=1` will suddenly execute essentially `--frontend-threads=4`.

Regarding the other question, figuring out if the compiler supports the `-Zthreads` flag is doable, but not worth the effort. We don't need to support such old compilers outside of stable benchmarks, and we only backfill those once every few years or so.

heinwol: Soooo if we're on the same page then we can always set this option by ourselves. In case we're not on legacy. With the caveat being if rustc's defaults change, we'll have to update our defaults manually

collector/src/compile/benchmark/mod.rs
heinwol: yes, i'll do this in presumably next pr

heinwol:
First, the spectrum of tasks we discussed here is presumably done. Probably some fixes are required still.

Second, rebased. Took a brief look at changes between previous and rebased versions. Seems like there are no semantic conflicts

heinwol:
CI (or, in general, runtime) failures are due to "master"/"main" mismatch in some parts of code (collector/lib.rs and several files in database) due to https://github.com/rust-lang/rustc-perf/pull/2495. Maybe I could make a pr with fixes?
cc @Kobzol

Kobzol:
The CI failures are caused by this PR, not by the one you linked, we have merged a bunch of other PRs in the meantime. It is caused by calling `version_supports_parallel_frontend` even for unreleased compiler artifacts, which won't work.

It is caused by this commit: https://github.com/rust-lang/rustc-perf/pull/2491/changes/887c27bbf9dd5346f40b1d27b32ae98ed22274c8 I don't understand what it is supposed to do, I think we should remove that commit and keep the previous behavior.

heinwol:
Ah, I get it, we can't get information about whether parallel frontend is supported for an artifact with commit sha. Then we have no way to pass information about `-Zthreads` support to rustc-fake. You were making this point before, i didn't understand it then.

At the same time, we can still run correctly on any _published_ artifact. i.e. when we run benches from `bench_published_artifact`. Instead of always passing just `Vec<FrontendThreads>` to `run_benchmarks` we can also  add a flag for whether we have `-Zthreads` support or not, so that `run_benchmarks` can delegate this information properly to rustc-fake. This way we won't be calling `version_supports_parallel_frontend` outside of `bench_published_artifact`.

So the only time we'll fail is when we bench a legacy non-published artifact, which i suppose is a rare situation.

I'll push a commit for that and then ask you to decide whether it's good or we should just stop thinking about such corner cases and move on

Kobzol:
Don't worry about the legacy toolchains, we don't have to deal with them for this PR ;)

heinwol:
Well, it turned out this was harder than i expected. And the change poisoned the entire crate with `FrontendThreadsOrLegacy`/`FrontendThreadsCountsOrLegacy`... But at least I had some fun!

Kobzol:
This is just needlessly complicating the benchmark parameters. Just keep the `Vec` of counts as it was before. We really don't need to support the released benchmarks here, we have to run them manually maybe once every 5 years, and it will always require a lot of other manual one-time changes to the codebase anyway, so there's no point in complicating the collector for it.

Kobzol:
Thank you, it looks great! Can you please squash the commits? Then I'll merge it.

heinwol:
Sure! Into a single one?

Kobzol:
Yeah, I think that would be fine.

heinwol:
ahem... Give me a minute to sort this out ;)

Kobzol (approved):
Thank you very much! Let's see what happens on production :)

heinwol:
fingers crossed ;)

in the meantime i'll start integrating database query changes and site (presumably in one go? Or maybe not)

Kobzol:
I think that website will require changes to the compile-time benchmark queries, so that can be done in one go.

Issue #1599, opened 2023-05-30
triage oddity: why did same PR show up in two consecutive weeks


pnkfelix:
2023-05-23 triage used this url: https://perf.rust-lang.org/?start=3ea9ad532474343426e564b997891e459cda89a6&end=cda5becc27cbc7106646fbc40aacea5e7896d954&absolute=false&stat=instructions%3Au

2023-05-30 triage used this url: https://perf.rust-lang.org/?start=cda5becc27cbc7106646fbc40aacea5e7896d954&end=1221e43bdf413f7c405e9b17ef19d76c88222098&absolute=false&stat=instructions%3Au

why did PR https://github.com/rust-lang/rust/pull/111807 show up in both triage runs? The end of the first is equal to the start of the second, so one would think that would imply that the PR would fall into at most *one* of the two ranges, not both.

PR #2492, opened 2026-07-06
Add perf triage for 2026-07-06

panstromek:


triage/2026/2026-07-06.md
```diff
@@ -0,0 +1,104 @@
...
+
+Regression matches pre-merge run, justified by https://github.com/rust-lang/rust/pull/158035#issuecomment-4833952569:
+"It looks acceptable now, unused-warnings is a stress benchmark with thousands of imports." (@petrochenkov)
+
+This is also a stepping stone towards future performance improvements (parallel compiler). IIUC the primary regressions were mostly mitigated by https://github.com/rust-lang/rust/pull/158604, mentioned below.
+
+#### Improvements
+
+Use `LowerAvailableDepth::No` for normalizes-to goal itself instead of its nested goals [#158705](https://github.com/rust-lang/rust/pull/158705) [(Comparison Link)](https://perf.rust-lang.org/compare.html?start=3659db0d3e2cd634c766fcda79ed118eca31a9fd&end=3c00c96d3af4d5b5e101e56cc161a608b21366ee&stat=instructions:u)
+
```
panstromek, line 43: There's something weird going on. Base commit of this PR is the end commit of the summary range. It looks like summary range "forgets" the last commit during generation. Even if I specify it explicitly in the url, it will silently only include the one before.

I think this is the same issue as https://github.com/rust-lang/rustc-perf/issues/1599. I think that that bug will  happen if the last commit in range has significant changes, so next triage will reproduce it, I believe.

I suspect we do something suspicious when iterating over commits, but I tried to stare at the main loop in `handle_triage` for an hour and I couldn't figure out whether it's right or not, it is quite confusing 😄

panstromek: Opened an issue for this https://github.com/rust-lang/rustc-perf/issues/2493

triage/2026/2026-07-06.md
```diff
@@ -0,0 +1,104 @@
...
+|:----------------------------------:|:------:|:---------------:|:-----:|
+| Regressions ❌ <br /> (primary)    | -      | -               | 0     |
+| Regressions ❌ <br /> (secondary)  | 441.1% | [0.2%, 1116.0%] | 7     |
+| Improvements ✅ <br /> (primary)   | -0.3%  | [-0.3%, -0.2%]  | 2     |
+| Improvements ✅ <br /> (secondary) | -0.6%  | [-1.2%, -0.3%]  | 6     |
+| All ❌✅ (primary)                 | -0.3%  | [-0.3%, -0.2%]  | 2     |
+
+Regressions caused by https://github.com/rust-lang/rust/pull/158606#issuecomment-4890167477
+
+Regressions are limited to new-solver, which is not enabled by default. I think we can mark this as triaged and let the types team folluwup on this as they see fit.
```
nnethercote:
```suggestion
Regressions are limited to new-solver, which is not enabled by default. I think we can mark this as triaged and let the types team follow up on this as they see fit.
```

nnethercote (approved):

PR #2494, opened 2026-07-07
database: replace postgres-native-tls with tokio-postgres-rustls

djc:
As suggested in

- https://github.com/rust-lang/rustc-perf/pull/2475#pullrequestreview-4640020616

-10/+1 dependencies seems nice enough, and no more OpenSSL, as you suggested.

This is untested, since I don't have credentials to RDS -- but seems likely to work?

(Might be nice to also migrate this repo from `master` -> `main`?)

r? @Kobzol

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -18,14 +18,16 @@ use crate::{
 use anyhow::Context as _;
 use chrono::{DateTime, TimeZone, Utc};
 use hashbrown::{HashMap, HashSet};
-use native_tls::{Certificate, TlsConnector};
-use postgres_native_tls::MakeTlsConnector;
+use rustls::pki_types::pem::PemObject;
+use rustls::pki_types::CertificateDer;
+use rustls::{ClientConfig, RootCertStore};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::Mutex;
 use tokio_postgres::Statement;
 use tokio_postgres::{GenericClient, Row};
+use tokio_postgres_rustls::MakeRustlsConnect;
 
 pub struct Postgres(String, std::sync::Once);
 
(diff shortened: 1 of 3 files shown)


database/src/pool/postgres.rs
```diff
@@ -77,32 +77,32 @@ pub async fn make_client(db_url: &str) -> anyhow::Result<tokio_postgres::Client>
...
-    use x509_cert::der::pem::LineEnding;
-    use x509_cert::der::EncodePem;
 
-    static CERTIFICATE_PEMS: Mutex<Option<Vec<u8>>> = Mutex::const_new(None);
+async fn make_certificates() -> Arc<RootCertStore> {
+    static ROOTS: OnceLock<Arc<RootCertStore>> = OnceLock::new();
 
-    let mut guard = CERTIFICATE_PEMS.lock().await;
-    if guard.is_none() {
+    if ROOTS.get().is_none() {
```
Kobzol: This should be using https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html#method.get_or_init, but we can't do that because of an `await`. There could be a race condition if implemented like this though, so I'd suggest using the previous solution with the async Mutex.

Kobzol (commented):
Thank you! I did some light experiments locally, that seemed to work, but a full test will have to happen on production.

Left one comment.

database/src/pool/postgres.rs
djc: `Mutex` feels like overkill for this -- how about this solution with an async `RwLock`?

djc: Alternatively, could use a tokio `SetOnce`, which seems like it might be a more idiomatic solution for this.

Kobzol: Unless I missed it, `SetOnce` doesn't have an atomic init method :thinking: We don't need this to be super fast, just atomic, which the Mutex did in a predictable way. With the `RwLock`, two threads could download the certificates concurrently at the start of the process, in theory.

djc: Back to the `Mutex` it is. Maybe I'll try a `SetOnce` PR after this is merged.

Kobzol (approved):
Thank you! :) Let's try this on production 😅

(btw I changed the default branch to `main` in the meantime, thanks for the reminder)

PR #2495, opened 2026-07-07
Change default branch to `main`

Kobzol:


lqd (approved):
This change shouldn't impact the r-l/r submodule, right?

Kobzol:
I seem to remember that I avoided changing the branch of this repo for some time, but I think that was only because of some code in the collector that hardcoded `master`. But I changed that a long time ago to `HEAD`, and checked now that it should be fine. Submodules shouldn't be affected, yeah.

PR #2496, opened 2026-07-07
Remove OpenSSL from Dockerfile

Kobzol:
Let's see if this works :)


Kobzol:
Looks like it does, yay.

lqd (approved):

Issue #1932, opened 2024-06-26
Add historical charts for compiler artifact sizes


Kobzol:
Would be nice to help investigating issues like [this](https://github.com/rust-lang/rust/pull/126965#issuecomment-2191101137).

Kobzol:
Implemented in https://github.com/rust-lang/rustc-perf/pull/2501 and https://github.com/rust-lang/rustc-perf/pull/2505.

PR #2501, opened 2026-07-15
Add artifact size history chart to toolchain page

Kobzol:
This PR renames the `bootstrap` page to `toolchain` and adds historical toolchain artifact size chart to it.

image

Related issue: https://github.com/rust-lang/rustc-perf/issues/1932 (the issue isn't solved by this PR yet, because we also want to add historical artifact size charts to the compage page).


teor2345:
Is there something wrong with the graph data or bounds, or is it intended to only have 2 data points per line?

site/src/server.rs
```diff
@@ -492,8 +492,10 @@ async fn handle_fs_path(req: &Request, path: &str, use_compression: bool) -> Opt
     let relative_path = path.trim_start_matches('/');
     let source = match path {
         "" | "/" | "/index.html" => resolve_template("graphs.html").await,
-        "/bootstrap.html"
-        | "/compare.html"
+        // /bootstrap.html is kept for backwards compatibility, the URL was renamed to
+        // /toolchain.html
+        "/toolchain.html" | "/bootstrap.html" => resolve_template("toolchain.html").await,
```
panstromek, line 497: We could also redirect here (maybe if we're concerned about SEO 😆)

Kobzol:
It's a screenshot from my local DB, which had two toolchains at the time.

site/frontend/src/pages/toolchain/page.vue
```diff
@@ -67,7 +67,8 @@ loadToolchainData(selector, loading);
       See <a href="/compare.html">compare page</a> for descriptions of what the
       names mean.
     </div>
-    <div id="byCrateChart"></div>
-    <div id="totalChart"></div>
+    <div id="artifactSizeChart"></div>
+    <div id="bootstrapByCrateChart"></div>
+    <div id="bootstrapTotalChart"></div>
```
panstromek: Note: Since these elements are used by external Vue library, it's good to make sure that Vue doesn't mess around with them during re-renders e.g. by replacing them or by deleting their contents.

Vue's built-in way to do this is `v-once` directive, but here there's also a problem that they are inside `v-else` branch. Ideally, you want to have these kinds of elements in part of the template that's always rendered - either outside of the `v-else` here, or  a common pattern is to have a wrapper component that contains only a single `v-once` element and have the `renderGraph` call in that component code. This way you ensure that the graph is always rendered when that component is rendered, no matter where is it in the parent component tree.

btw. the `await nextTick()` call above is a bit of a smell, it's a symptom of this kind of a problem.

panstromek (commented):
Hey, I went through this, because I'm trying to get more involved, so think I should also contribute some review time in general to increase our review capacity, but I'm not familiar with the website codebase enough to have 100% confidence I can approve it.

I have a lot of experience with Vue and SPAs, so I can offer my expertise there, but I don't know our DB setup and backend code enough to judge that part well.

---

In general I think it looks good, with just a meta point that I think it's a bit healthier to have more of the data-processing code on the backend.

e.g. I'd try pre-prepare more of the data for `renderPlots` in BE and send down just a generic data structure and keep `renderPlots` just for mapping that data to uPlot-specific properties. In other words, `renderPlots` would not mention anything like `artifact`.

We don't have a gummy typed language on the backend, and Rust is really good at this (and also better known in the Rust project), but the bigger reason to do this is that if you want to show a very similar looking page for some different data, it requires you to rewrite both frontend and backend, when ideally you would want to just use the same page and point it to a different data route.

It's not a blocking concern, I don't think we need to change this now, just a philosphical direction I'd suggest into the future.

I hope this is useful at least a bit ;)

site/src/server.rs
Kobzol: I don't think that SEO is really important for this website :laughing: In fact I think that even keeping the legacy URL is kinda useless.

site/frontend/src/pages/toolchain/page.vue
Kobzol: Just to provide a bit more context, this page used to be written without Vue at all (and in JS, not TS), and then it was "ported" to Vue and to TypeScript, just so that we have a bit more unification. So if it was created from scratch, it would probably look quite different, and maybe we'd also even use a different charting library (though this one was selected mostly for its performance, IIRC).

I don't think I ever saw actual issues with this in practice on this page, but it seems like a good thing to do regardless. I will move the charts below the if/else condition. Is `v-once` needed also for HTML elements (like the chart `div`s) that don't contain any dynamic Vue variables?

Kobzol:
Thanks a lot for taking a look! I definitely agree that doing data processing in TS is a bit PITA, and there are already a few places in our endpoints where we moved the processing to Rust to make it more pleasant. I think that in theory we could do that here, but the devil is in the details. I don't think that returning fully generic data from the backend is a good idea, because the frontend always has to do a lot of finishing touches - setting labels, font sizes, axis margins, value normalizations, etc., and it doesn't make a lot of sense, I think, to prepare all of that in the backend. It is the duty of the frontend to know what data it is dealing with and how to visualize them. In this case, calculating the artifact totals could be on the backend for sure, but I don't see much more computation that it might make sense to move there.

site/frontend/src/pages/toolchain/page.vue
panstromek: > it was "ported" to Vue 

Yea, I figured. It definitely looks like it. It feels very imperative, and mixes the reactivity system with manual updates.

> I don't think I ever saw actual issues with this in practice on this page

Yes, I looked through the code and didn't see an issue right now, but it would become an issue if we start flipping the `loading` state after the initial load (to load more data in place for exaple), or if we add more stuff into the template, nest this into another div, or add a `key` to some parent element. In other words, it's not a problem now, but it's a bit fragile, it could break in weird ways with seemingly unrelated changes in the future.

> Is v-once needed also for HTML elements (like the chart divs)

No, `v-once` applies to the whole sub-tree and it only exist in Vue's internal representation, so it doesn't make sense to add it to HTML elements if they are managed by the charting library and not by Vue.

> I will move the charts below the if/else condition

Good. Just for completeness - if you move it out, I think `v-once` should mostly be a no-op, because Vue doesn't recreate those elements anyway, but I don't think they guarantee this. I think Vue is allowed to assume that the specific element identity doesn't matter and it can always recreate it from scratch if it wants to (maybe as a part of some other optimization).

site/src/server.rs
panstromek: I'll appreciate it, I usually go through this page, because it's fastest to type for autocomplete :smile:

Either way, people sometimes link perf page from blog posts or in chat, so I think it's nice to keep it working

panstromek:
> I don't think that returning fully generic data from the backend is a good idea, because the frontend always has to do a lot of finishing touches - setting labels, font sizes, axis margins, value normalizations, etc., and it doesn't make a lot of sense, I think, to prepare all of that in the backend.)

Yea, generic is a loaded word, anything style related should definitely be in FE code (fonts, margins, etc.). The balance is pretty tricky and somewhat domain specific.

For example of what I mean, instead of sending:

```typescript
artifact_sizes: Dict<[number | null]>`
```
I'd  send something like:

```typescript
data: { label: string,  kind: 'size' | 'time', values: (number | null)[] }
```

This way, the FE knows it's displaying size and can format it nicely, but it's agnostic over size of what. And then, it's up for debate whether the FE should also understand something like `kind: 'commit'` or just treat it as a generic string, and other stuff like that.

The danger is that if you go too far down this route, you'll reimplement some domain specific HTML in JSON :skull:. But too far down to the other side means the frontend codebase explodes and a ton of concerns have to be handled on both frontend and backend for each new kind of data, so there's a balance.

---

Either way, I think this is fine for what we do, I wouldn't worry about it very much here. Definitely don't burn time on refactoring this right now :smile:

site/frontend/src/pages/toolchain/plots.ts
```diff
@@ -179,12 +197,62 @@ export function renderPlots(data: BootstrapData, selector: BootstrapSelector) {
...
+    height: window.innerHeight * 0.4,
+    yAxisLabel: "",
+    series: artifactSizeSeriesOpts,
+    commits: data.commits,
+    kind: "size",
+  });
+
+  new uPlot(
+    artifactSizePlotOpts,
+    artifactSizePlotData as any as TypedArray[],
```
panstromek: Oh, also I just noticed this. This is definitely not `TypedArray`, typed arrays are special JS data structures backed by byte buffers. This TypeScript so it's whatever 😆, but it is slightly misleading.

Kobzol: Changed it to `AlignedData`, which is the actual type that uPlot expects, thanks.

panstromek (approved):
The last commit looks good, I have nothing to add, but I got the review request notification, so  I'm approving this formally, just to make that clear.

(with the caveat mentioned initially. I'll leave it up to you to judge whether somebody else should look at this, too).

Kobzol:
Thanks! This repo doesn't really get a lot of review(ers), so unless the change is something critical, I rather merge it eagerly, we can always revert if something breakas.

PR #2502, opened 2026-07-15
Properly set rustc-fake extension on Windows

Kobzol:
Another fix for https://github.com/rust-lang/rust/pull/159091.


lqd (approved):
something fishier than this is going on, but in the meantime it should work and unblock the rustdoc PGO work

PR #2508, opened 2026-07-22
Download Clippy when a Clippy profile is requested

Kobzol:
Now I know why the Clippy benchmarks didn't work; I never implemented support for actually downloading the Clippy part of the toolchain, lol.

### DIFF
--- a/collector/src/toolchain.rs
+++ b/collector/src/toolchain.rs
@@ -41,6 +41,7 @@ impl Sysroot {
         cache_directory: &Path,
         sha: String,
         triple: &str,
+        profile: Profile,
         backends: &[CodegenBackend],
     ) -> Result<Self, SysrootDownloadError> {
         // The structure of this directory is load-bearing.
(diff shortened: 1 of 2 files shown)


collector/src/toolchain.rs
```diff
@@ -280,12 +293,13 @@ impl SysrootDownload {
         let mut archive = Archive::new(reader);
         let prefix = match component {
             Component::Std => format!("rust-std-{}", self.triple),
-            Component::Cranelift => format!("{component}-preview"),
-            _ => component.to_string(),
+            Component::Clippy | Component::Cranelift => format!("{component}-preview"),
```
lqd, line 296: is the clippy component actually named clippy-preview?

Kobzol: It is indeed, I tried it locally. I think that rustup ignores the preview part, but we can't get rid of it, or something. The whole preview thing is pretty broken (but also Hyrumized at this point).

lqd (approved):

PR #2509, opened 2026-07-26
feat: support `@argfile` for rustc-fake

weihanglo:
This PR includes

* A verbatim copy of `arg_file_command.rs` from rust-lang/rust
* adapted copies of some other helpers from bootstrap.

This found because we tried to enable new build-dir layout for nightly
in Cargo, but failed with

```
error: Unrecognized option: 'skip-this-rustc'
```

I _guess_ this was the root cause.

See

* https://triage.rust-lang.org/gha-logs/rust-lang/rust/89681531404
* https://github.com/rust-lang/rust/pull/159857#issuecomment-5078882351

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -1,9 +1,21 @@
 use std::env;
 use std::ffi::OsString;
 use std::fs;
-use std::path::PathBuf;
+use std::io::BufRead;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+// Vendored from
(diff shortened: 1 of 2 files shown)


ranger-ross (approved):

collector/src/bin/rustc-fake.rs
```diff
@@ -93,6 +104,9 @@ fn main() {
 
         benchlib::process::raise_process_priority();
 
+        let expanded_args = args;
+        let (args, _argfile) = pack_args(&tool, expanded_args.clone());
```
Kobzol: This is a bit confusnig, because the `tool` argument is essentially unused. Let's use some placeholder command inside `pack_args` and avoid passing the tool in.

Kobzol (commented):
Seems reasonable, thank you. Left one comment.

collector/src/bin/rustc-fake.rs
weihanglo: Nice catch. Fixed!

Kobzol (approved):
Thanks!

PR #2510, opened 2026-07-26
Stop passing -Zincremental-verify-ich

xmakro:
rustc-perf forces `-Zincremental-verify-ich` so that incremental cache verification, which normally covers a random 1/32 of loaded values, does not add noise: the subset was picked by value hash, so it differed between the two builds being compared.

Since [rust-lang/rust#160130](https://github.com/rust-lang/rust/pull/160130) the subset is picked by key fingerprint and session count. Both are the same for the two builds in a comparison, because rustc-perf [pins the rustc version](https://github.com/rust-lang/rustc-perf/blob/37f6778aae91f649efac6f02852621a17d864f31/collector/src/bin/rustc-fake.rs#L28) and runs the same scenarios, so the verification cost is identical on both sides and no longer noise.

We also want to build on this in a follow-up to [rust-lang/rust#160214](https://github.com/rust-lang/rust/pull/160214), where we want to avoid loading query results that are already green. Those values would then not be available for verification at all, so for that work it is desirable that rustc-perf measures exactly what users will run.

### DIFF
--- a/collector/src/bin/rustc-fake.rs
+++ b/collector/src/bin/rustc-fake.rs
@@ -82,7 +82,16 @@ fn main() {
     // avoid. rustc-perf can accept the higher cost of always verifying hashes,
     // and we currently prefer to avoid exposing a means of hard-disabling
     // verification.
-    args.push(OsString::from("-Zincremental-verify-ich"));
+    //
+    // Only pass this for incremental compiles. For non-incremental compiles
+    // the flag only triggers a costly query key verification sweep.
+    let is_incremental = args.iter().any(|arg| {
+        arg.to_str()
+            .is_some_and(|a| a.starts_with("incremental=") || a.starts_with("-Cincremental="))
+    });
+    if is_incremental {
+        args.push(OsString::from("-Zincremental-verify-ich"));
+    }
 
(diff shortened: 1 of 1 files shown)


nnethercote:
Seems ok to me, though we'll need to be careful when we update the rustc-perf version used by CI, otherwise the improvement might be attributed to a compiler change.

Kobzol:
This shouldn't be needed if https://github.com/rust-lang/rust/pull/159796 lands, right? Let's wait for that.

xmakro:
I updated the PR to drop the flag entirely instead of only skipping it for non-incremental compiles. This is now possible because https://github.com/rust-lang/rust/issues/160130 landed.

The concrete motivation is a follow-up to [rust-lang/rust#160214](https://github.com/rust-lang/rust/issues/160214) that avoids loading query results that are already green. Those values would not be available for verification at all, so forcing full verification would not show the perf saving of that work.

I think it is desirable that perf CI runs the same setting that real users would run, so we  measure the right thing.

Kobzol:
So if I understand it correctly, after https://github.com/rust-lang/rust/pull/160130, the verification is based on the hash of the key, not on the hash of the value, so you're saying that the keys should be the same in both the compared runs, and we thus don't have to verify everything anymore?

But aren't some of the keys based on values from the compilation session?

PR #2513, opened 2026-07-29
Run DocJson benchmarks on CI

Kobzol:
And also rebalance the compilation benchmarks to better split CI time.


.github/workflows/ci.yml
lqd: Docjson too ig

lqd (approved):

PR #2516, opened 2026-07-30
Add `doc-json` filter to the compare page

Kobzol:
Useful now that we run `doc-json` benchmarks by default.


lqd (approved):
I haven't tested this locally but it looks good to me.

PR #2517, opened 2026-07-30
Add `*.map` files as static assets to the site.

heinwol:
It could help debugging frontend issues (helping me a lot now).

While this could theoretically increase size or provide some excessive information (might possibly broaden attack surface? I don't really know), I think we could tell parcel not to create map files and therefore they wouldn't be bundeled. Therefore, it could be said there's no need to hide this behind an optional feature

### DIFF
--- a/site/src/resources.rs
+++ b/site/src/resources.rs
@@ -24,6 +24,8 @@ struct StaticAssets;
 #[include = "*.js"]
 #[include = "*.br"]
 #[include = "*.css"]
+// Source maps for debugging; only in debug mode since it increases binary size by ~6%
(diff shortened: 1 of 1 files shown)


Kobzol:
The source code is public, and we build the `.map` files on CI, so I don't think that security is an issue here.

But this increases the size of the release `site` binary from 117 MiB to 124 MiB, which is not trivial (though also not a big issue). That being said, I think that gating this behind `debug_assertions` (using `cfg_attr`) might be better (and add a comment that explains why we do it).

Kobzol (approved):
Thanks!

PR #2522, opened 2026-08-07
Create three NLL benchmarks

Kobzol:
Polonius is now used on nightly by default, and its performance is quite worse on some crates than with NLL. This PR creates three secondary copies of existing benchmarks that use NLL instead of Polonius, so that we can track NLL performance until Polonius becomes stabilized. The benchmarks are `clap_derive`, `ripgrep` and `serde`. I tried to spread them across the existing collectors to balance benchmark time.


lqd (approved):
Great, thank you!

Do we still want to add them to https://github.com/rust-lang/rustc-perf/blob/main/collector/compile-benchmarks/README.md?

Kobzol:
I don't think so, they are more or less synthetic temporary benchmarks. We didn't add the `-new-solver` variants there either.

PR #2524, opened 2026-08-07
Remove unrolling functionality

Kobzol:
The bors implementation seems to be [up-to-speed](https://github.com/rust-lang/rust/pull/161134#issuecomment-5304267942), so this should be ready.


lqd (approved):
🚀

PR #2525, opened 2026-08-07
Create a `parse_command` function to parse rust-timer commands

JonathanBrouwer:
This separates the parsing code from the logic and makes the tests cover more of the parsing logic.
This is the last refactor to this file for now before I will add new things to it.

I would again recommend enabling "Hide whitespace" while reviewing

r? @Kobzol

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -61,6 +61,8 @@ async fn handle_push(ctxt: Arc<SiteCtxt>, push: github::Push) -> ServerResult<gi
     Ok(github::Response)
 }
 
+const RUST_TIMER_PREFIX: &str = "@rust-timer";
+
 async fn handle_issue(
     ctxt: Arc<SiteCtxt>,
     issue: github::Issue,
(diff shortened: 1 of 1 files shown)


JonathanBrouwer:
^ Rebased on top of the force pushed main branch to solve conflicts

site/src/request_handlers/github.rs
```diff
@@ -474,171 +471,163 @@ mod tests {
     use super::*;
 
     #[test]
-    fn build_command_missing() {
-        assert!(parse_build_command("").is_none());
+    fn command_missing() {
+        insta::assert_compact_debug_snapshot!(parse_command(""),
+            @r#"Err("Cannot find @rust-timer command even though `@rust-timer` is tagged")"#);
```
Kobzol, line 476: This error message is misleading, because the test code doesn't have the if condition that gets executed on the live service, but that doesn't matter much.

Kobzol (approved):
Thank you, this is much more readable than before!

PR #2528, opened 2026-08-10
fix operator in renovate configuration

marcoieni:
there is an error in the renovate dashboard:
image

You can see from the [docs](https://docs.renovatebot.com/configuration-options/#packagerulesmatchjsonata) that we should use `=` here for equality.


Kobzol:
I think you'll need to rebase again.

Kobzol (approved):
Thanks!

PR #2531, opened 2026-08-14
More refactoring + Implement step 1 of triage command

JonathanBrouwer:
This PR implements step 1 of the triage command as described in the initial pitch [#t-compiler/performance > Proposal for &#96;@rust-timer triage&#96; command @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/Proposal.20for.20.60.40rust-timer.20triage.60.20command/near/606899059).

It consists of 4 commits, which I recommend to review individually:
1. [Simplify enqueue_sha function](https://github.com/rust-lang/rustc-perf/commit/2b36ac3201a9626b8432c78479f61e115c113c1b) -> Simplifies the `enqueue_sha` by pushing the comment posting logic out of the function & removing to option to provide multiple commits. This PR benefits from "hide whitespace" during review again.
2. [Factor enqueue_sha_build into its own function](https://github.com/rust-lang/rustc-perf/commit/322ada1557b044f1e607a3aa7c63401c6a449d1c) -> Factor part of the `@rust-timer build` command into its own function, which can be re-used by the triage command.
3. [Add command parsing for triage command](https://github.com/rust-lang/rustc-perf/commit/65a83d0354647c380215d0b2495bc44a26908cc4)
4. [Implement triage command](https://github.com/rust-lang/rustc-perf/commit/65ed0ec052d914857301ee1ea29540244b6105f4)

r? @Kobzol

### DIFF
--- a/site/src/github.rs
+++ b/site/src/github.rs
@@ -227,79 +227,63 @@ pub async fn rollup_pr_number(
 }
 
 /// Enqueues the given SHAs and returns the SHAs that were actually enqueued.
-pub async fn enqueue_shas<'a>(
+pub async fn enqueue_sha<'a>(
     ctxt: &SiteCtxt,
     gh_client: &client::Client,
     pr_number: u32,
-    commits: impl Iterator<Item = &'a str>,
-) -> Result<Vec<&'a str>, String> {
-    let mut enqueued = vec![];
-    let mut msg = String::new();
-    for commit in commits {
-        let mut commit_response = gh_client
-            .get_commit(commit)
-            .await
-            .map_err(|e| e.to_string())?;
-        if commit_response.parents.len() != 2 {
-            log::error!(
-                "Bors try commit {} unexpectedly has {} parents.",
-                commit_response.sha,
-                commit_response.parents.len()
-            );
-            continue;
-        }
-        let try_commit = TryCommit {
-            sha: commit_response.sha,
-            parent_sha: commit_response.parents.remove(0).sha,
-        };
-        let conn = ctxt.conn().await;
+    commit_sha: &'a str,
+) -> Result<String, String> {
+    let mut commit = gh_client
+        .get_commit(commit_sha)
+        .await
+        .map_err(|e| e.to_string())?;
+    if commit.parents.len() != 2 {
+        return Err(format!(
+            "Bors try commit {} unexpectedly has {} parents.",
+            commit.sha,
+            commit.parents.len()
+        ));
+    }
+    let try_commit = TryCommit {
+        sha: commit.sha,
+        parent_sha: commit.parents.remove(0).sha,
+    };
+    let conn = ctxt.conn().await;
 
-        let queued = conn.attach_shas_to_try_benchmark_request(
-                pr_number,
-                &try_commit.sha,
-                &try_commit.parent_sha,
-                commit_response.commit.committer.date,
+    let queued = conn.attach_shas_to_try_benchmark_request(
+            pr_number,
+            &try_commit.sha,
+            &try_commit.parent_sha,
+            commit.commit.committer.date,
             )
             .await
             .map_err(|error| format!("Cannot attach SHAs to try benchmark request on PR {pr_number} and SHA {}: {error:?}", try_commit.sha))?;
-        if queued {
-            enqueued.push(commit);
-            if !msg.is_empty() {
-                msg.push('\n');
-            }
+    if !queued {
+        return Err("Commit was not enqueued, since no benchmark request was found".to_string());
+    }
 
-            let (preceding_artifacts, expected_duration) =
-                estimate_queue_info(conn.as_ref(), &try_commit)
-                    .await
-                    .map_err(|e| format!("{e:?}"))?;
+    let (preceding_artifacts, expected_duration) = estimate_queue_info(conn.as_ref(), &try_commit)
+        .await
+        .map_err(|e| format!("{e:?}"))?;
 
-            let verb = if preceding_artifacts == 1 {
-                "is"
-            } else {
-                "are"
-            };
-            let suffix = if preceding_artifacts == 1 { "" } else { "s" };
-            let queue_msg = format!(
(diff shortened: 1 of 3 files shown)


site/src/github.rs
```diff
@@ -227,79 +227,63 @@ pub async fn rollup_pr_number(
 }
 
 /// Enqueues the given SHAs and returns the SHAs that were actually enqueued.
```
Kobzol:
```suggestion
/// Enqueues the given SHA and returns a message that should be sent as a comment to the corresponding PR.
```

site/src/github.rs
```diff
@@ -227,79 +227,63 @@ pub async fn rollup_pr_number(
...
-
-            msg.push_str(&format!(
-                "Queued {} with parent {}, future [comparison URL]({}).\n{queue_msg}",
-                try_commit.sha,
-                try_commit.parent_sha,
-                try_commit.comparison_url(),
-            ));
-        }
+    if !queued {
+        return Err("Commit was not enqueued, since no benchmark request was found".to_string());
```
Kobzol:
```suggestion
        return Err("Commit was not enqueued, since no previous benchmark request was found".to_string());
```

site/src/request_handlers/github.rs
```diff
@@ -237,61 +240,114 @@ async fn handle_rust_timer(
...
+    main_client: &Client,
+    issue_number: u32,
+    cmd: &BuildCommand<'_>,
+) -> Result<String, String> {
+    // requested artifacts do not exist errors
+    validate_build_command(cmd).await?;
+
+    {
+        let conn = ctxt.conn().await;
+        record_try_benchmark_request_without_artifacts(
```
Kobzol, line 338: Note (pre-existing): this is a bit sketchy, we record a request without artifacts, and then immediately attach the artifacts. We could just do it in one step for `build`. Though I suspect that would actually make the surrounding code more complicated, so it's not worth it.

site/src/request_handlers/github.rs
```diff
@@ -336,6 +393,18 @@ fn parse_build_command_args(args: &str) -> Result<BuildCommand<'_>, String> {
...
 }
 
+/// Parses the arguments of `@rust-timer triage <sha>+`
+fn parse_triage_command_args(args: &str) -> Result<TriageCommand<'_>, String> {
+    let shas = args
+        .split_whitespace()
+        .map(parse_sha)
+        .collect::<Result<Vec<_>, _>>()?;
+    if shas.is_empty() {
+        return Err("The triage comment requires a list of SHAs as an argument.".to_string());
```
```suggestion
        return Err("The triage comment requires a space-separated list of SHAs as an argument.".to_string());
```
Though we use comma-separated arguments for other commands (usually also for other bots), so I think that would be more consistent.

site/src/request_handlers/github.rs
```diff
@@ -430,6 +499,7 @@ fn parse_command_arguments(args: &str) -> Result<HashMap<&str, &str>, String> {
 enum RustTimerCommand<'a> {
     Queue(QueueCommand<'a>),
     Build(BuildCommand<'a>),
+    Triage(TriageCommand<'a>),
```
Kobzol, line 511: Could you please add doc comments here to explain what do these individual commands do? Thanks!

site/src/request_handlers/github.rs
```diff
@@ -237,61 +240,114 @@ async fn handle_rust_timer(
...
         }
     }
 
     Ok(github::Response)
 }
 
+fn find_pr_number_for_unrolled_build(commit_message: &str) -> Result<u32, String> {
+    // Find PR number for this unrolled build sha
+    const COMMIT_NAME_START: &str = "Unrolled build for #";
+    let first_line = commit_message.lines().next().unwrap();
```
Kobzol: Could you please handle this unwrap? It's unlikely, but I don't want an empty commit message to bring down the bot 😅 Especially since people can provide their own SHA, so they could point it to an empty commit, in theory.

site/src/request_handlers/github.rs
```diff
@@ -444,6 +514,11 @@ struct BuildCommand<'a> {
 }
 
 #[derive(Debug)]
+struct TriageCommand<'a> {
+    shas: Vec<&'a str>,
+}
+
+#[derive(Debug, Default)]
```
Kobzol, line 530: Why was this needed? 🤔

Kobzol (commented):
Thank you! Could you please also document the new command in the website, at https://perf.rust-lang.org/help.html?

site/src/request_handlers/github.rs
JonathanBrouwer: Because of this line https://github.com/rust-lang/rustc-perf/pull/2531/changes#diff-28b7b28ff9e7ea6e743602d351d074d79de4eeeee423dfdee2fcb921f35a352dR293

Kobzol: Oh, didn't notice that. Thanks!

site/src/request_handlers/github.rs
JonathanBrouwer: I can take a look in a follow-up PR if this is worth it, but I think indeed it would just make everything more complicated

site/src/request_handlers/github.rs
JonathanBrouwer: Done. I kept these descriptions a bit shorter than the `help.html` descriptions in order to not completely duplicate everything.

site/src/request_handlers/github.rs
JonathanBrouwer: Fixed, I assumed `lines` would always emit an item even for an empty string, my bad.

JonathanBrouwer:
@Kobzol I think I handled all the review comments

site/src/request_handlers/github.rs
JonathanBrouwer: Applied the suggestion. I prefer space as they're a bit easier to type, and it feels more like a list of independent arguments than a comma-separated arguments list. Happy to change it tho if you insist

Kobzol: Ok, feel free to keep spaces.

Kobzol (approved):
Thanks, let's try this!

PR #2532, opened 2026-08-14
Fix "detailed results" button for doc-json benchmarks

JonathanBrouwer:
Clicking "detailed results" on one of the doc-json benchmarks, for example [in this run](https://perf.rust-lang.org/compare.html?start=93c9086fdd5b80d286480a19ac047746ecc5fa1f&end=059bf4a660ddea5bd8302ecdd4f7e40dd7a04313&stat=instructions:u), returns an error:
```
invalid profile: "json is not a profile"
```

r? @Kobzol

### DIFF
--- a/site/frontend/src/pages/detailed-query/page.vue
+++ b/site/frontend/src/pages/detailed-query/page.vue
@@ -197,7 +197,12 @@ async function loadData() {
   // To maintain backwards compatibility of URLs, parse the profile from the
   // benchmark URL parameter
   // We turn <benchmark>-<profile> into two separate parameters for the backend
-  const parts = benchmark.split("-");
+  let parts: string[];
+  if (benchmark.endsWith("-doc-json")) {
+    parts = [benchmark.slice(0, -"-doc-json"), "doc-json"];
+  } else {
+    parts = benchmark.split("-");
+  }
   const profile = parts[parts.length - 1];
   const benchmarkSeparate = parts.slice(0, parts.length - 1).join("-");
 
--- a/site/frontend/src/pages/detailed-query/page.vue
+++ b/site/frontend/src/pages/detailed-query/page.vue
@@ -199,7 +199,7 @@ async function loadData() {
   // We turn <benchmark>-<profile> into two separate parameters for the backend
   let parts: string[];
   if (benchmark.endsWith("-doc-json")) {
-    parts = [benchmark.slice(0, -"-doc-json"), "doc-json"];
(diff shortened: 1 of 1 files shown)


site/frontend/src/pages/detailed-query/page.vue
```diff
@@ -197,7 +197,12 @@ async function loadData() {
   // To maintain backwards compatibility of URLs, parse the profile from the
   // benchmark URL parameter
   // We turn <benchmark>-<profile> into two separate parameters for the backend
-  const parts = benchmark.split("-");
+  let parts: string[];
+  if (benchmark.endsWith("-doc-json")) {
```
JonathanBrouwer, line 201: I'm not happy with this code, but not experienced enough in TS to know if there's a better way

Kobzol: It seems weird to use `-` on a string, but it's JavaScript, so I had to check what that does 😆 Probably a typo? The code doesn't seem to work though.

Kobzol: Maybe something like this? To keep it fully general:
```javascript
[...a.slice(0, -"-doc-json".length).split("-"), "doc-json"]
```

JonathanBrouwer: Woops my intended code was adding ` .length` there, typo indeed.
I'm not entirely sure what splitting and then `...` again achieves, could you explain?

JonathanBrouwer: (and sorry for the late reply I forgot about this PR)

Kobzol: It would split benchmark name with a dash, consider:
```javascript
let benchmark = "libc-0.2.172-doc-json";
console.log([benchmark.slice(0, -"-doc-json".length), "doc-json"]); // ['libc-0.2.172', 'doc-json']
console.log([...a.slice(0, -"-doc-json".length).split("-"), "doc-json"]); // ['libc', '0.2.172', 'doc-json']
```
But since we do `parts.slice(0, parts.length - 1).join("-");` below anyway, it shouldn't really matter.

Thanks!

Kobzol (approved):

site/frontend/src/pages/detailed-query/page.vue
JonathanBrouwer: Right, I see, that makes sense but indeed shouldn't matter

PR #2537, opened 2026-08-17
forcefully disable next-solver for all tests

lcnr:
cc https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/next-solver.20by.20default.20PR/near/616986124 @JakobDegen @lqd

I can't test this locally as I can't get it to find my `perf` even though it's installed :sweat_smile: 

Made sure I actually did this for all tests by making sure all folders except the next-solver ones are orange in my file view :sparkles: 

cc https://github.com/rust-lang/compiler-team/issues/1014

### DIFF
--- a/collector/compile-benchmarks/await-call-tree/perf-config.json
+++ b/collector/compile-benchmarks/await-call-tree/perf-config.json
@@ -1,4 +1,5 @@
 {
+    "cargo_rustc_opts": "-Znext-solver=coherence",
     "category": "secondary",
     "artifact": "library"
 }
--- a/collector/compile-benchmarks/bitmaps-3.2.1/perf-config.json
+++ b/collector/compile-benchmarks/bitmaps-3.2.1/perf-config.json
@@ -1,4 +1,5 @@
 {
+  "cargo_rustc_opts": "-Znext-solver=coherence",
   "artifact": "library",
   "category": "primary"
 }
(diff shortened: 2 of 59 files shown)


Kobzol:
Huh, I was thinking of something a bit more involved in the collector's source code, but in the end this is actually the simplest solution, and we should be able to easily revert it before landing the next trait solver on stable.

lqd:
> cc https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/next-solver.20by.20default.20PR/near/616986124 @JakobDegen @lqd

I've never seen both in the same room but Jakob Degen and Jakub Beránek are not the same person 🙃

lqd:
AFAICT the unstable -Z flag shouldn't be added to the benchmarks we run on stable releases only? https://github.com/rust-lang/rustc-perf/tree/main/collector/compile-benchmarks#stable

Kobzol:
It would break on older versions, but we probably won't rebenchmark those soon (maybe only for ARM), and for the new stable versions, we pass RUSTC_BOOTSTRAP. But yeah, stable benchmarks shouldn't get the flag.

lcnr:
> I've never seen both in the same room but Jakob Degen and Jakub Beránek are not the same person 🙃

wooops, in theory I should be aware. I have talked with both of them on separate occasions :sparkles:

lcnr:
removed it for stable benchmarks again, don't fully get whether that's actually good :thinking: https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/Big.20changes.20coming/near/617126098

lcnr:
ah alright, think I get why this makes sense

Kobzol (approved):
Thanks, let's try to see what this does on production.

PR #2540, opened 2026-08-19
Parse new unrolled PR commit format

JonathanBrouwer:
Should be merged together with https://github.com/rust-lang/bors/pull/824

This will break the `@rust-timer triage` command for all previous rollups, but since the triage command is so new I think that's fine? Otherwise I can implement parsing for the old style of message too

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -237,14 +237,15 @@ async fn handle_rust_timer(
                 };
 
                 // Find PR number from commit message
-                let pr_number = match find_pr_number_for_unrolled_build(&commit.commit.message) {
-                    Ok(r) => r,
-                    Err(err) => {
-                        writeln!(&mut result, "### {sha}").unwrap();
-                        writeln!(&mut result, "{err}").unwrap();
-                        continue;
-                    }
-                };
+                let unrolled_build_message =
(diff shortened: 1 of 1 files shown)


site/src/request_handlers/github.rs
```diff
@@ -278,22 +284,48 @@ async fn handle_rust_timer(
...
+}
+
+fn parse_unrolled_build_message(commit_message: &str) -> Result<UnrolledBuildMessage, String> {
     let first_line = commit_message.lines().next().unwrap_or("");
-    let Some(pr_number) = first_line.strip_prefix(COMMIT_NAME_START) else {
+
+    // The first line of the commit message will look like
+    // `Unrolled build for #123 in rollup #123`
+    let words = first_line.split(" ").collect::<Vec<_>>();
+    let ["Unrolled", "build", "for", member_pr_number, "in", "rollup", rollup_pr_number] =
```
Kobzol, line 300: Oh wow. I was wondering how will you parse the title. I did not expect this 😆

Kobzol (approved):
Nah, that's fine.

site/src/request_handlers/github.rs
JonathanBrouwer: Haha it was the easiest way I could think of

PR #2544, opened 2026-08-21
Update rust toolchain version

JonathanBrouwer:
I got annoyed by my local clippy suggesting these changes enough to do something about it
r? @Kobzol


Kobzol (commented):
Does it really need 1.98.0? In that case we should also update https://github.com/rust-lang/rustc-perf/blob/main/Cargo.toml#L7, but doing that a day after that version was released is quite soon.. :)

Kobzol (approved):
Thanks!

PR #2543, opened 2026-08-21
Make comment of triage build link back to the triage invocation

JonathanBrouwer:
r? @Kobzol 

Recommend reviewing commit by commit. 
Again the "did not test this" disclaimer other than what I mentioned in the self-review

PR #2545, opened 2026-08-22
Update the triage comment with results when a benchmark run completes

JonathanBrouwer:
Blocked on https://github.com/rust-lang/rustc-perf/pull/2543

Recommend reviewing commit by commit.

### DIFF
--- a/site/src/github/comparison_summary.rs
+++ b/site/src/github/comparison_summary.rs
@@ -180,12 +180,25 @@ async fn summarize_run(
     let bootstrap = summarize_bootstrap(&inst_comparison);
     let artifact_size = summarize_artifact_size(&inst_comparison);
 
+    write!(&mut message, "{}", metrics_result(ctxt, &commit).await?).unwrap();
+    write!(&mut message, "\n{bootstrap}").unwrap();
+    write!(&mut message, "\n{artifact_size}").unwrap();
+
+    Ok(message)
+}
+
+async fn metrics_result(
+    ctxt: &SiteCtxt,
+    commit: &QueuedCommit,
+) -> Result<String, String> {
+    let benchmark_map = ctxt.get_benchmark_category_map().await;
+    let mut metrics_result = String::new();
     let metrics = vec![
         (
             "Instruction count",
             Metric::InstructionsUser,
             DefaultMetricVisibility::Shown,
-            inst_comparison,
+            calculate_metric_comparison(ctxt, &commit, Metric::InstructionsUser).await?,
         ),
         (
             "Max RSS (memory usage)",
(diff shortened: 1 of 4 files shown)


site/src/github/comparison_summary.rs
```diff
@@ -38,7 +38,25 @@ pub async fn post_comparison_comment(
...
-    } else if let Some(triage_run) = is_triage_run(&commit, &mut client, &mut graph_client).await? {
+    } else if let Some(mut triage_run) =
+        is_triage_run(&commit, &mut client, &mut graph_client).await?
+    {
+        // Update the triage comment
+        let triage_summary = metrics_result(ctxt, &commit)
+            .await
+            .unwrap_or_else(|error| error);
+        if let Err(err) =
+            update_triage_body(&mut triage_run.triage_comment.body, pr, triage_summary)
```
Kobzol: If this fails for some reason, we probably shouldn't try to update the comment?

Also I'd do this whole thing as a best effort, if it fails, it shouldn't cancel posting the comment to the unrolled PR itself.

site/src/github/triage.rs
```diff
@@ -38,3 +39,147 @@ pub async fn is_triage_run(
         triage_comment,
     }))
 }
+
+pub fn update_triage_body(
```
Kobzol, line 50: Can you please add a comment to this function, that it should be kept in sync with a function that generates the triage template string? Maybe those functions could live below one another to make that easier.

Also, maybe it would be easier to wrap the content to be replaced with something like
```
<!-- rust-timer:triage-content-start-pr-123456 -->
Queued commit ...
<!-- rust-timer:triage-content-end-pr-123456 -->
```
And then replace the content between those two comments?

site/src/github/comparison_summary.rs
```diff
@@ -38,7 +38,25 @@ pub async fn post_comparison_comment(
 
     let source = if is_master_commit {
         PerfRunSource::MasterCommit
-    } else if let Some(triage_run) = is_triage_run(&commit, &mut client, &mut graph_client).await? {
+    } else if let Some(mut triage_run) =
+        is_triage_run(&commit, &mut client, &mut graph_client).await?
+    {
+        // Update the triage comment
+        let triage_summary = metrics_result(ctxt, &commit)
```
Kobzol, line 45: Maybe just add a comment here that this is racy, but it shouldn't really ever happen in practice that two perf runs. finish at the same time, so it should be fine.

Kobzol (commented):
This will be super cool :)

site/src/github/triage.rs
JonathanBrouwer: I changed it to generate the labels you mentioned, that is way easier indeed, thanks for the idea!

site/src/github/comparison_summary.rs
JonathanBrouwer: Done

site/src/github/comparison_summary.rs
JonathanBrouwer: I changed the implementation so we don't update the comment if `update_triage_body` fails.
I do update the comment if `metrics_result` failed (and show the error), as I think that may be useful for debugging.
It indeed doesn't cancel posting the comment to the PR with the current implementation :)

Kobzol: The question mark at the end of this:
```
graph_client
                    .update_comment_content(
                        &triage_run.triage_comment.id,
                        &triage_run.triage_comment.body,
                    )
                    .await?;
```
might cancel sending the comment, that's what I meant :)

Kobzol (approved):
Thanks, impl looks good, let's try it on prod!

JonathanBrouwer:
^ Added the [Recover on update_comment_content failure](https://github.com/rust-lang/rustc-perf/pull/2545/commits/4430c9daec79dddb6d25dea6c3f2e6d01916d50b) commit

PR #2547, opened 2026-08-24
Fix triage comment update

JonathanBrouwer:
Fixes:
- There are not enough newlines in the `triage_summary`, causing the title of the next entry to not be rendered (because markdown titles require two newlines before them)
- The triage URL needs to have the comment database ID, not the ID
- Move the triage comment to be just below "Finished benchmarking commit"

r? @Kobzol


Kobzol (approved):
Thanks!

PR #2549, opened 2026-08-26
Replace archived actions-rs/cargo with cargo directly

Sandijigs:
actions-rs/cargo has been archived for a while now, so this swaps it in CI for cargo directly. The toolchain is already installed via rustup earlier in the job, so the build runs the same as before.
This only touches .github/workflows/nightly.yml. The other archived-action references under collector/compile-benchmarks/ are vendored benchmark crates, not rustc-perf's own CI, so they're intentionally left as-is.
This is part of a wider effort to remove archived actions across rust-lang repos, tracked in rust-lang/crabwatch#49.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -45,10 +45,7 @@ jobs:
           npm run build
 
       - name: Compile site
-        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
-        with:
-          command: build
-          args: --release --bin site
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Yeah, this was really unnecessary usage of that action. Thank you!

.github/workflows/nightly.yml
lqd, line 31: This PR makes me see we download and install these two components, but I don't think we use them in this job?!

JonathanBrouwer:
CI failed because of github weirdness, let's retry

PR #2550, opened 2026-08-29
Improve support for ARM benchmarks

Kobzol:
Since we now finally have an ARM machine for benchmarking, I tested having both x64 and ARM data locally, and found some issues, which this PR solves.

This PR:
- Ensures that both GitHub PR comments and perf. triage only considers x64, for now.
- Fixes a pretty nasty bug (https://github.com/rust-lang/rustc-perf/pull/2550/commits/6f2df118f27372ed55f94864796b00bd295b3382). If we got different target/frontend thread data into the DB without this fix, the compare page would start showing random subsets of data (yikes!). Found this out while testing the new self-comparison functionality locally.
- Allow self-comparing target results for the same commit (so that we can compare x64 vs ARM perf.).

It doesn't yet enable running ARM benchmarks by default, because we don't yet build aarch64 artifacts by default in try builds.

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -484,7 +484,7 @@ fn parse_benchmark_parameters<'a>(
         parse_targets(targets).map_err(|e| {
             format!(
                 "Cannot parse targets: {e}. Valid values are: {}",
-                Target::primary_targets()
+                Target::available_targets()
                     .iter()
                     .map(|b| b.as_str())
                     .collect::<Vec<_>>()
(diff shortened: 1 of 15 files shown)


site/src/request_handlers/github.rs
```diff
@@ -728,7 +728,7 @@ Otherwise LGTM."#),
         insta::assert_compact_debug_snapshot!(parse_command("@rust-timer queue targets=x86_64-unknown-linux-gnu"),
             @r#"Ok(Queue(QueueCommand { params: BenchmarkParameters { backends: None, profiles: None, targets: Some("x86_64-unknown-linux-gnu") } }))"#);
         insta::assert_compact_debug_snapshot!(parse_command("@rust-timer queue targets=x86_64-unknown-linux-gnu,67-unknown-none"),
-            @r#"Err("Cannot parse targets: Only primary targets can be specified. Valid values are: x86_64-unknown-linux-gnu, aarch64-unknown-linux-gnu")"#);
+            @r#"Err("Cannot parse targets: Only the available targets `x86_64-unknown-linux-gnu, aarch64-unknown-linux-gnu` can be specified. Valid values are: x86_64-unknown-linux-gnu, aarch64-unknown-linux-gnu")"#);
```
lqd, line 731: The valid values were already specified

lqd (approved):
I haven't tested this locally, but reading it through made sense to me

site/src/request_handlers/github.rs
Kobzol: Ah, right. Well, repetition is the mother of learning, after all :laughing:

Kobzol:
Let's try on production!

PR #2551, opened 2026-08-29
Correctly store targets of benchmark requests into the database

Kobzol:
Turns out that we just weren't storing them at all.. :)

### DIFF
--- a/database/src/pool/postgres.rs
+++ b/database/src/pool/postgres.rs
@@ -1439,6 +1439,16 @@ where
         &self,
         benchmark_request: &BenchmarkRequest,
     ) -> anyhow::Result<BenchmarkRequestInsertResult> {
+        let BenchmarkRequest {
+            commit_type,
+            commit_date,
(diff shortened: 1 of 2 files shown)


database/src/pool/postgres.rs
```diff
@@ -1452,21 +1462,23 @@ where
                     created_at,
                     backends,
                     profiles,
+                    targets,
                     commit_date
```
JonathanBrouwer, line 1466: Nit: It's kinda confusing to me that the order here is different from the order of the struct destructure

JonathanBrouwer (approved):

database/src/pool/postgres.rs
Kobzol: I tried to batch the benchmark parameters together, but yeah, the ordering is a bit off. It also is tricky because the DB column ordering is based on when was the column added, so e.g. `targets` is actually the last column right now.

PR #2552, opened 2026-08-30
Add `benchmarks` parameter to `@rust-timer queue/build`

JonathanBrouwer:
I recommend reviewing commit by commit. Every commit has a single purpose, compiles and passes tests.
This modifies the database schema, but provides defaults for the new `benchmarks` column, and the default matches the old behavior, so I believe this should migrate fine.

This work prepares for implementing step 3 of [#t-compiler/performance > Proposal for &#96;@rust-timer triage&#96; command](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/Proposal.20for.20.60.40rust-timer.20triage.60.20command/with/619288960)

r? @Kobzol

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -1689,10 +1689,17 @@ async fn create_benchmark_configs(
         }
     }
 
+    let benchmark_filter_accepts = |benchmark: &str| {
+        job.benchmarks().is_empty() || job.benchmarks().iter().any(|b| b == benchmark)
+    };
+
     let benchmarks: Vec<Benchmark> = all_compile_benchmarks
         .iter()
         .filter(|b| {
             if !bench_compile_benchmarks.contains(&b.name) {
+                return false;
+            }
+            if !benchmark_filter_accepts(&b.name.0) {
                 return false;
             }
             // Run stable benchmarks for releases and non-stable benchmarks for everything
@@ -1702,6 +1709,10 @@ async fn create_benchmark_configs(
         })
         .cloned()
         .collect();
+
+    if !benchmark_filter_accepts("rustc") {
+        bench_rustc = false;
+    }
 
     let compile_config = if bench_rustc || !benchmarks.is_empty() {
         if toolchain.components.rustdoc.is_none()
@@ -1757,7 +1768,7 @@ async fn create_benchmark_configs(
         .await?;
         Some(RuntimeBenchmarkConfig {
             runtime_suite,
-            filter: RuntimeBenchmarkFilter::keep_all(),
+            filter: RuntimeBenchmarkFilter::new(vec![], job.benchmarks().to_vec()),
             iterations: DEFAULT_RUNTIME_ITERATIONS,
             target: job.target().into(),
         })
--- a/database/schema.md
+++ b/database/schema.md
@@ -270,6 +270,7 @@ Columns:
   test (e.g. llvm, cranelift).
 * **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
   execute.
+* **benchmarks** (`string NOT NULL`): Comma-separated list of benchmarks to execute. If empty, all benchmarks in `benchmark_set` will be executed.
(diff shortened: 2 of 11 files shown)


collector/src/bin/collector.rs
```diff
@@ -1703,6 +1710,10 @@ async fn create_benchmark_configs(
         .cloned()
         .collect();
 
+    if !benchmark_filter_accepts("rustc") {
```
JonathanBrouwer, line 1713: I'm not a big fan of hardcoding the "rustc" name here but couldn't think of anything better

site/src/request_handlers/github.rs
```diff
@@ -493,6 +497,10 @@ fn parse_benchmark_parameters<'a>(
         })?;
     }
 
+    if let Some(benchmarks) = &params.benchmarks {
+        parse_benchmarks(benchmarks).map_err(|e| format!("Cannot parse benchmarks: {e}"))?;
```
JonathanBrouwer: This only checks if the benchmarks are a comma separated list of strings. 
I'd love to check if the benchmark names are actual benchmarks, but could not find an easy way to do so

site/src/request_handlers/github.rs
Kobzol, line 473: You can use `get_compile_benchmarks_metadata` to find known benchmarks.

database/src/lib.rs
```diff
@@ -1150,6 +1155,12 @@ impl BenchmarkRequest {
         parse_targets(&self.targets).map_err(|e| anyhow::anyhow!("{e}"))
     }
 
+    /// Get the targets for the request
```
Kobzol, line 1158:
```suggestion
    /// Get the benchmarks for the request
```

database/schema.md
```diff
@@ -270,6 +270,7 @@ Columns:
   test (e.g. llvm, cranelift).
 * **benchmark_set** (`int NOT NULL`): ID of the predefined benchmark suite to
   execute.
+* **benchmarks** (`string NOT NULL`): Comma-separated list of benchmarks to execute. If empty, all benchmarks in `benchmark_set` will be executed.
```
Kobzol: Comma-separated list of benchmarks that are subset of the given `benchmark_set` to execute.

collector/src/bin/collector.rs
Kobzol: That's fine. In fact I would go further and always set `bench_rustc = false` when the benchmark filter is non-empty.

Kobzol (commented):
Looks great, you're a natural in this codebase :) Left a few nits.

site/src/request_handlers/github.rs
JonathanBrouwer: Hmmm is there a similar function for runtime benchmark names? I couldn't find it and would like to include those in the allowed names

JonathanBrouwer: Seems like there isn't and it would require some work to make it available, as it would have to be serialized in the `build.rs`. I'm happy do to that work but would prefer to do that in a follow-up PR

Kobzol: I don't think we have to support runtime benchmarks at all right now, I would just ignore them same as we do `rustc` if the benchmarks filter is non-empty.

JonathanBrouwer: That does mean that runtime benchmarks will never run as part of a `rust-timer triage` command, is that ok?

Kobzol: I think so, we don't really look at them anyway during triage, I think?

JonathanBrouwer: Added checking for valid benchmarks :)

Kobzol (approved):
Great, thanks! I will hold off merging this for a little bit (an hour at least), just to confirm that the ARM PR doesn't cause any issues.

JonathanBrouwer:
^ Rebased to solve conflicts

PR #2553, opened 2026-08-30
Only run changed benchmarks during triage

JonathanBrouwer:
This PR is based on https://github.com/rust-lang/rustc-perf/pull/2552, and because stacked PRs aren't working cross-fork also contains it's commits :c

I recommend reviewing commit by commit. Every commit has a single purpose, compiles and passes tests.

This implements step 3 of [#t-compiler/performance > Proposal for `@rust-timer triage` command](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/Proposal.20for.20.60.40rust-timer.20triage.60.20command/with/619288960)

r? @Kobzol

### DIFF
--- a/site/src/comparison.rs
+++ b/site/src/comparison.rs
@@ -28,7 +28,6 @@ use std::error::Error;
 use std::fmt::Write;
 use std::hash::Hash;
 use std::iter;
-use std::ops::Deref;
 use std::sync::Arc;
 
 type BoxedError = Box<dyn Error + Send + Sync>;
@@ -1062,7 +1061,7 @@ impl ArtifactComparison {
         ) = self
             .compile_comparisons
             .into_iter()
-            .partition(|s| category_map.get(&s.benchmark()) == Some(&Category::Primary));
+            .partition(|s| category_map.get(&s.test_case.benchmark) == Some(&Category::Primary));
         (
             ArtifactComparisonSummary::summarize(
                 primary
@@ -1256,7 +1255,7 @@ impl TestResultComparison {
     }
 
     /// Whether the comparison yielded a statistically significant result
-    pub fn is_significant(&self) -> bool {
+    fn is_significant(&self) -> bool {
         self.relative_change().abs() >= self.significance_threshold()
     }
 
@@ -1285,7 +1284,7 @@ impl TestResultComparison {
     /// Whether the comparison is relevant or not.
     ///
     /// Relevance is a function of significance and magnitude.
-    fn is_relevant(&self) -> bool {
+    pub fn is_relevant(&self) -> bool {
         self.is_significant() && self.magnitude().is_small_or_above()
     }
 
@@ -1359,14 +1358,8 @@ impl From<TestResultComparison> for StatComparison {
 
 #[derive(Debug, Clone)]
 pub struct CompileTestResultComparison {
-    test_case: CompileTestCase,
-    comparison: TestResultComparison,
-}
-
-impl CompileTestResultComparison {
-    pub fn benchmark(&self) -> Benchmark {
-        self.test_case.benchmark
-    }
+    pub test_case: CompileTestCase,
+    pub comparison: TestResultComparison,
 }
 
 impl cmp::PartialEq for CompileTestResultComparison {
@@ -1393,16 +1386,8 @@ impl std::hash::Hash for CompileTestResultComparison {
 
 #[derive(Debug, Clone)]
 pub struct RuntimeTestResultComparison {
-    test_case: RuntimeTestCase,
-    comparison: TestResultComparison,
-}
-
-impl Deref for RuntimeTestResultComparison {
-    type Target = TestResultComparison;
-
-    fn deref(&self) -> &Self::Target {
-        &self.comparison
-    }
+    pub test_case: RuntimeTestCase,
+    pub comparison: TestResultComparison,
 }
 
 impl cmp::PartialEq for RuntimeTestResultComparison {
--- a/site/src/github/triage.rs
+++ b/site/src/github/triage.rs
@@ -78,7 +78,7 @@ pub async fn changed_benchmarks_in_rollup(
         .find(|commit| commit.pr.is_some_and(|pr| pr == rollup))
         .cloned()
     else {
-        bail!("The `@rust-timer triage` command can only be executed in rollups. If this is a rollup, it might be too old.")
+        bail!("The `@rust-timer triage` command can only be executed in merged rollups. If this is a merged rollup, it might be too old.")
     };
     drop(master_commits);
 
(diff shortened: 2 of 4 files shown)


site/src/comparison.rs
```diff
@@ -1387,16 +1380,8 @@ impl std::hash::Hash for CompileTestResultComparison {
 
 #[derive(Debug, Clone)]
 pub struct RuntimeTestResultComparison {
-    test_case: RuntimeTestCase,
-    comparison: TestResultComparison,
-}
-
-impl Deref for RuntimeTestResultComparison {
```
JonathanBrouwer, line 1400: I found the deref here a bit weird, I prefer the fields just being pub. Happy to revert if you like it tho

site/src/request_handlers/github.rs
```diff
@@ -229,6 +231,27 @@ async fn handle_rust_timer(
...
+                    Ok(benches) => benches,
+                    Err(err) => {
+                        main_client
+                            .post_comment(issue.number, format!("{err}"))
+                            .await;
+                        return Ok(github::Response);
+                    }
+                };
+            writeln!(&mut result, "<details>
+<summary>Running triage with {} benchmarks</summary>
```
```suggestion
<summary>Running triage with {} benchmark{suffix}</summary>
```
```rust
suffix = if benchmarks_to_run.len() == 1 { "" } else { "s" }
```

site/src/github/triage.rs
```diff
@@ -64,6 +67,52 @@ pub fn update_triage_body(
...
+    }
+
+    let comparison = calculate_metric_comparison(
+        ctxt,
+        &QueuedCommit {
+            sha: rollup.head.sha,
+            parent_sha: rollup.base.sha,
+
+            // Fields that are not used by `calculate_metric_comparison`
+            pr: 0,
```
Kobzol:
```suggestion
            pr: rollup,
```
We actually know the PR number, right?

Kobzol: But what would be even nicer is to just pass head and base sha to this function.. Could you refactor that please?

site/src/github/triage.rs
```diff
@@ -64,6 +67,52 @@ pub fn update_triage_body(
...
+    )
+    .await
+    .map_err(|e| anyhow::anyhow!("{e}"))?;
+
+    let compile_benchmarks = comparison.compile_comparisons.iter().flat_map(|c| {
+        c.comparison
+            .is_relevant()
+            .then_some(c.test_case.benchmark.to_string())
+    });
+    let runtime_benchmarks = comparison.runtime_comparisons.iter().flat_map(|c| {
```
Kobzol: Since we don't support runtime benchmarks in the filtering atm, let's ignore those here.

JonathanBrouwer: I'm still including runtime benchmarks here, because I had already implemented it.
I can remove it if you want tho

site/src/github/client.rs
```diff
@@ -49,6 +49,19 @@ impl Client {
             .map_err(|e| anyhow::anyhow!("cannot deserialize commit: {:?}", e))
     }
 
+    pub async fn get_pr(&self, pr: u32) -> anyhow::Result<PullRequest> {
```
JonathanBrouwer: I tested this function locally

site/src/github/triage.rs
```diff
@@ -64,6 +67,52 @@ pub fn update_triage_body(
...
+) -> anyhow::Result<Vec<String>> {
+    let rollup = client.get_pr(rollup).await?;
+    if !rollup.title.starts_with("Rollup of") {
+        bail!("The `@rust-timer triage` command can only be executed in rollups.")
+    }
+
+    let comparison = calculate_metric_comparison(
+        ctxt,
+        &QueuedCommit {
+            sha: rollup.head.sha,
```
Kobzol: This won't really work, because the PR's HEAD SHA is not the merge SHA that was merged and benchmarked. We could query the PR from the `benchmark_request` DB table, we even have a unique index on (pr, commit_type), but adding a new query is a bit annoying, just for this. We could query the last N completed requests and try to find the PR amongst them, which would work in practice, but is also a bit meh.

Maybe the easiest would be to use `ctxt.get_master_commits()`, iterate the master commits to find one with the same rollup PR number, and then get its SHA.

Kobzol (commented):
Looks simple enough, though the PR fetching should be unnecessary.

site/src/comparison.rs
Kobzol: Yeah, that seems unnecessary, thanks.

JonathanBrouwer:
Thanks for the review, will fix the PR later today, gotta get some work done for my job which is not yet rustc :)

site/src/github/triage.rs
JonathanBrouwer: Good catch, I totally didn't consider that.
I indeed went for `get_master_commits` was that is by far the easiest and should work for 99% of usecases. I can always add a new query in the future if this becomes a problem.

site/src/github/triage.rs
```diff
@@ -64,6 +67,40 @@ pub fn update_triage_body(
...
+    rollup: u32,
+) -> anyhow::Result<Vec<String>> {
+    let master_commits = ctxt.get_master_commits();
+    let Some(rollup) = master_commits
+        .commits
+        .iter()
+        .find(|commit| commit.pr.is_some_and(|pr| pr == rollup))
+        .cloned()
+    else {
+        bail!("The `@rust-timer triage` command can only be executed in rollups. If this is a rollup, it might be too old.")
```
Kobzol:
```suggestion
        bail!("The `@rust-timer triage` command can only be executed in merged rollups. If this is a merged rollup, it might be too old.")
```

site/src/github/triage.rs
```diff
@@ -64,6 +67,40 @@ pub fn update_triage_body(
...
+    )
+    .await
+    .map_err(|e| anyhow::anyhow!("{e}"))?;
+
+    let compile_benchmarks = comparison.compile_comparisons.iter().flat_map(|c| {
+        c.comparison
+            .is_relevant()
+            .then_some(c.test_case.benchmark.to_string())
+    });
+    let newly_failed_benchmarks = comparison.newly_failed_benchmarks.into_keys();
```
Kobzol, line 100: What's the reasoning for including failed benchmarks? To figure out which PRs broke them? If yes, could you please add a comment here that explains that?

site/src/request_handlers/github.rs
```diff
@@ -229,6 +231,31 @@ async fn handle_rust_timer(
...
+            };
+            writeln!(&mut result, "<details>
+<summary>Running triage with {} benchmark{plural}</summary>
+
+Triage only executes the benchmarks on rollup members, that were changed significantly on the rollup.
+For this rollup, these benchmarks are:\n", benchmarks_to_run.len()).unwrap();
+            for benchmark in &benchmarks_to_run {
+                writeln!(&mut result, "* {benchmark}").unwrap();
+            }
+            writeln!(&mut result, "</details>").unwrap();
```
```suggestion
            writeln!(&mut result, "</details>\n").unwrap();
```

Kobzol (approved):
Looks great! Left a few remaining nits.

site/src/github/triage.rs
JonathanBrouwer: Yes, exactly. Added a comment :)

PR #2556, opened 2026-08-31
Separate artifact size records by target

Kobzol:
Without this, benchmarking both x64 and aarch64 was overwriting each other's artifact size results. The UI will only show x64 for now, I'll implement support for switching the targets in the UI later.

### DIFF
--- a/collector/src/bin/collector.rs
+++ b/collector/src/bin/collector.rs
@@ -2406,21 +2406,33 @@ async fn record_toolchain_sizes(
         aid: ArtifactIdNumber,
         component: &str,
         path: Option<&Path>,
+        target: database::Target,
     ) {
         if let Some(path) = path {
             if let Ok(size) = fs::metadata(path).map(|m| m.len()) {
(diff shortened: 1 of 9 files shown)


collector/src/compile/execute/mod.rs
```diff
@@ -544,6 +544,7 @@ impl<'a> CargoProcess<'a> {
                 client.configure(&mut cmd);
             }
 
+            eprintln!("{cmd:?}");
```
lqd, line 547:
```suggestion
```

Kobzol: Already removed ^^ Thanks!

lqd: Race condition 🏎️

lqd (approved):
(unless you need james' r+, otherwise this looks good to me)

Kobzol:
Involves migrations, so always a bit scary, but I tested them locally, so let's :crossed_fingers:

PR #2561, opened 2026-09-01
Use only master builds for "average duration" of collectors

JonathanBrouwer:
Again the "did not test this code" disclaimer

### DIFF
--- a/site/frontend/src/pages/status/collector.vue
+++ b/site/frontend/src/pages/status/collector.vue
@@ -148,15 +148,11 @@ function jobDuration(job: BenchmarkJob): string {
 }
 
 function averageCollectorDuration(collector: CollectorConfig): string {
-  if (collector.pastRequestDurations.length === 0) {
+  if (collector.averageDurationS) {
+    return formatSecondsAsDuration(collector.averageDurationS);
+  } else {
     return "Unknown";
   }
-  const durationSum = collector.pastRequestDurations.reduce(
-    (acc, req) => acc + req.jobDurationS,
-    0
-  );
-  const averageDuration = durationSum / collector.pastRequestDurations.length;
-  return formatSecondsAsDuration(averageDuration);
 }
 </script>
 
--- a/site/frontend/src/pages/status/data.ts
+++ b/site/frontend/src/pages/status/data.ts
@@ -30,11 +30,6 @@ export type BenchmarkJob = {
   dequeCounter: number;
 };
 
-export type PastRequestDuration = {
-  requestTag: String;
-  jobDurationS: number;
-};
-
 export type CollectorConfig = {
   name: string;
   target: string;
@@ -44,7 +39,7 @@ export type CollectorConfig = {
   dateAdded: string;
   commitSha: string;
   jobs: BenchmarkJob[];
-  pastRequestDurations: PastRequestDuration[];
+  averageDurationS: number | null;
 };
 
 export type StatusResponse = {
--- a/site/src/api.rs
+++ b/site/src/api.rs
@@ -418,14 +418,6 @@ pub mod status {
 
     #[derive(Serialize, Debug)]
     #[serde(rename_all = "camelCase")]
-    pub struct PastRequestDuration {
-        pub request_tag: String,
-        /// Sum of (completed_at - started_at) for all jobs this collector ran for the request, in seconds.
-        pub job_duration_s: u64,
-    }
-
-    #[derive(Serialize, Debug)]
-    #[serde(rename_all = "camelCase")]
     pub struct Collector {
         pub name: String,
         pub target: String,
@@ -435,7 +427,8 @@ pub mod status {
         pub date_added: DateTime<Utc>,
         pub commit_sha: String,
         pub jobs: Vec<BenchmarkJob>,
-        pub past_request_durations: Vec<PastRequestDuration>,
+        /// Average duration of recent master benchmark requests in seconds
+        pub average_duration_s: Option<u64>,
     }
 
     #[derive(Serialize, Debug)]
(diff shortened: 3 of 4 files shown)


site/src/request_handlers/status_page.rs
```diff
@@ -135,11 +136,17 @@ async fn build_collectors(
...
             .or_default()
-            .entry(job.request_tag())
-            .or_default() += job_duration;
+            .entry(job.request_tag());
+        let entry = entry.or_insert(PastRequestDuration {
+            request_tag: job.request_tag().to_string(),
+            job_duration_s: 0,
+            was_filtered: false,
+        });
+        entry.job_duration_s += job_duration.as_secs();
```
JonathanBrouwer: This technically looses some accuracy over summing `Duration`s (which is what we did previously). I don't think this is a problem tho

site/frontend/src/pages/status/collector.vue
```diff
@@ -151,11 +151,16 @@ function averageCollectorDuration(collector: CollectorConfig): string {
...
+
+  let durationSum = 0;
+  let countedRequests = 0;
+  for (const req of collector.pastRequestDurations) {
+    if (req.wasFiltered) continue;
+    durationSum += req.jobDurationS;
+    countedRequests += 1;
+  }
+
+  const averageDuration = durationSum / countedRequests;
```
Kobzol: It is possible that we divide by zero now. The check that returns `Unknown` above should be moved below the for loop and check `countedRequests`.

site/src/request_handlers/status_page.rs
```diff
@@ -135,11 +136,17 @@ async fn build_collectors(
...
+        entry.was_filtered |= !job.benchmarks().is_empty();
```
Kobzol: Let's also check if we have non-default targets, profiles or codegen backends.

site/src/request_handlers/status_page.rs
Kobzol: Probably doesn't matter too much, but also we could just store `(Duration, bool)` there? 😇

Kobzol (commented):
Thanks!

site/src/request_handlers/status_page.rs
JonathanBrouwer: I didn't include a non-default target as a collector presumably benchmarks a single target, so this would just hide the average duration for non-x86 runners

site/frontend/src/pages/status/collector.vue
JonathanBrouwer: Good point, fixed, thanks!

site/src/request_handlers/status_page.rs
Kobzol: Oh. Hmm, actually, this shows a deeper issue, sorry :laughing: We should ignore all jobs of benchmark requests that have non-default parameters. Not just ignore jobs with non-default parameters from all requests :sweat_smile:

Kobzol: But we don't have that data readily available.

Kobzol: Yeah we should only benchmark master benchmark requests. Actually I thought we already do. I knew I wanted to do it at some point =D

JonathanBrouwer: I had deleted my comment earlier because I thought the data was also not easily available, but it is thanks to `get_master_commits`. I hope this isn't terrible for performance

site/frontend/src/pages/status/data.ts
```diff
@@ -44,7 +39,7 @@ export type CollectorConfig = {
   dateAdded: string;
   commitSha: string;
   jobs: BenchmarkJob[];
-  pastRequestDurations: PastRequestDuration[];
+  averageDuration: number | null;
```
Kobzol:
```suggestion
  averageDurationS: number | null;
```

site/src/api.rs
```diff
@@ -435,7 +427,7 @@ pub mod status {
         pub date_added: DateTime<Utc>,
         pub commit_sha: String,
         pub jobs: Vec<BenchmarkJob>,
-        pub past_request_durations: Vec<PastRequestDuration>,
+        pub average_duration: Option<u64>,
```
Kobzol:
```suggestion
        /// Average duration of recent master benchmark requests in seconds
        pub average_duration_s: Option<u64>,
```

site/src/request_handlers/status_page.rs
```diff
@@ -112,15 +110,28 @@ pub async fn handle_status_page(ctxt: Arc<SiteCtxt>) -> anyhow::Result<status::R
...
 async fn build_collectors(
+    ctxt: &SiteCtxt,
     conn: &dyn Connection,
     in_progress_jobs: &HashMap<String, Vec<BenchmarkJob>>,
     past_jobs: &[BenchmarkJob],
 ) -> anyhow::Result<Vec<status::Collector>> {
     // Collector -> request tag -> sum of job durations
     let mut collector_request_durations: HashMap<&str, HashMap<&str, Duration>> = HashMap::new();
     let mut request_start_time: HashMap<&str, DateTime<Utc>> = HashMap::new();
+    let master_commits = ctxt
```
Kobzol: I looked at this a bit closer and there's an easier way to solve this (even though it is not perfect). In `handle_status_page`, when computing `completed_tags`, just filter only master requests.

Kobzol (commented):
I realized there is a simpler way to do this.

As in https://github.com/rust-lang/rustc-perf/pull/2564, the right solution is actually loading last N *master* requests, but on this page that would mean an additional DB query per page load.

site/src/request_handlers/status_page.rs
JonathanBrouwer: That is way better, thanks!

Kobzol (approved):
Looks great, thank you! Let's test this in prod 😁

PR #2562, opened 2026-09-01
Spawn a task for github webhooks to get around timeout

JonathanBrouwer:
To fix the problem that happened here: https://github.com/rust-lang/rust/pull/162117#issuecomment-5497344848

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -33,21 +33,23 @@ pub async fn handle_github_webhook(
             if action != "created" {
                 return Ok(github::Response);
             }
-            handle_issue(ctxt, issue, comment).await
+
+            // Spawn a task to get around the 10s timeout of github webhooks
+            tokio::spawn(async {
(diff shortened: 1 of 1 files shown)


site/src/request_handlers/github.rs
```diff
@@ -187,9 +187,16 @@ async fn handle_rust_timer(
     main_client: &client::Client,
     comment: github::Comment,
     issue: github::Issue,
-) -> ServerResult<github::Response> {
+) {
+    let authorized_users = match get_authorized_users().await {
+        Ok(authorized_users) => authorized_users,
+        Err(err) => {
+            main_client.post_comment(issue.number, err).await;
```
JonathanBrouwer, line 194: I decided to post this as a comment, but maybe just `log::error!` is better?

Kobzol: It's good if we respond with *something* to the command, so this seems fine.

Kobzol (approved):
Thanks. I'll take a look at making the handler a bit faster.

PR #2563, opened 2026-09-01
Dedup changed benchmarks list

JonathanBrouwer:
Prevents the issue which is observable in the hidden part of this comment: https://github.com/rust-lang/rust/pull/162117#issuecomment-5499272803


Kobzol (approved):
Heh, that's funny. Good catch, thanks!

PR #2565, opened 2026-09-01
Speed up triage command

Kobzol:
By removing unnecessary network requests. We could also remove the queue estimation, but DB queries should be relatively fast, and I didn't want to bother refactoring the code further.

### DIFF
--- a/site/src/request_handlers/github.rs
+++ b/site/src/request_handlers/github.rs
@@ -6,7 +6,7 @@ use crate::load::SiteCtxt;
 use std::fmt::Write;
 
 use crate::benchmark_metadata::get_compile_benchmarks_metadata;
-use crate::github::client::Client;
+use crate::github::client::Commit;
 use crate::github::triage::{
     changed_benchmarks_in_rollup, triage_body_end_marker, triage_body_start_marker, TRIAGE_MARKER,
 };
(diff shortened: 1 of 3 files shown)


site/src/request_handlers/github.rs
```diff
@@ -384,13 +416,10 @@ pub fn parse_unrolled_build_message(commit_message: &str) -> Result<UnrolledBuil
 
 async fn enqueue_sha_build(
     ctxt: &Arc<SiteCtxt>,
-    main_client: &Client,
+    commit: Commit,
     issue_number: u32,
     cmd: &BuildCommand<'_>,
 ) -> Result<String, String> {
-    // requested artifacts do not exist errors
-    validate_build_command(cmd).await?;
```
JonathanBrouwer, line 392: We no longer run `validate_build_command` for triage builds now.
I assume what this means is that the triage build will be queued, and then when the run starts we will get some kind of gentile error (i.e. not a panic), lemme check

JonathanBrouwer: Yeah the error handling seems sound there, so we can accept the risk of people miscopying shas to gain some perf :)

https://github.com/rust-lang/rustc-perf/blob/5905632d20b64a3ccf1d8d5861ed3aa96cd0f94f/collector/src/toolchain.rs#L281

JonathanBrouwer (approved):

site/src/request_handlers/github.rs
Kobzol: I only implemented the validation because we had the perf. builds table available too soon. That no longer happens today, so I think that the validation in general has less value. We could still validate concurrently, but probably it's not worth it.

PR #2567, opened 2026-09-02
Fetch GitHub commits concurrently in the triage command

Kobzol:
This should make the command much faster. Provided that it works.


JonathanBrouwer (approved):
I'm a little worried about github rate-limiting us at some point, but 5 at a time seems like it should be fine?

Kobzol:
Based on past experiments, 4-5 usually is fine. But I agree that it can be a problem. If we run into the first issue in production, I'll just revert this.

PR #2568, opened 2026-09-04
Add support for frontend threads to the UI

Kobzol:
This PR adds support for frontend threads to the UI. Even after many refactorings, adding a new benchmark axis to the frontend still requires a lot of changes :( And there is quite a lot of duplication, particularly in the various used TypeScript types. I did a very minor refactoring while adding frontend threads, but otherwise kept things as-is.

I did not adapt the changes from https://github.com/rust-lang/rustc-perf/pull/2529 in this PR, because I wanted to first get the frontend threads working at all, and then do refactorings separately.

After merging this, we should consider doing some cleanup of the current `-4threads` benchmarks, because currently they will be shown as having one thread, which doesn't make much sense.

### DIFF
--- a/site/frontend/src/graph/api.ts
+++ b/site/frontend/src/graph/api.ts
@@ -5,16 +5,17 @@ import {GRAPH_DATA_URL} from "../urls";
 export async function loadGraphs(
   selector: GraphsSelector
 ): Promise<CompileGraphData> {
-  const params = {
+  const params: Dict<string> = {
     start: selector.start,
     end: selector.end,
     kind: selector.kind as string,
     stat: selector.stat,
-    benchmark: selector.benchmark,
-    scenario: selector.scenario,
-    profile: selector.profile,
-    backend: selector.backend,
-    target: selector.target,
+    benchmark: selector.benchmark as string,
+    scenario: selector.scenario as string,
+    profile: selector.profile as string,
+    backend: selector.backend as string,
+    target: selector.target as string,
+    frontend_threads: selector.frontend_threads as string,
   };
   return await getJson<CompileGraphData>(GRAPH_DATA_URL, params);
 }
--- a/site/frontend/src/graph/api.ts
+++ b/site/frontend/src/graph/api.ts
@@ -5,17 +5,17 @@ import {GRAPH_DATA_URL} from "../urls";
 export async function loadGraphs(
   selector: GraphsSelector
 ): Promise<CompileGraphData> {
-  const params: Dict<string> = {
+  const params: Dict<string | null> = {
     start: selector.start,
     end: selector.end,
     kind: selector.kind as string,
     stat: selector.stat,
-    benchmark: selector.benchmark as string,
-    scenario: selector.scenario as string,
-    profile: selector.profile as string,
-    backend: selector.backend as string,
-    target: selector.target as string,
-    frontend_threads: selector.frontend_threads as string,
+    benchmark: selector.benchmark,
+    scenario: selector.scenario,
+    profile: selector.profile,
+    backend: selector.backend,
+    target: selector.target,
+    frontend_threads: selector.frontend_threads,
   };
   return await getJson<CompileGraphData>(GRAPH_DATA_URL, params);
 }
--- a/site/frontend/src/graph/data.ts
+++ b/site/frontend/src/graph/data.ts
@@ -11,6 +11,7 @@ export interface GraphsSelector {
   profile: string | null;
   backend: string | null;
   target: string | null;
+  frontend_threads: string | null;
 }
 
 export interface Series {
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -109,6 +109,9 @@ export interface CompileBenchmarkParameters {
   scenario: string;
   backend: CodegenBackend;
   target: Target;
+  // We treat the frontend thread count as a categorical variable, which is why
+  // it is represented as a string, and not a number.
+  frontend_threads: string;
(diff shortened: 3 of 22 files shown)


site/frontend/src/pages/compare/compile/common.ts
```diff
@@ -103,22 +103,25 @@ export interface CompileBenchmarkMetadata {
...
 }
 
-export interface CompileTestCase {
-  benchmark: string;
-  profile: Profile;
-  scenario: string;
-  backend: CodegenBackend;
-  target: Target;
+export type CompileBenchmarkComparison = CompileBenchmarkParameters & {
+  comparison: StatComparison;
```
panstromek: nit: It's often preferred to use `extends` if you don't need `&` for some more complicated semantics:

```typescript
interface CompileBenchmarkComparison extends CompileBenchmarkParameters {
    comparison: StatComparison
}
```

It looks like the same thing but there's a subtle difference between the two. `&` allows overriding types of base properties, so it's possible that `CompileBenchmarkComparison` doesn't satisfy `CompileBenchmarkParameters` by accident. `extends` is stricter, it will fail to compile if the extension wouldn't be compatible.

site/frontend/src/graph/api.ts
```diff
@@ -5,16 +5,17 @@ import {GRAPH_DATA_URL} from "../urls";
...
-    scenario: selector.scenario,
-    profile: selector.profile,
-    backend: selector.backend,
-    target: selector.target,
+    benchmark: selector.benchmark as string,
+    scenario: selector.scenario as string,
+    profile: selector.profile as string,
+    backend: selector.backend as string,
+    target: selector.target as string,
+    frontend_threads: selector.frontend_threads as string,
```
panstromek: do we ensure these are not null at this point somewhere?

site/frontend/src/graph/data.ts
```diff
@@ -11,6 +11,7 @@ export interface GraphsSelector {
   profile: string | null;
   backend: string | null;
   target: string | null;
+  frontend_threads: string | null;
```
panstromek, line 14: or maybe we should make them non-nullable here?

site/frontend/src/pages/compare/compile/table/detail-resolver.ts
```diff
@@ -66,7 +65,7 @@ export const COMPILE_DETAIL_GRAPHS_RESOLVER: CachedDataLoader<
   CompileDetailGraphs
 > = new CachedDataLoader(
   (key: CompileDetailGraphsSelector) =>
-    `${key.benchmark};${key.profile};${key.scenario};${key.backend};${key.start};${key.end};${key.stat};${key.kinds};${key.target}`,
+    `${key.benchmark};${key.profile};${key.scenario};${key.backend};${key.target};${key.frontendThreads};${key.start};${key.end};${key.stat};${key.kinds}`,
```
panstromek: nit: these would be a bit easier to review for differences if we do something like:

```typescript
[
  key.benchmark,
  ...,
  key.profile
].join(';')
```

panstromek (approved):
Looks ok to me, I only have minor notes and some questions. I can't tell very well whether this might have missed some more places in the code where we should handle this, but that also doesn't seem like a super blocking issue, we can always fix it up later.

site/frontend/src/graph/data.ts
Kobzol: They can be actually null, you can select only a subset of those filters when loading the graph page. But `getJson` expects `Dict<string>`, and I found it annoying to only add fields that are non-null to the parameters.

site/frontend/src/pages/compare/compile/common.ts
Kobzol: I find `&` easier to comprehend, and it looks less like inheritance :laughing: But good point, changed it.

site/frontend/src/pages/compare/compile/table/detail-resolver.ts
Kobzol: Good point, changed it in a few places. We should really just do something like this: https://github.com/rust-lang/rustc-perf/blob/20205dfc793204f6dacfc156c376bc8ddcadc4c7/site/frontend/src/pages/compare/compile/common.ts#L286 But it would need some refactoring to ensure that there are no extra fields that we don't want to store into the keys.

I want `derive(Hash)` in TypeScript :(

panstromek: > ensure that there are no extra fields that we don't want to store into the keys.

yea, the problem is that typescript doesn't really help with that so it's easy to regress that, you can't enforce a lack of properties

site/frontend/src/pages/compare/compile/common.ts
panstromek: Yea, I also like `&` more, but I already got bitten by this :D

btw. TS compiler perf doc also recommends `extends` for better typechecking performance (because the type checker can shortcut the check for whether child satisfies the parent, which doesn't necessarily hold with `&`).

site/frontend/src/graph/data.ts
panstromek: I see, so the `as string` cast is a lie :D. We could change `getJson` param to `Dict<string | null>`, `getJson` already handles them when constructing params:

https://github.com/rust-lang/rustc-perf/blob/7215c554fe99a2b3d648bd6dda87e040b7f4834d/site/frontend/src/utils/requests.ts#L28-L33

Kobzol: Ah, good catch. Changed it to `Dict<string | null>`.

PR #2569, opened 2026-09-06
Collapse irrelevant results in triage summaries

JonathanBrouwer:
I implemented this because triage comments were getting a bit long for my taste, for example https://github.com/rust-lang/rust/pull/162310#issuecomment-5557882488

For an after and then before, see:
https://github.com/rust-lang/bors-kindergarten/pull/65#issuecomment-5369930341


Kobzol (approved):
It's not really "shorten", but more like "collapse" (by default)? In any case, looks fine. Thanks!

JonathanBrouwer:
Right, my initial idea was to actually only show the text without a collapsed block, and I changed that to a collapsed block forgetting to rename the variables. Fixed

PR #2571, opened 2026-09-06
Add `tokio-1.53.1` benchmark

Kobzol:
As I noted on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/247081-t-compiler.2Fperformance/topic/Adding.20triagebot.20to.20the.20suite/near/621948686), we currently do not benchmark almost *any* async Rust code in `rustc-perf`, which I think is quite unfortunate. Async Rust is used very heavily in the Rust ecosystem, and async functions, async closures, and coroutines in general exercise parts of the compiler that we are currently not benchmarking at all.

This PR adds `tokio` as a new benchmark. It is one of the most commonly used crates in the Rust async ecosystem, with 71 thousand reverse dependencies, and almost a billion all-time downloads. Most async Rust codebases will have `tokio` somewhere in their dependency path, so I think that it would be good if we can spot regressions for it.

With three iterations, the full `tokio` benchmarks take ~100s locally.

Here are the leaf crate benchmark results:

image


nnethercote (approved):

Kobzol:
Ok, let's try this!

PR #2572, opened 2026-09-07
Share target directory for different values of frontend threads

Kobzol:
This optimization will make it faster to run benchmarks for multiple values of frontend threads.

### DIFF
--- a/collector/src/compile/benchmark/mod.rs
+++ b/collector/src/compile/benchmark/mod.rs
@@ -287,13 +287,19 @@ impl Benchmark {
             return Ok(());
         }
 
+        // These two "leaf" benchmark parameters are nested together,
+        // because different values of scenarios and frontend threads can share the same target dir.
+        struct ScenarioWithThreads {
+            scenario: Scenario,
+            frontend_threads: FrontendThreads,
+        }
+
         struct BenchmarkDir {
             dir: TempDir,
-            scenarios: Vec<Scenario>,
             profile: Profile,
             backend: CodegenBackend,
             target: Target,
-            frontend_threads: FrontendThreads,
+            scenarios: Vec<ScenarioWithThreads>,
         }
 
         // Materialize the test cases that we want to benchmark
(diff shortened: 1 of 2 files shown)


collector/src/compile/benchmark/mod.rs
```diff
@@ -432,65 +451,58 @@ impl Benchmark {
...
-                    )
-                    .processor(processor, Scenario::Full, "Full", None)
-                    .run_rustc(true)
-                    .await?;
-                }
+            {
+                let mut frontend_threads = scenarios
+                    .iter()
+                    .map(|s| s.frontend_threads)
+                    .collect::<HashSet<_>>()
```
JonathanBrouwer: Instead of collecting to a `HashSet` and then immediately iterating again, I think it's nicer to just call `.dedup()` after the `.sort_unstable()`, that makes the intent clearer

collector/src/compile/benchmark/mod.rs
```diff
@@ -322,17 +330,26 @@ impl Benchmark {
                         if remaining_scenarios.is_empty() {
```
JonathanBrouwer: What is the point if this `if`?

collector/src/compile/benchmark/mod.rs
JonathanBrouwer, line 1: Is there a risk that rustc re-uses part of the incremental cache between different `frontend_threads` values? If so this could make subsequent runs in the same target directory faster than they would otherwise be

JonathanBrouwer (changes requested):

collector/src/compile/benchmark/mod.rs
```diff
@@ -287,13 +287,19 @@ impl Benchmark {
...
+        }
+
         struct BenchmarkDir {
             dir: TempDir,
-            scenarios: Vec<Scenario>,
             profile: Profile,
             backend: CodegenBackend,
             target: Target,
-            frontend_threads: FrontendThreads,
+            scenarios: Vec<ScenarioWithThreads>,
```
JonathanBrouwer: nit: The naming here is a little inconsistent. Maybe consistently call these "leaf_parameters" / "leafs" and also rename the `ScenarioWithThreads` type? I'm also ok with other names
Feel free to ignore this comment tho

collector/src/compile/benchmark/mod.rs
Kobzol: Ah, no point anymore, thanks.

collector/src/compile/benchmark/mod.rs
Kobzol: Every iteration of the loop is executed on a separate copy of the target directory. I'll add a comment, as it wasn't obvious and also confused me.

collector/src/compile/benchmark/mod.rs
Kobzol: Renamed it.

collector/src/compile/benchmark/mod.rs
JonathanBrouwer: Aah I see, that makes a lot more sense, thanks!

JonathanBrouwer (approved):

PR #2573, opened 2026-09-07
UI improvements related to frontend threads

Kobzol:
Follow-up to the recent PR that added frontend threads to the UI. Found some improvements locally.

### DIFF
--- a/site/frontend/src/pages/compare/compile/common.ts
+++ b/site/frontend/src/pages/compare/compile/common.ts
@@ -4,6 +4,7 @@ import {benchmarkNameMatchesFilter, targetMatchesFilter} from "../shared";
 import {
   DEFAULT_COMPILE_TARGET_TRIPLE,
   DEFAULT_FRONTEND_THREAD_COUNT,
+  sortTargets,
 } from "../../../api";
 
 export type CompileBenchmarkFilter = {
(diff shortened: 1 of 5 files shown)


site/frontend/src/pages/compare/compile/compile-page.vue
```diff
@@ -305,7 +305,7 @@ function storeFilterToUrl(
 }
 
 function updateFilter(newFilter: CompileBenchmarkFilter) {
-  storeFilterToUrl(newFilter, defaultCompileFilter, getUrlParams());
+  storeFilterToUrl(newFilter, defaultCompileFilter.value, getUrlParams());
```
panstromek, line 308: The code here and in other places references a variable before it is defined (lexically). That throws a runtime error if the function is called before the variable is initialized (in so called Temporal Dead Zone).

This is a second time today I was initially a bit confused looking for variable definition because I wasn't expecting it to come after it's being mentioned. It's somewhat unusual, because it can cause "spooky errors at a distance" during refactoring, there's an ESLint lint against it: https://eslint.org/docs/latest/rules/no-use-before-define

I'd probably recommend against doing that, but I'm not sure it's worth the churn to change that atm.

site/frontend/src/pages/compare/compile/common.ts
```diff
@@ -86,9 +101,15 @@ export type Target = "x86_64-unknown-linux-gnu" | "aarch64-unknown-linux-gnu";
 
 export type CompileBenchmarkMap = Dict<CompileBenchmarkMetadata>;
 
+type AssertCompileParameterKey<U extends keyof CompileBenchmarkParameters> = U;
```
panstromek, line 104: hope this works as expected, I sometimes find that TypeScript unexpectedly coerces a union like this to a `string` and your code will happily assert that `string extends string` tsc verified 🫡

panstromek (approved):

site/frontend/src/pages/compare/compile/compile-page.vue
Kobzol: I agree that it can be an error, but at the same time it's quite unlikely to happen, I think. I try to order the code so that functions are above `Ref`s and `Computed`s, it makes it easier for me to orient myself within the code. And here the code is short enough to check that this function is not called too soon (it's only called from the UI anyway).

site/frontend/src/pages/compare/compile/common.ts
Kobzol: It did work when I tried to use an invalid key locally. But this is just a small nice-to-have, it's not really that important anyway.

PR #2577, opened 2026-09-07
add ctrl-c handling to compare TUI

malezjaa:
I found it weird that this doesn't already exit with ctrl+c


Kobzol (approved):
Makes sense, thanks!

PR #2579, opened 2026-09-13
Fix normalization of profiles and scenario in the frontend

### LINKED ISSUE #2578: <invalid profile> in "Command for profiling locally" for doc json
Image

Should be DocJson instead.

Kobzol:
And also make it forward compatible.

Fixes: https://github.com/rust-lang/rustc-perf/issues/2578


Kobzol:
https://github.com/rust-lang/rustc-perf/pull/2581 should fix CI.

site/frontend/src/pages/compare/compile/table/shortcuts/utils.ts
```diff
@@ -2,14 +2,22 @@ import {Profile} from "../../common";
...
+}
+
+// From https://stackoverflow.com/a/54651317/1107768.
+function kebabToPascalCase(text: string): string {
+  return text.replace(/(^\w|-\w)/g, clearAndUpper);
+}
+
+function clearAndUpper(text: string): string {
+  return text.replace(/-/, "").toUpperCase();
 }
```
panstromek, line 23: btw. `Vue` exports some of these case transform functions because they use it to normalize prop and component names. It's `camelize`, `hyphenate` and`capitalize`.

Technically they are marked as `@private` but this is JavaScript, the land of the free, where privacy doesn't really exist and no GDPR will protect these poor functions from being Hyrumized.

panstromek (approved):
I think this is reasonable for now.

In the long term, I think we should probably let the backend generate those commands and just pass them to the frontend? In the end, letting the code that parses these params also generate them makes the most sense to me, but I'm not sure how much code sharing we do between site backend and collector.

site/frontend/src/pages/compare/compile/table/shortcuts/utils.ts
Kobzol: Hmm, good to know. But also, sounds quite dirty to reuse those functions :laughing:

PR #2584, opened 2026-09-15
Edition 2024 + Don't post results on members of triage runs, unless results are relevant

JonathanBrouwer:
I wanted to use an if let chain, so I upgraded the crates to edition 2024. 

I recommend reviewing this commit by commit. 
I did the upgrade by running `cargo fix --edition` and upgraded the crates, this is the first commit.
I then undid some of the changes in the first commit which were false positives, this is the second commit.
The rest of the commits should be obvious.


nnethercote (approved):
- If-let chains: hooray!
- Somewhat-arbitrary changes to use item sort order: boo!

(I know the latter is out of your control.)

It would have been better to do the Edition 2024 changes in one PR and the "Don't post results" change in another PR, but now that they're together I'll just r+.

PR #2585, opened 2026-09-16
Allow filtering Clippy benchmarks in the compare page

Kobzol:
I found this missing while analyzing the results of the parallel frontend experiment.


panstromek (approved):
cool. I thought this is missing because we don't run `clippy` benchmarks by default

Kobzol:
Ideally, the list would be dynamic, based on the data returned from the backend, but that would require some annoying refactoring, so this is a hotfix in the meantime :laughing:

panstromek:
I agree. I'm a bit busy with work and some other stuff, but I plan to look at the site code a bit more when I have time and help with some cleanups.

PR #2587, opened 2026-09-21
Add deref-sidebar-stress benchmark.

xonx4l:
This PR adds a stress benchmark for rustdoc sidebar generation . 

A deep `Deref` chain where every target type defines a method with the same name, so many sidebar links want the same anchor and rustdoc has to make each one unique.

The issue can be seen here rust-lang/rust#158174 and the fix here  rust-lang/rust#162976.


collector/compile-benchmarks/deref-sidebar-stress/README.md
```diff
@@ -0,0 +1,22 @@
+A deep `Deref` chain where every target type defines a method with the same
```
Kobzol, line 1: Can you turn this into `README.md` and wrap the code block in
```
```python
```
please? Thanks!

collector/compile-benchmarks/deref-sidebar-stress/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "deref-sidebar-stress"
+version = "0.1.0"
+edition = "2021"
```
Kobzol:
```suggestion
edition = "2024"
```

Kobzol (commented):
Seeing the code, I think it would stress more things than just the rustdoc sidebar 😆 In any case, looks reasonable. Left two nits.

collector/compile-benchmarks/deref-sidebar-stress/README.md
xonx4l: Done. Thanks!

Kobzol (approved):
Thanks!

PR #2588, opened 2026-09-22
Rollup some perf-sensitive PRs

JonathanBrouwer:


Kobzol:
What's the thinking behind basing this only on the number of changed benchmarks? That if the PR gets into a rollup, the small number of changed benchmarks will make it possibel to do `@rust-timer triage all`? I'm not sure if that is always the case, because if more of those PRs get into a rollup, it can start to add up. I think that we should still pay attention to the magnitude of the regressions, rather than just the number of changed benchmarks.

JonathanBrouwer:
The reason we don't roll up all perf sensitive prs is that it would cost too much perf capacity to benchmark a rollup containing PRs that change a lot of benchmarks. The magnitude of the benchmark change does not matter for this problem, which is why I didn't pay attention to it. 

> if more of those PRs get into a rollup, it can start to add up

This is true, but taking the benchmark magnitude into account does not lessen this problem. Once a benchmark result is significant the benchmarks need to run, regardless of magnitude. 

To improve this problem I propose we mark PRs as rollup=iffy if they affect performance, so rollupers can take it into account and not put multiple perf sensitive PRs in one rollup

Kobzol:
I think it's not just about the capacity, in general it is easier to have perf. affecting PRs be separate, it's slightly easier to revert, it's slightly easier to bisect, and it's also easier to show wins/losses by pointing people to the PR directly.

I'm fine with experimentally marking the PRs with your heuristic as `iffy`. Though it should be noted that this heuristic can actually mark some PRs as `iffy` that wouldn't be even `never` before, i.e. act in the opposite manner that you intend them to. It might be interesting to analyse historical data to see how often that would happen.

